PR #17, opened 2014-09-24
Added -stdlib=libc++ so gcc/clang will find C++ 11 features on the Mac

EricRFMA:
Added c++ compiler flag so that the compiler will find include files on Mac for C++ 11.  Don't know if this will work on Linux, I'm trying to set up an environment to test that in.  Closes #1786

### DIFF
--- a/anagram/CMakeLists.txt
+++ b/anagram/CMakeLists.txt
@@ -13,7 +13,7 @@ find_package(Boost 1.55 REQUIRED COMPONENTS unit_test_framework date_time regex)
 
 # gcc/clang won't enable C++11 features without this flag
 if("${CMAKE_CXX_COMPILER_ID}" MATCHES "(GNU|Clang)")
-    set(CMAKE_CXX_FLAGS "-std=c++11")
+    set(CMAKE_CXX_FLAGS "-std=c++11 -stdlib=libc++")
 endif()
 
 # Configure to run all the tests?


kytrinyx:
I just realized that it's been over 6 months on this one. Has anyone had a chance to ask some folks to test? Should we get the the build passing and just merge, and then let people submit issues if they run into trouble?

LegalizeAdulthood:
It needs more testing.  I have access to a Mac machine via work now, so I can do some testing.  The problem is that it can't just be merged, because it's a bit of a jello bug.  You poke here and it pushes out there.  You "fix" one thing for one person and then it breaks for other people.

LegalizeAdulthood:
This originally came up because the poster was using Xcode and not the gcc from homebrew as recommended in the setup instructions.

I've created https://github.com/exercism/xcpp/issues/45 to address the Xcode/gcc use cases.

kytrinyx:
OK, excellent. Thanks.

masters3d:
This seems to be resolved on the mac. 
As of xcode 7.3 / boost 1.60.0_1 / cmake 3.5.1 using clang AppleClang 7.3.0.7030029

LegalizeAdulthood:
@masters3d So it now works as-is with Xcode?  Have you tried with gcc from homebrew?

masters3d:
I haven't tried it with GCC. Xcode doesn't support it by default, I also have Clion but I don't see the option to set it. I think it must be a CMAKE command line option only. 

http://stackoverflow.com/questions/24380456/how-can-i-make-cmake-use-gcc-instead-of-clang-on-mac-os-x

masters3d:
I got a chance to test it with GCC-5  using CLion and it works fine. 

```
$ cmake -DCMAKE_C_COMPILER=/usr/local/Cellar/gcc/5.3.0/bin/gcc-5 -DCMAKE_CXX_COMPILER=/usr/local/Cellar/gcc/5.3.0/bin/g++-5
-- The CXX compiler identification is GNU 5.3.0
-- Checking whether CXX compiler has -isysroot
-- Checking whether CXX compiler has -isysroot - yes
-- Checking whether CXX compiler supports OSX deployment target flag
-- Checking whether CXX compiler supports OSX deployment target flag - yes
-- Check for working CXX compiler: /usr/local/Cellar/gcc/5.3.0/bin/g++-5
-- Check for working CXX compiler: /usr/local/Cellar/gcc/5.3.0/bin/g++-5 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Boost version: 1.60.0
-- Found the following Boost libraries:
--   unit_test_framework
--   date_time
--   regex
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_C_COMPILER


-- Build files have been written to: #### my folder ####
```

LegalizeAdulthood:
Great, thanks for your assistance!

PR #42, opened 2015-05-25
queen-attack is missing tests (and example implementation) for the other diagonal

siebenschlaefer:
The unit tests and the example implementation test the north-west / south-east diagonal, but are missing the north-east / south-west diagonal.
Also there is no test case where the queens are _not_ attacking each other.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,7 @@
 tmp
 bin/configlet
 bin/configlet.exe
+CMakeCache.txt
+CMakeFiles/
+Makefile
+cmake_install.cmake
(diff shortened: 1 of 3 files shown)


LegalizeAdulthood:
Thanks for the additon!

I mimiced the JavaScript test cases; does the example JavaScript solution also suffer from the same problem?

siebenschlaefer:
@LegalizeAdulthood Yes it does. It only tests cases where ``xDistance === yDistance``, but not cases where ``xDistance === -yDistance``. I just added two test cases and amended the example implementation. It's in my repository.

PR #46, opened 2015-07-09
Add test case for bob with leading and trailing whitespace

daniellehrner:


LegalizeAdulthood:
The reference implementation didn't need updating to handle this?

daniellehrner:
No, the test is successful with the example.cpp.
I've written the test, because I saw some solutions which just check the first and the last char for whitespaces, which in my opinion could lead to some errors. These checks would fail with this test.

LegalizeAdulthood:
When a test is added without any changes in the implementation it tells me that this test and another test are doing the same thing.  However, this may not be the case for the solutions you are describing, it just happens to be the case for my solution.

It's hard for me to tell from the diff; do we have multiple leading and trailing whitespace in the new test?  It looks like there might just be a single leading space while there are multiple trailing spaces.  Can we make sure it has multiple leading and trailing spaces in the test input?

PR #51, opened 2015-07-27
Add testcase with invalid char at the end

NobbZ:
Made this to make more clear how input shall be handled that looks valid right away from start but turns out to be invalid somewhere in the middle or the end.

The behaviour I test against should match the current behaviour of the reference-implementation. But perhaps we should require to `throw` something instead?


LegalizeAdulthood:
LGTM.  Thanks.  I don't think an exception is needed here; otherwise we should also throw for 'carrot'.

PR #56, opened 2015-12-19
Switch to container-based Travis CI infrastructure

kytrinyx:
This gets us off the legacy build system, which means
that the builds will start more quickly and also likely
run more quickly (and with more consistent build times).

See https://docs.travis-ci.com/user/migrating-from-legacy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,4 @@ script:
 - bin/configlet .
 - cmake -G "Unix Makefiles"
 - make
+sudo: false


LegalizeAdulthood:
Why are the travis CI builds failing with this change?

kytrinyx:
The previous build is also failing, and that was a change to a text file.

Could a change on Travis' side have changed anything? I know they changed something with regard to the legacy platform a few weeks ago: https://blog.travis-ci.com/2015-11-27-moving-to-a-more-elastic-future

kytrinyx:
According to https://docs.travis-ci.com/user/languages/cpp:

> Travis CI VMs are 64-bit and provide versions of:
> [...]
> core GNU build toolchain (autotools, make), cmake, scons

Seems like it should be there.

LegalizeAdulthood:
By turning sudo off, you're preventing these commands from working:
```
before_install:
- sudo add-apt-repository -y ppa:kalakris/cmake
- sudo add-apt-repository -y ppa:boost-latest/ppa
- sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
- sudo apt-get -qq -d update
```

By default the travis image doesn't have cmake, boost or a C++11 compatible gcc, which is why I had to install these packages before running the build.

LegalizeAdulthood:
Also, I'm not seeing any failed builds except the one for your pull request.

Were you [looking here](https://travis-ci.org/exercism/xcpp/builds)?

LegalizeAdulthood:
OK, so they now have cmake, which is new.  However, I couldn't find anything stating what version of gcc/clang they have in their image.

LegalizeAdulthood:
From the last passing build:

```
$ gcc --version
gcc (Ubuntu/Linaro 4.6.3-1ubuntu5) 4.6.3
```

so we will need to keep `sudo` in order to install a more recent version of gcc.

kytrinyx:
Ah, ok, that makes sense. Thanks for digging.

I was looking at the other open pull request. Not sure what's going on there though.

LegalizeAdulthood:
The other open pull request is related to making things play nice on Mac based on certain versions of OS and Xcode.  More investigation is needed to determine a solution that works everywhere and not just for that specific case.

PR #60, opened 2015-12-31
Fix inconsistency in triangle test

### LINKED ISSUE #59: Inconsistency in `triangle`s test
There is some inconsistency in the tests for `triangle` exercise.

The tests do expect to throw an `DomainError` when all sides of the triangle are `0`, but to return `triangle::illegal` when we have negative length sides or don't fullfil the triangle inequality.

To make this consistent, we should either throw something in all three cases or return `triangle::illegal` in all cases.

To line up with other tracks that support some kind of throwing, I think we should throw in every case and get rid of `triangle::illegal`.

NobbZ:
FIX #59


LegalizeAdulthood:
Looks great, thanks!

PR #62, opened 2016-01-04
Robot names begin with two alphabetic characters

LegalizeAdulthood:
Addresses discussion in exercism/exercism.io#2441


kytrinyx:
Looks good to me.

PR #64, opened 2016-01-25
Creating ABOUT.md

devonestes:
I saw there was no description of C++ on the main page for exercism.io, so I wanted to help create one!

### DIFF
--- a/docs/ABOUT.md
+++ b/docs/ABOUT.md
@@ -1,0 +1,6 @@
+[C++](https://en.wikipedia.org/wiki/C%2B%2B) (pronounced `cee plus plus`) is a widely used, compiled, multi-paradigm language that extends
+the basic features found in C. Originally created by Bjarne Stroustrup at Bell Labs, it is often used in places where speed and very fine
+grained control over a program's execution are needed. It is most often found in use as a systems language, and is also great for
+situations where resources are significantly constrained. The standard for C++ is maintained by the International Organization for
+Standardization (ISO), and the current version of C++ as of December 2014 is C++14 (named not as the 14th version of C++, but rather
+signifying that the standard was ratified in 2014).


docs/ABOUT.md
```diff
@@ -0,0 +1,9 @@
+[C++](https://en.wikipedia.org/wiki/C%2B%2B) (pronounced `cee plus plus`) is a widely used, compiled, multi-paradigm language that extends
+the basic features found in C. Originally created by Bjarne Stroustrup at Bell Labs, it is often used in places where speed and very fine
+grained control over a program's execution are needed. It is most often found in use as a systems language, and is also great for
+situations where resources are significantly constrained. The standard for C++ is maintained by the International Organization for
+Standardization (ISO), and the current version of C++ as of December 2014 is C++14 (named not as the 14th version of C++, but rather
+signifying that the standard was ratified in 2014).
+
+_Help us explain this better! File a GitHub issue at https://github.com/exercism/xcpp/issues if you have suggestions, or submit a
```
kytrinyx: This bit is included in the website for all the languages--it doesn't need to be included here.

kytrinyx:
Thanks!

docs/ABOUT.md
devonestes: Got it - removed!

LegalizeAdulthood:
Thanks for the contribution!

I have some comments and suggestions I'd like to see here before merging.

One tiny nit: please format lines for about 72 columns in width.

Please link Bjarne Stroustrup's name to his wikipedia page.

Please don't describe C++ as an extension to C.  For detailed reasons why, watch Kate Gregory's excellent video ["Stop Teaching C"](https://www.youtube.com/watch?v=YnWhqhNdYyk).  In summary, describing C++ as an extension to C makes people think they should learn C first and that is incorrect.  When teaching C++ to C programmers, they have to **unlearn** a bunch of habits that are either unnecessary in C++ or error-prone or both.  C++ shares many things with C (the good bits) and diverges from C in significant ways (by providing better mechanisms that avoid the bad bits).  While C++ does retain syntactic and semantic compatibility with a majority of C source code, I'd rather just not mention it at all because for exercism.io it is irrelevant.  C style techniques are exactly the **opposite** of what I am trying to encourage with the C++ language track on exercism.io.

The way the language is described makes it sound very special case as if it weren't suitable for general programming.  This is not the case.  I would take my cues from Bjarne Stroustrup's ["Essense of C++"](https://www.youtube.com/watch?v=86xWVb4XIyE) video where he describes his goals for C++ on slide 5 (11m50s):

* Type safety
  * Encapsulate necessary unsafe operations
* Resource safety
  * Not all resource management is managing memory
* Performance
  * For some parts of almost all systems, it's important
* Predictibility
  * For hard and soft real-time systems
* Teachability
  * Complexity of code should be proportional to the complexity of the task
* Readability
  * People and machines ("analyzability")

On slide 8 (22m00s), he describes what C++ **is** in two lines:

* Direct map to hardware
  * of instructions and fundamental data types
* Zero-overhead abstraction
  * Classes with constructors and destructors, inheritance, generic programming, functional programming techniques

Both of these contribute to C++ code obtaining performance advantages over other languages.  However, the zero-overhead abstraction point is really important.  People often confuse "fast" with "low-level" and C++ dramatically proves that you don't have to give up one in order to obtain the other.

I'm not sure if the average programmer knows what it means for a language to be "multi-paradigm"; I think it would be better to say that C++ supports procedural, object-oriented, functional and generic programming.  Each of these can be linked to an appropriate page on wikipedia.

IMO, it isn't "fine grained control over a program's execution" that people want from C++, but deterministic resource management.  That is, the power of the language lies in the destructor.  This is the mechanism that achieves resource safety or as Scott Meyer's calls it "the language has a built-in undo".  The abstraction mechanisms are the means by which people obtain the deterministic resource management.

Describing C++ as (my paraphrase) a language for control freaks obsessed with performance used in systems programming is an inaccurate summary of how the language is used in practice.  The language is used in a widespread set of problem domains and often by developers without formal training in software engineering.  For instance, [Cinder](http://libcinder.org) (a multimedia C++ library) won the [Cannes Innovation Grand Prix award](http://adage.com/article/special-report-cannes-2013/coding-platform-cinder-innovation-grand-prix/242175/).  Using Cinder, a self taught programmer created the [Planetary music browser](http://planetary.bloom.io/)

LegalizeAdulthood:
Thanks for your contribution!  In pull request #69 most of the requested changes have been made.

PR #66, opened 2016-02-14
Hopefully fix formatting of bulleted list.

toolness:
A bulleted list in the [Installing C++](http://exercism.io/languages/cpp) page looks fine when rendered in GitHub as markdown, but I guess Redcarpet or something might be parsing/rendering slightly differently, because it shows up like this on the website:

> 

I am guessing that my fix will fix things but I didn't actually test it out.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -64,6 +64,7 @@ latest version of the Microsoft Visual C++ compiler.
 CMake is an open source cross-platform build system that generates build
 scripts for your native build system (`make`, Visual Studio, Xcode, etc.).
 Exercism's C++ track uses CMake to give you a ready-made build that:
+
 * compiles the tests
 * compiles your solution
 * links the test executable


kytrinyx:
Thanks!

toolness:
Thanks @kytrinyx!

I've been noticing more instances of this same bug--at least once more in the C++ docs, and once in the JS docs--which makes me wonder if the problem should actually be fixed further upstream? I mean, GitHub *does* render the markdown in a way that will lead contributors to assume things will look okay if they don't put an empty line between a paragraph and a bulleted list...

This bug seems particularly odd to me given that I'd assumed GitHub uses a lot of the same underlying code as exercism, like Redcarpet, to render the markdown. Ah well. I could also just continue to file PRs like this when I run into the bugs, it's all good either way. :grin:

kytrinyx:
If you have time to look at why the missing blank line causes the problem here, that would be great.

LegalizeAdulthood:
I'm pretty sure I relied on github's rendering to check the markup on these docs.  Thanks for the fixes!

PR #67, opened 2016-02-14
Hopefully fix formatting of list in TESTS.md.

toolness:
[Running the Tests](http://exercism.io/languages/cpp) has a similar problem as #66 in that the first bulleted list appears inline and unformatted. Hopefully this fixes it.


kytrinyx:
Yeah, that should fix it. Thanks!

LegalizeAdulthood:
Thanks!

PR #70, opened 2016-03-05
fix pair printing in word-count to work in boost v 1.59 and above

lukasz-m-maciejewski:
I think a Travis job that that would run the tests against latest version of boost would be a good idea.

### DIFF
--- a/etl/etl_test.cpp
+++ b/etl/etl_test.cpp
@@ -2,6 +2,7 @@
 #define BOOST_TEST_MAIN
 #include <boost/test/unit_test.hpp>
 
+#if BOOST_VERSION < 105900
 namespace boost
 {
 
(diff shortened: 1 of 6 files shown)


kytrinyx:
Nice! Yeah, this is a great idea. Would you mind squashing the commits, please?

lukasz-m-maciejewski:
Done. :-D
And it was so easy. After all the drudgery I experience daily working with SVN, this is both refreshing and exhilarating. :-)

LegalizeAdulthood:
I already have a changeset I'm preparing for this, I need some cycles to merge it into master.

lukasz-m-maciejewski:
Let me know if  I can do anything to help then. My fixes were just a quick-and-dirty solution to make this pain of having to fix tests compilation go away.

LegalizeAdulthood:
I've rebased boost branch on latest master and if travis passes, I will merge that to master.

LegalizeAdulthood:
Pushed.  Please pull the latest and see how that works for you.

LegalizeAdulthood:
Also, I looked into getting boost >=1.59 on travis and I couldn't find a linux PPA for it, which would mean downloading the tarball and building from source.  That is likely to make the travis job take so long that it will be killed before completing.  If you know a linux PPA where I can get prebuilt boost binaries for >=1.59, then I can enable that in the build matrix.

lukasz-m-maciejewski:
Yeah, it works. This is so good.

As to Ubuntu repository with latest boost, I can't seem to find one either. I'll look into it some more and get back to you if I find something. 

I suspect, that simply downloading the source tarball whouldn't be such a bad idea. From what I can see in the build process, the only compiled libraries from boost are the unit testing framework, regex and date_time. I'm not sure if and how they are actually used in the reference solutions - I didn't want to spoil my own fun doing the exercises. I guess, however, that date_time and regex could be replaced by standard library. Then maybe we could get away with building only the unit testing framework. I'll try to hack this and I'll let you know if anything comes of it.

LegalizeAdulthood:
The test stuff is used for all the tests in every exercise.  Date/time and regex are used where appropriate in various exercises, but obviously not in every exercise.

Since my solution is working for you, I'm going to close this pull request.  Thanks for your help!

PR #72, opened 2016-03-25
Fix markdown formatting of bullets

k4rtik:
They currently look as follows in the testing section of [C++ track doc](http://exercism.io/languages/cpp#tests):

> Once the build environment has been created by CMake, you can build your code using the appropriate command for your environment: * Linux with make: make * Windows with Visual Studio 2013: Select Build / Build Solution from the menu. * MacOS with Xcode: Select Build from the toolbar

instead of

> Once the build environment has been created by CMake, you can build your code using the appropriate command for your environment:
> * Linux with make: make
> * Windows with Visual Studio 2013: Select Build / Build Solution from the menu.
> * MacOS with Xcode: Select Build from the toolbar

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -50,6 +50,7 @@ batch file (Windows) for this command.
 
 Once the build environment has been created by CMake, you can build your
 code using the appropriate command for your environment:
+
 * Linux with make: `make`
 * Windows with Visual Studio 2013: Select Build / Build Solution from the menu.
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks! There's an ongoing fix in progress at the level of the markdown parser (https://github.com/exercism/exercism.io/pull/2775) but we're stuck on some dependencies.

k4rtik:
@kytrinyx oh I wasn't aware. Thanks for the pointer.

PR #85, opened 2016-06-11
Update Cmake with full path for cmake

masters3d:
It is not clear that ``..`` is mean for the source path. 
I followed this to make the changes
http://stackoverflow.com/questions/18826789/cmake-output-build-directory/24435795#24435795

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -49,8 +49,9 @@ $ sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90
 
 #### MacOS
 
-MacOS users can install gcc 4.8 with [Homebrew](http://brew.sh/) via
-`brew install gcc`.
+MacOS users can install gcc 5.x with [Homebrew](http://brew.sh/) via
+`brew install gcc`.  
+Alternatively, you can use Clang/LLVM included in Xcode 7.x and above.
 
 #### Windows
 
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -140,6 +140,25 @@ $ cmake -G "Unix Makefiles" /path/to/my/source/folder/bob/
 This example shows creating empty files for the implementation before
 running CMake.
 
+
+#### MacOS with GGC from Homebrew with Xcode Installed 
+
+Xcode will overide `gcc` to link to llvm so in order to use GCC from homebrew  
+you have to find the version of GCC you have installed by `$ brew list --versions`  
+Assming that you have version 5.3.0 of GCC, the current exercise is `bob`   
+and we're in the exercise folder:
+
+```
+$ cd /path/to/my/source/folder/bob/
+$ touch bob.{h,cpp}
+$ mkdir gccbuild
+$ cd gccbuild
+$ cmake -DCMAKE_CXX_COMPILER=/usr/local/Cellar/gcc/5.3.0/bin/g++-5 /path/to/my/source/folder/bob/
+```
+
+This example shows creating empty files for the implementation before
+running CMake.
+
 ## Boost.Test Documentation
 
 The Boost.Test documentation is being rewritten to improve usability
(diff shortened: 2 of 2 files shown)


k4rtik:
I do not see how this change is helpful for a beginner.

Given the previous two instructions `mkdir build` and `cd build`, `..` is completely clear and concise about what it is referring to.

masters3d:
If you follow the directions exactly this is the error I get. 

`"myDir///////leap/build" does not appear to contain CMakeLists.txt.
Specify --help for usage, or press the help button on the CMake GUI.
`
which makes sense because if I switch the an empty `build` folder then cmake doesn't know where the source files are.  I am missing something?

masters3d:
In fact I would argue that the ``..`` make it more ambiguous because I I were to actually copy and paste the command I would get:
``$ cmake -G Xcode ..
CMake Error: Error: generator : Xcode
Does not match the generator used previously: Unix Makefiles
Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
``

k4rtik:
@masters3d I am not sure why you got the error you mentioned in the previous to last comment. I have tested the steps multiple times on both OS X and Linux and never observed any issue with them.

But for the error in the last comment, you seem to be mixing two different generators. You must have issued `cmake -G "Unix Makefiles" ..` or equivalent first and then are trying to use the second generator `Xcode` about which cmake is not complaining.

masters3d:
Cmake does not complain if I do an `in source` generation (having the all the cmake files generated at root) but if I want to do an `out-of-source` generation (using a separate folder like `build`) then you get errors.  Which one are you generating?  `in source` or  `out-of-source`? 
".. is completely clear and concise about what it is referring to." What is that?

masters3d:
@k4rtik I am not asking why I get the errors; I know why the error occur but would a beginner know? If you call `cmake -G Xcode ..` (including the .. which are not valid commands) then you get one error but if you just call ``cmake -G Xcode` on the source folder (where the problem is ) then it works but it creates and `in source` generation.

I don't understand what you find confusing about the PR. What would you change?

k4rtik:
According to http://exercism.io/languages/cpp out-of-source build generation is preferred: 

> To keep those generated files separate from your exercise source code, it is common to create a directory called build to hold these generated build files, as well as the compiled code. This will keep your exercise folder uncluttered and tidy.

`..` is referring to parent directory of `build`, here is a trace from my environment, it works just fine:

```
k4rtik at PlatiniumAir in ~/exercism/cpp/gigasecond/build
$ rm -rf *
k4rtik at PlatiniumAir in ~/exercism/cpp/gigasecond/build
$ cmake -G "Unix Makefiles" ..
-- The CXX compiler identification is AppleClang 7.3.0.7030031
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Boost version: 1.60.0
-- Found the following Boost libraries:
--   unit_test_framework
--   date_time
--   regex
-- Configuring done
-- Generating done
-- Build files have been written to: /Users/k4rtik/exercism/cpp/gigasecond/build
k4rtik at PlatiniumAir in ~/exercism/cpp/gigasecond/build
$
```

I wouldn't replace any of the `..` with fullpaths in the PR.

masters3d:
:) You are totally right! I stand corrected. Thanks for putting up with me. 
Where can I learn more about the command line?

masters3d:
I think I am just going to have to learn bash once and for all. https://github.com/exercism/xbash/

PR #89, opened 2016-06-15
Porting the circular buffer challenge.

TrophyNinjaShrub:
Sorry I accidentally closed 
https://github.com/exercism/xcpp/pull/80
by one means or another and it didn't seem to want to let me reopen.

### DIFF
--- a/circular-buffer/CMakeLists.txt
+++ b/circular-buffer/CMakeLists.txt
@@ -1,0 +1,42 @@
+# Get the exercise name from the current directory
+get_filename_component(exercise ${CMAKE_CURRENT_SOURCE_DIR} NAME)
+
+# Basic CMake project
+cmake_minimum_required(VERSION 2.8.11)
+
+# Name the project after the exercise
+project(${exercise} CXX)
+
+# Locate Boost libraries: unit_test_framework, date_time and regex
+set(Boost_USE_STATIC_LIBS ON)
+set(Boost_USE_MULTITHREADED ON)
+set(Boost_USE_STATIC_RUNTIME OFF)
+find_package(Boost 1.55 REQUIRED COMPONENTS unit_test_framework  regex)
+
+# Configure to run all the tests?
+if(${EXERCISM_RUN_ALL_TESTS})
+    add_definitions(-DEXERCISM_RUN_ALL_TESTS)
+endif()
+
+FILE(GLOB sources *.cpp)
+
+# Build executable from sources (including headers is bad practice)
+add_executable( ${exercise} ${sources} )
+
+#Use a recent-ish standard
+target_compile_features( ${exercise} PRIVATE cxx_override )
+
+# We need boost includes
+target_include_directories(${exercise} PRIVATE ${Boost_INCLUDE_DIRS})
+
+# We need boost libraries
+target_link_libraries(${exercise} ${Boost_LIBRARIES})
+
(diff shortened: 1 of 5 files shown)


LegalizeAdulthood:
The travis CI build is failing; you'll need to fix that before we can accept the pull request.

LegalizeAdulthood:
If you need help figuring it out, just shout :smile_cat:

LegalizeAdulthood:
OK, now the Travis build is passing, so I've taken a more detailed look at your commit.  Thanks for this contribution.

Your code isn't written in the same style as the rest of the code in this repo.  This implies that I should more formally document the formatting conventions of this repo (See [issue #92](https://github.com/exercism/xcpp/issues/92) and [issue #93](https://github.com/exercism/xcpp/issues/93))  Most of the rest of these comments revolve around the style and formatting of your code and how it differs from the existing code.  Beyond that, I think your tests need some restructuring so that they test only one behavior per test.

In particular, in this repo we don't put spaces inside the parentheses of arguments to functions.  So write: `foo(5)`, not `foo( 5 )`.  This applies to the `CMakeLists.txt` as well as the source code.

Similarly, multiple template or function arguments are separated with a space between them.  So write `circular_buffer<char, 2UL>`, not `circular_buffer<char,2UL>`.

I don't understand the point of this test:

```
//This is some simple book-keeping to let people who are giving feedback know which version of the exercise you solved.
BOOST_AUTO_TEST_CASE(version_test)
{
   BOOST_CHECK_EQUAL( containers::VERSION, 1 );
}
```

None of the other exercises have this sort of a test and I don't see why this should have one.

Your tests aren't following the same whitespace style as the other tests; you're using an indent level of 2 and not 4 as is used consistently throughout the other files in the repo.

Tests should be formatted so that the different phases of execution (setup, exercise, verify, teardown) are clearly seen at a glance.  This means putting a blank line between them.

Separate each test case from the previous by a blank line.

Tests cases should be isolated and test only one thing.  You have test cases that test several things in succession.  There are also many, redundant assertions in your test cases.  If you've already tested some behavior in a previous test case, there is no need to test it again in a subsequent test case.  Each test case should test some new aspect of the behavior and verify that new behavior only.

I don't understand the point of `circular_buffer_test2.cpp`.  Again, none of the other exercises have this.  Presumably you're trying to check that the header doesn't violate the one definition rule, but that isn't really the point of the C++ language track in exercism, nor should it be the point of this exercise, IMO.  I would simply remove this file.  This is something for bringing up in discussion with someone's solution and I don't see the need to try and enforce that through a single exercise.

In `example.hpp`, you've indented everything inside the namespace.  None of the other code in this repo indents because of namespace.

In this repo, we prefer `#if !defined(HEADER_GUARD)` over `#ifndef HEADER_GUARD` because `#ifdef` and `#ifndef` differ only by a single character.  Existing header guard symbols are based off the filename as in `ANAGRAM_H`, yours differ from this convention and should be changed to match the existing code.

For comments, put a space after the `//` as in `// comment`, not `//comment`.

Separate methods in a class by a blank line.

Place braces around all blocks in control structures, even when they are a single statement.

Using `const` on a numeric template argument serves no purpose since the template can't change the argument anyway.

Reference (`&`) and pointer (`*`) annotations hug the identifier, not the type.  So write `int *i`, not `int* i`.

patricksjackson:
Closing this old PR. If you (or anyone else) wants to pick this back up please either reopen this or open a new PR :)

Thanks

PR #90, opened 2016-06-28
sum-of-multiples: Removing implicit list of multiples

patricksjackson:
Updates the tests and example solution to require an explicit
declaration of multiples, matching the test cases from x-common/sum-of-multiples.json

To fix Issue #71


LegalizeAdulthood:
Thanks for the fix!

PR #94, opened 2016-07-20
robot-name: Require uppercase letters

patricksjackson:
Update robot-name to allow only uppercase letters. This keeps the xcpp
track updated with the current instructions. See exercism/x-common#283


kytrinyx:
Sweet, thanks!

@LegalizeAdulthood Would you mind having a look at this one when you have time?

LegalizeAdulthood:
Looks good, thanks for the fix!

PR #95, opened 2016-07-22
Pass bracket-push

patricksjackson:
I added it to the end of the list of exercises, since I wasn't sure where in the list it should go.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ foreach(exercise
     queen-attack
     allergies
     atbash-cipher
+    bracket-push
 )
     set(exercise_dir exercises/${exercise})
     travis_fixup(${exercise})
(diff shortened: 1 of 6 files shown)


kytrinyx:
Thanks!

I think adding it to the end is probably reasonable—later we can perhaps do a full-track audit like Ian Whitney did for Rust, as described here: http://designisrefactoring.com/2016/07/09/exercism-shouldnt-make-you-cry/

Leaving this for the track maintainers to review and merge :)

patricksjackson:
I finally got around to updating this to the new format.

@vivshaw, feel like reviewing one of mine?

vivshaw:
Alright, can do! Checked the test suite against the canonical data & saw no problems, so that's good. Stacks look like the most intuitive representation of the data, so that's good. I didn't spot any little `const`-correctness errors or anything like that. Lastly, the implementation is clear enough that C++ newcomers can easily grok it.

Looks good to me!

patricksjackson:
Thanks :)

PR #99, opened 2016-08-15
clock: Adding full test suite.

### LINKED ISSUE #78: clock: canonical test data has been improved
The JSON file containing canonical inputs/outputs for the Clock exercise has gotten new data.

There are two situations that the original data didn't account for:
- Sometimes people perform computation/mutation in the display method instead of in add. This means that you might have two copies of clock that are identical, and if you add 1440 minutes to one and 2880 minutes to the other, they display the same value but are not equal.
- Sometimes people only account for one adjustment in either direction, meaning that if you add 1,000,000 minutes, then the clock would not end up with a valid display time.

If this track has a generator for the Clock exercise, go ahead and regenerate it now. If it doesn't, then please verify the implementation of the test suite against the new data. If any cases are missing, they should be added.

See https://github.com/exercism/x-common/issues/166

patricksjackson:
I updated the test suite for clock to match the new JSON, and fixed the example solution to work with it.

Because there are a _lot_ of test cases for clock, I copied how [the Go track handles the test cases](https://github.com/exercism/xgo/tree/master/exercises/clock), placing the test cases into a vector and iterating over them. This might make it a little difficult to tell which test case is failing, but I believe the Boost output messages handle it well enough.

This is different from all of the other test suites for the cpp track (I think), so it may require some review/rework.

Closes #78

### DIFF
--- a/exercises/clock/clock_test.cpp
+++ b/exercises/clock/clock_test.cpp
@@ -4,81 +4,188 @@
 
 using namespace std;
 
-BOOST_AUTO_TEST_CASE(prints_the_hour)
+struct timeTest {
+    int hour;
+    int minute;
+    string expected;
+};
+
+vector<timeTest> timeCases = {
+    {8, 0, "08:00"},        // on the hour
+    {11, 9, "11:09"},       // past the hour
+    {24, 0, "00:00"},       // midnight is zero hours
+    {25, 0, "01:00"},       // hour rolls over
+    {100, 0, "04:00"},      // hour rolls over continuously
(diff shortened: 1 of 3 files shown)


kytrinyx:
One thing that we sometimes do in Go is to move the comments into a field in the vector and then use that to generate the error message when it fails. I have no idea if that would work here.

patricksjackson:
That is doable, and that solves the problem. Thanks :)

kytrinyx:
@LegalizeAdulthood this is ready for review when you have a moment.

PR #101, opened 2016-08-30
anagram: clarify solution requirements

### LINKED ISSUE #98: anagram problem description in C++ does not agree with tests
Is this issue: https://github.com/exercism/exercism.io/issues/2431 still relevant on the cpp track?

patricksjackson:
This exercise is the first that requires the solver to define a class. This patch helps users who may be unfamiliar with the syntax determine what is required.

Learners (especially those new to programming or new to programming with classes) may have difficulty determining a class is required for the existing tests. This gives them something to google if they don't know what's going on.

Closes #98

### DIFF
--- a/exercises/anagram/anagram_test.cpp
+++ b/exercises/anagram/anagram_test.cpp
@@ -6,7 +6,8 @@ using namespace std;
 
 BOOST_AUTO_TEST_CASE(no_matches)
 {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks @patricksjackson. I've pinged @LegalizeAdulthood to take a look at all of the open pull requests.

PR #102, opened 2016-08-30
gigasecond: use times (not dates)

### LINKED ISSUE #34: gigasecond: use times (not dates) for inputs and outputs
A duration of a gigasecond should be measured in seconds, not
days.

The `gigasecond` problem has been implemented in a number of languages,
and this issue has been generated for each of these language tracks.
This may already be fixed in this track, if so, please make a note of it
and close the issue.

There has been some discussion about whether or not gigaseconds should
take daylight savings time into account, and the conclusion was "no", since
not all locations observe daylight savings time.

patricksjackson:
Update gigasecond exercise to use times instead of dates.

Closes #34

### DIFF
--- a/exercises/gigasecond/example.cpp
+++ b/exercises/gigasecond/example.cpp
@@ -3,13 +3,9 @@
 namespace gigasecond
 {
 
-boost::gregorian::date advance(const boost::gregorian::date& start)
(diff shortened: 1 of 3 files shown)


patricksjackson:
Merging this to resolve the 2+ year old issue around the gigasecond test...

PR #105, opened 2016-10-15
replace 'problems' key with 'exercises' in config.json

cgrayson:
see #100


cgrayson:
I forgot to point out that as of [exercism/x-api#137](https://github.com/exercism/x-api/pull/137), it is safe to remove `problems`.

kytrinyx:
Thanks! :heart:

Issue #63, opened 2016-01-21
Move exercises to subdirectory


kytrinyx:
The problems api (x-api) now supports having exercises collected in a subdirectory
named `exercises`.

That is to say that instead of having a mix of `bin`, `docs`, and individual exercises,
we can have `bin`, `docs`, and `exercises` in the root of the repository, and all
the exercises collected in a subdirectory.

In other words, instead of this:

```
x{TRACK_ID}/
├── LICENSE
├── README.md
├── bin
│   └── fetch-configlet
├── bowling
│   ├── bowling_test.ext
│   └── example.ext
├── clock
│   ├── clock_test.ext
│   └── example.ext
├── config.json
└── docs
│   ├── ABOUT.md
│   └── img
... etc
```

we can have something like this:

```
x{TRACK_ID}/
├── LICENSE
├── README.md
├── bin
│   └── fetch-configlet
├── config.json
├── docs
│   ├── ABOUT.md
│   └── img
├── exercises
│   ├── bowling
│   │   ├── bowling_test.ext
│   │   └── example.ext
│   └── clock
│       ├── clock_test.ext
│       └── example.ext
... etc
```

This has already been deployed to production, so it's safe to make this change whenever you have time.

Ref: https://github.com/exercism/exercism.io/issues/3336

LegalizeAdulthood:
Does this only affect the xcpp repo or does it also affect what the user sees when they fetch exercises for a language track?

kytrinyx:
It only affects the xcpp repo -- everything else is unchanged.

masters3d:
@LegalizeAdulthood  Could I submit a PR for this with out breaking the build scripts?

LegalizeAdulthood:
You can always submit a pull request.  If it breaks the build scripts, I believe travis will show that when it runs an integration build on the branch.  We can look at the build log and see what (if anything) needs to be fixed.

petertseng:
Hello, since #108 is merged, let's close this!

kytrinyx:
Good call. Thanks!

PR #108, opened 2016-12-28
moved exercises into subdirectory

cyborgsphinx:
The issue is alive as #63, but it doesn't look like there's a corresponding pull request. Now there is.


kytrinyx:
Thanks @cyborgsphinx! This is the very last active track that needs to move the exercises. It will be great to all the code that deals with the deprecated exercise locations.

The build is failing—would you have a chance to take a look at that?

Insti:
I suspect the root `CMakelists.txt` file needs to be updated to reference the new exercise locations.

cyborgsphinx:
I agree that the `CMakeLists.txt` file needs updating. I've never actually used cmake before, aside from executing `cmake -G "Unix Makefiles"`, so I'm not sure how to make it work. I've tried a few things, like just modifying the working directory, to no real success.

patricksjackson:
[Something like this patch should work, though I didn't fully test it.](https://github.com/patricksjackson/xcpp/commit/62c7238dc3007462277f44ac5ae3ceb2fedf1ec1)

I can look up how to make a PR against this PR (?) tomorrow, but feel free to make those changes yourself if you get to it first.

cyborgsphinx:
That seems to work. Thanks.

patricksjackson:
LGTM 👍

kytrinyx:
I'm going to go ahead and merge this. If it needs improvements later it can be done in a separate pull request. Thanks all :heart: :heart:

PR #110, opened 2017-01-23
queen-attack patch

liquidcore7:
There is an issue in newer Boost versions: __require_equal_containers.h__ does not add the left-shift operator to std::pair. Changing **BOOST_REQUIRE_EQUAL(L, R)** to **BOOST_TEST(L == R)** as mentioned [here](https://github.com/exercism/xcpp/issues/91) isn`t working.


patricksjackson:
I was a bit behind on this, but thanks for the PR :)

PR #112, opened 2017-02-25
Pass hello-world

samsondstl:
Hey all, I am new to open source and Git, so if I apologize if I messed up this process somewhere on the way and feel free to let me know. Thanks.


config.json
```diff
@@ -14,6 +14,11 @@
     "point-mutations"
   ],
   "exercises": [
+	{
+		"difficult": 1,
+		"slug": "hello-world",
+		"topics": []
```
patricksjackson: Couple things here:

Make sure you turn off tabs and format this just like the rest of the file using spaces.

Spelling of `difficulty`

Also, please add "strings" to the `topics`. I know there aren't any for the other problems here, but we're supposed to be starting to add them :) You can look at https://github.com/exercism/xc/blob/master/config.json if you want to look at their formatting.

exercises/hello-world/example.cpp
```diff
@@ -0,0 +1,13 @@
+#include "hello_world.h"
+
+using namespace std;
+
+namespace hello_world
+{
+	
+std::string hello()
+{
+	return "Hello, World!";
```
patricksjackson: I think this is another tab, please change it to 4 spaces so the repo stays consisted :)

exercises/hello-world/example.cpp
```diff
@@ -0,0 +1,13 @@
+#include "hello_world.h"
+
+using namespace std;
+
+namespace hello_world
+{
+	
+std::string hello()
```
patricksjackson: You don't need `std::` since you're already using that namespace.

exercises/hello-world/hello_world_test.cpp
```diff
@@ -0,0 +1,12 @@
+#include "hello_world.h"
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+
+BOOST_AUTO_TEST_CASE(test_hello)
+{
+    BOOST_REQUIRE_EQUAL("Hello, World!", hello_world::hello());
+}
+
+#if defined(EXERCISM_RUN_ALL_TESTS)
```
patricksjackson: Take out this extra #if block, it's unnecessary.

patricksjackson:
This overall looks pretty good :) I just pointed out a few things to make sure the repo style stays consistent.

One other thing I'd like you to do is to rename "example.h" to "hello_world.h". That will make it get downloaded along with the tests so that people can see how the namespaces and header files work.

samsondstl:
@patricksjackson Excellent feedback! I went ahead and made the changes you suggested. I now have many commits tied to this pull request, is that okay? I thought I read that the project likes a single commit to a pull request, but I can't find that in the documentation, so it may simply be something I am dreamed up.

patricksjackson:
Thanks for making the changes :) I'll leave it up a couple days in case somebody else wants to give feedback, but it all looks good to me.

Multiple commits is perfectly fine (and almost always what happens). If it's useful to have the history of changes I can merge all of the commits, but I can also squash them into a single commit from Github's web interface.

Issue #111, opened 2017-01-28
I have to hack every c++ test case because of `<<` operator


Skarlso:
Hi. 

So, every test case I'm running throws this error into my face:

```bash
/usr/local/Cellar/boost/1.62.0/include/boost/test/tools/detail/print_helper.hpp:47:5: error: static_assert failed "Type has to implement operator<< to be printable"
    BOOST_STATIC_ASSERT_MSG( (boost::has_left_shift<std::ostream,T>::value),
```

I can see that there is a hack there called `#include "require_equal_containers.h"` but it's not working. 

Is there a solution for this to make it work or build it with some other version of boost? Or anything? It's becoming rather tedious to always delete it, and change it to `BOOST_EQUAL`.

Thanks.

Skarlso:
Okay, so I've got it, I think... 

I was using plain `cmake ..` instead of `cmake -G Xcode ..` because I thought it would detect my operating system. Afterwards, the case is, that I don't have `make` command anymore, so I have to use `cmake --build .`. Is this the intended usage?

Skarlso:
Nope, still says the same thing ( I just had a typo. ). So `cmake --build .` and using `-G Xcode` didn't help either.

patricksjackson:
Those are two different issues.

Using `cmake` is one. If you use `cmake -G Xcode ..` it will generate an Xcode project, and you will need to use Xcode to build it. I'm not sure if Unix Makefiles is the default generator or Mac, but calling it with `cmake -G "Unix Makefiles" ..` would select it.

The Boost issue is something that needs a fix in the repo. The `require_equal_containers` workaround breaks with newer versions of Boost for reasons I'm not entirely sure about. I'm planning to sort out a fix soon I've just been swamped with real work.

Skarlso:
Thanks for the response @patricksjackson. Sorry, I did not denote that I'm actually using osx here. So I need the xcode stuff. Also, after that I'm building with `cmake --build . ` thus it should be operating system agnostic.

sunzenshen:
I'm also affected by the Boost require_equal_containers issue as well, so let me know if I can help with reproduction or general triage.

sunzenshen:
@patricksjackson I found this post about the operator << override, and the proposed solution variant seems to work on my end:

http://bcreane.tumblr.com/post/148999817295/boost-unit-test-compilation-error

[...]

PR #118, opened 2017-04-16
Removing require_equal_containers.h

patricksjackson:
The use of require_equal_containers.h has been buggy for a while, and was totally broken for modern versions of Boost. This removes that and replaces it with builtin boost test macros. The error messages take a slight hit from what used to be there*, but this makes it actually work without having to maintain an undocumented hack of Boost.Test.

I will be trying to test this with different versions of Boost using Travis, as I'm not sure how many versions of Boost this supports. As long as it works with at least Boost 1.59 I think this is a good improvement, though the requirements for the repo/docs will need to be updated (they currently require only Boost 1.55).

*Error messages in all of the 'containers of containers' will only report that the two containers are non-equal, and not _what_ is different about them. It's a small loss to usability.


patricksjackson:
Ok, Travis has been updated to pass, and the documentation is updated as well.

I will be merging this in a week unless someone raises a problem. This is a long-standing problem for the repo that made many of the tests unusable for anyone on a recent distro.

See #110, #111, #91.

sunzenshen:
Branch works OK for me with: 

- OS X El Capitan 10.11.6
- AppleClang 8.0.0.8000038
- Boost version 106300

patricksjackson:
I took another look through the documentation today, and added additional details around how to acquire Boost 1.59. I expect that after this change we may see a few people coming in who have a broken install (i.e. they've been using 1.55 and will then get a failed build once they get a new version), but it should not be that many people.

I expect that the vast majority of people are already using Boost 1.59+, as it is about 2 years old and every major Linux distro has updated by now (and the install process for Windows/Mac usually has people install the latest version). The notable exception is Ubuntu 16.04 LTS, which uses Boost 1.58. I've added explicit install instructions for that distro.

Skarlso:
Awesomesauce!

PR #119, opened 2017-04-17
Let first 'make' fail for the right reason

objarni:
Currently, first make of the problem "etl" fails because the etl_tests.cpp file uses std::map without including <map>. Adding #include <map> fixes it. (Another way to "fix" is to include map inside of etl.h, but that feels like a "hack". The unit tests should include type declarations for types it's using itself! :) )

### DIFF
--- a/exercises/etl/etl_test.cpp
+++ b/exercises/etl/etl_test.cpp
@@ -1,5 +1,6 @@
 #include "etl.h"
 #define BOOST_TEST_MAIN
+#include <map>
 #include <boost/test/unit_test.hpp>
 #include "require_equal_containers.h"
 


patricksjackson:
I think the current implementation is fine. It does fail to compile until you get an accurate function declaration in the header, which is pretty normal for many of the later exercises. The function declaration uses std::map (and std::vector), so those would need to be included in etl.h.

How bad was the initial experience without this include?

objarni:
It was a difference from previous exercises' unit tests, which failed because of missing user types, not errors in the tests themselves.

I think it is also good style to include types you are using in translation units, not rely on includes to include the types you are using.

patricksjackson:
Yea, I think you're right. It would generally be better practice to do that.

I'm going to go ahead and merge this. I expect that you will find this to be an issue with a number of future problems as well, so feel free to submit a PR for them if you like.

Thanks :)

objarni:
Happy to help out :)

I did notice same problem with following exercise, I'll try and get that in a PR too then ...

PR #121, opened 2017-04-24
Include map in two tests using map, but not including it

objarni:
Added #include \<map\> to another two unit tests lacking it.

### DIFF
--- a/exercises/grade-school/grade_school_test.cpp
+++ b/exercises/grade-school/grade_school_test.cpp
@@ -1,5 +1,6 @@
 #include "grade_school.h"
 #define BOOST_TEST_MAIN
+#include <map>
 #include <boost/test/unit_test.hpp>
 #include "require_equal_containers.h"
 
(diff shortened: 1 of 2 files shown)


exercises/grade-school/grade_school_test.cpp
```diff
@@ -1,5 +1,6 @@
 #include "grade_school.h"
 #define BOOST_TEST_MAIN
+#include <map>
```
patricksjackson: It looks like this file also needs `vector` and `string` included.

patricksjackson:
Thanks for doing this! (Sorry it took a bit to come in to review)

Looks like there actually are a few things other than map that need to be included as well.

exercises/grade-school/grade_school_test.cpp
objarni: Nice spot

objarni:
Yes, I only searched for std::map in test/cpp files, haven't check vector or string. Up for it :) ?

patricksjackson:
I would feel better about this if we updated all of the other necessary includes as well.

If I get around to it I can do the other changes myself if you don't want to.

objarni:
Sure, but what types is there to look for..?
Map and vector of course, but what more? I have only implemented the first few exercises of exercism C++ track so I don't really know what is used, and looking throw all test files kind of spoils the fun of exercism a bit ...

If I get an exhaustive list of types, I can easily fix this.

patricksjackson:
Off the top of my head it's probably just vector and string, but I really don't know without looking through them all. 

I will just go ahead and merge this one. Thanks :)

PR #122, opened 2017-05-04
Add `apt-get update` to boost install

dragondave:
Didn't work for me without adding it. Copied from g++ installation docs further up page.


patricksjackson:
Oops, my mistake. Thanks for the fix :)

Out of curiosity, does this mean you're on the LTS build of Ubuntu? I couldn't figure out how common that was for people who would be doing active development.

dragondave:
Yes; I believe 16.04 is the most recent supported version for the version
of Linux that's integrated with Windows 10.

patricksjackson:
Ah right, the Creator's Update does use 16.04. I've been using that lately as well.
Thanks!

PR #127, opened 2017-05-31
Move bob forward in the config.json

stevejb71:


stevejb71:
bob is considerably harder than leap, but is before it in the config.json, meaning people fetch it earlier.

patricksjackson:
That place works for me. Thanks!

PR #131, opened 2017-06-19
[WIP] Improve robot-name tests.

patricksjackson:
Improves the tests for robot-name (making sure that it can generate at least 1000 unique robot names), and transition to regex from the std library instead of Boost.

*This may break in Travis, as there were supposedly some issues with the regex implementation in old versions of the stdlib. Currently this is a WIP to test against that passing.

I will update this PR to remove boost regex as a dependency for all the tests if there are no issues.


patricksjackson:
Nevermind, GCC didn't implement regex until version 4.9 (Travis has 4.8, and I'm not up for updating the compiler dependency yet.)

PR #132, opened 2017-06-24
Port allergies exercise

vivshaw:
First time porting one over; let me know if I need to fix anything / if I made some style errors & I'll get them fixed.

Wasn't sure where best to put it in config.json so it's at the end.

Also, would you prefer I squashed it all to 1 patch, or leave the commits separate like this?

### DIFF
--- a/exercises/allergies/allergies_test.cpp
+++ b/exercises/allergies/allergies_test.cpp
@@ -1,0 +1,22 @@
+#include "allergies.h"
+#include <stdexcept>
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+
+BOOST_AUTO_TEST_CASE(has_no_allergies)
+{
+    const allergies::allergy_test test(0);
+    
+    BOOST_REQUIRE_EQUAL(False, test.is_allergic_to("Peanuts"));
+    BOOST_REQUIRE_EQUAL(False, test.is_allergic_to("Cats"));
+    BOOST_REQUIRE_EQUAL(False, test.is_allergic_to("Strawberries"));
+}
+
+#if defined(EXERCISM_RUN_ALL_TESTS)
+BOOST_AUTO_TEST_CASE(allergic_to_eggs)
+{
+   const allergies::allergy_test test(1);
+   
+   BOOST_REQUIRE_EQUAL(True, test.is_allergic_to("Strawberries"));
+}
+#endif
--- a/exercises/allergies/allergies_test.cpp
+++ b/exercises/allergies/allergies_test.cpp
@@ -2,6 +2,9 @@
 #define BOOST_TEST_MAIN
 #include <boost/test/unit_test.hpp>
 
+#include <string>
+#include <set>
+
 BOOST_AUTO_TEST_CASE(has_no_allergies)
 {
     allergies::allergy_test test(0);
@@ -74,6 +77,22 @@ BOOST_AUTO_TEST_CASE(allergic_to_eggs_and_other_stuff)
     BOOST_REQUIRE_EQUAL(true, test.is_allergic_to("eggs"));
     BOOST_REQUIRE_EQUAL(true, test.is_allergic_to("shellfish"));
     BOOST_REQUIRE_EQUAL(false, test.is_allergic_to("peanuts"));
+}
+
+BOOST_AUTO_TEST_CASE(allergic_to_nothing)
+{
+    allergies::allergy_test test(0);
+    const std::set<std::string> no_allergies;
+
+    BOOST_TEST(no_allergies == test.get_allergies());
+}
+
+BOOST_AUTO_TEST_CASE(allergic_to_only_peanuts)
+{
+    allergies::allergy_test test(2);
+    const std::set<std::string> only_peanuts = {"peanuts"};
+
+    BOOST_TEST(only_peanuts == test.get_allergies());
 }
 
 
--- a/exercises/allergies/allergies_test.cpp
+++ b/exercises/allergies/allergies_test.cpp
@@ -3,7 +3,7 @@
 #include <boost/test/unit_test.hpp>
 
 #include <string>
-#include <vector>
+#include <unordered_set>
 
 BOOST_AUTO_TEST_CASE(no_allergies_means_not_allergic)
 {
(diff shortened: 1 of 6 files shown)


exercises/allergies/allergies_test.cpp
```diff
@@ -0,0 +1,148 @@
...
+    allergies::allergy_test score(2);
+   
+    BOOST_REQUIRE_EQUAL(true, score.is_allergic_to("peanuts"));
+}
+
+BOOST_AUTO_TEST_CASE(allergic_to_shellfish)
+{
+    allergies::allergy_test score(4);
+   
+    BOOST_REQUIRE_EQUAL(true,score.is_allergic_to("shellfish"));
```
patricksjackson: Nitpick - it's missing a space in the call. `(true, score...`

patricksjackson:
The config.json looks perfect. There are plans to reconfigure all of the track progressions, so I'm not going to worry about setting them in exactly the right order just yet.

I can squash all the commits when I merge it (it's a button in github) so don't worry about squashing any of them.

When I look at these PRs I care a lot more about the tests and the interface than I do about your implementation. I'll probably give you some comments about the implementation anyways, but as long as it's pretty reasonable it's not generally a problem.

The test suite looks totally fine except for one thing I think we should think about.

You use a vector<string> for getting and comparing the full list of allergies. The problem is that when you compare these lists together they must be in the same order, so if an implementation wants to return them in a different order (maybe they generate the list using the highest bit first) then this won't work.

There are two solutions I can think of:
1. Write some code in the test suite to normalize the vectors (sort them maybe) before you do the compare.
2. Have the `get_allergies()` function return a `set<string>` or `unordered_set<string>`.

Personally I prefer solution number 2. What do you think?

exercises/allergies/example.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include <string>
+#include <map>
+#include <vector>
+#include <algorithm>
+
+namespace allergies
+{
+
+allergy_test::allergy_test(const int test_result)
```
patricksjackson: The `const` in this function definition doesn't have any real effect since it's passed by value instead of reference. It works fine, but that's not usually the convention.

exercises/allergies/example.cpp
```diff
@@ -0,0 +1,46 @@
...
+    : result(test_result)
+    , allergens({
+    	{"eggs", 1},
+	{"peanuts", 2},
+	{"shellfish", 4},
+	{"strawberries", 8},
+	{"tomatoes", 16},
+	{"chocolate", 32},
+    	{"pollen", 64},
+	{"cats", 128}
```
patricksjackson: This is a little weird. For every one of these objects you create an identical copy of this map object inside of it.

What you _could_ do instead is create a `const` global object in an anonymous namespace in this file. Then you'd only ever have one copy, and the objects would be small.

(Also, the formatting here is a bit messed up)

exercises/allergies/example.cpp
```diff
@@ -0,0 +1,46 @@
...
+   })
+{} 
+
+bool allergy_test::is_allergic_to(std::string const& allergen) const
+{
+    int allergen_value = allergens.at(allergen);
```
patricksjackson: What happens if this string is not in the `map`?

exercises/allergies/example.cpp
```diff
@@ -0,0 +1,46 @@
...
+    return (result & allergen_value) == allergen_value;
+}
+
+std::vector<std::string> allergy_test::get_allergies() const
+{
+    std::vector<std::string> allergy_list({"eggs", "peanuts", "shellfish", "strawberries",
+		    			   "tomatoes", "chocolate", "pollen", "cats"});
```
patricksjackson: Instead of listing all these strings twice (here and in the map), do you think there's a way you can only list them once? Maybe by reading them from the map object here? Then, if you ever needed to update the names or what they mapped to you'd only have to change it in one place.

exercises/allergies/example.cpp
```diff
@@ -0,0 +1,46 @@
...
+    auto is_not_allergic = [this](std::string allergen){return !(this->is_allergic_to(allergen));};
+
+    allergy_list.erase(
+        std::remove_if(allergy_list.begin(),
+		       allergy_list.end(),
+		       is_not_allergic),
+	allergy_list.end());
```
patricksjackson: Formatting?

I'd align this line with the `std::remove_if`. It's possible you've mixed tabs and spaces in this doc.

exercises/allergies/example.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    int allergen_value = allergens.at(allergen);
+    return (result & allergen_value) == allergen_value;
+}
+
+std::vector<std::string> allergy_test::get_allergies() const
+{
+    std::vector<std::string> allergy_list({"eggs", "peanuts", "shellfish", "strawberries",
+		    			   "tomatoes", "chocolate", "pollen", "cats"});
+    auto is_not_allergic = [this](std::string allergen){return !(this->is_allergic_to(allergen));};
```
patricksjackson: The argument for the lambda could be labeled `const&`.

vivshaw:
Thanks for the feedback, that was immensely helpful! Working on fixes now, hope to push by tomorrow.

As for the test suite, I agree that vector semantics are incorrect here; I only used them because [the canonical test data](https://github.com/exercism/problem-specifications/blob/master/exercises/allergies/canonical-data.json) said it expected a list, and all the other language tracks I checked did in fact use some sort of list-esque data structure rather than a set. Personally, I'd love to reimplement with std::unordered_set<string> if that's all right; it'd make things so much simpler.

exercises/allergies/example.cpp
vivshaw: Wanted to note re: this one in particular, this happened because of trying to shoehorn this into a vector when it shouldn't be. Problem is, std::map is sorted by key, so when iterating over the map's keys to produce the allergen vector, said vector comes back in a different order than the test cases. AFAIK, solving this in a semi-reasonable manner while sticking with vectors would require either normalizing the vector in each test case or moving to boost::bi_map. 

One more reason to prefer using a set, no?

patricksjackson:
Each of the tracks has some leeway around not implementing _exactly_ the format of the canonical tests. (I also have several issues with how some of them are made anyways).

In this case, it needs to return an unordered list of strings, so that sounds like an `unordered_set` to me 👍

patricksjackson:
Looks good to me :) Thanks!

PR #134, opened 2017-07-13
Port atbash-cipher exercise

vivshaw:
Here's a port of the atbash-cipher exercise. Was originally gonna do the example implementation with some tidy modern c++ std::erase_if and std::transform stuff until I realized it'd take 3 passes over the string, yuck. So, for-loops in 1 pass it is. Let me know if anything needs changed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,6 +48,7 @@ foreach(exercise
     meetup
     queen-attack
     allergies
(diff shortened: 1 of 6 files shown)


patricksjackson:
I probably would have done a few things different in the implementation, but those would mostly be nitpicks or my personal taste really.

Looks great to me, thanks!

PR #143, opened 2017-08-22
Add default introductory code example

kytrinyx:
Instead of having logic for a fallback in the backend, we're choosing a
default for the snippet file.

For tracks that have core exercises, we pick the first core exercise.
For tracks without these, we pick the first exercise listed in the config.

Note that we're aiming for 10 lines and a max-width of 40 columns.
This solution has 13 lines, so we may want to adjust it.

See https://github.com/exercism/meta/issues/89

### DIFF
--- a/docs/SNIPPET.txt
+++ b/docs/SNIPPET.txt
@@ -1,0 +1,13 @@
+#include "hello_world.h"
+
+using namespace std;
+
+namespace hello_world
+{
+	
+string hello()
+{
+    return "Hello, World!";
+}
+
+}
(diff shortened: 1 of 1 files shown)


patricksjackson:
I adjusted the number of lines down to 10, the excess was only due to some coding conventions in this repo.

I don't really like using Hello World as "The C++ snippet" since it's all boilerplate, but it's a good default for the moment.

kytrinyx:
> I don't really like using Hello World as "The C++ snippet" since it's all boilerplate, but it's a good default for the moment.

Agreed—it's just so we have _something_ to show. Can always improve later!

PR #148, opened 2017-08-28
Implementing pangram.

Is0metry:
description to come

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,436 +1,436 @@
 {
+  "language": "C++",
   "active": true,
-  "exercises": [
-    {
-      "core": false,
-      "difficulty": 1,
-      "slug": "hello-world",
-      "topics": [
-        "strings"
-      ],
-      "unlocked_by": null,
-      "uuid": "3eeadff6-82b6-43be-8834-d9ef258e454d"
-    },
-    {
-      "core": false,
-      "difficulty": 1,
-      "slug": "leap",
-      "topics": [
-        "control-flow-(conditionals)",
-        "logic"
-      ],
-      "unlocked_by": null,
-      "uuid": "76a840c7-24f1-455d-b62e-da42b13f8dd5"
-    },
-    {
-      "core": false,
-      "difficulty": 2,
-      "slug": "gigasecond",
-      "topics": [
-        "dates",
-        "interfaces"
-      ],
-      "unlocked_by": null,
-      "uuid": "4dbe1bb3-0419-4749-8c2a-ec91539a8640"
-    },
-    {
-      "core": false,
-      "difficulty": 3,
-      "slug": "hamming",
-      "topics": [
-        "control-flow-(loops)",
-        "strings"
-      ],
-      "unlocked_by": null,
-      "uuid": "a9716747-8cc6-4529-b5d2-dc24bd8641a8"
-    },
-    {
-      "core": false,
-      "difficulty": 5,
-      "slug": "bob",
-      "topics": [
-        "control-flow-(conditionals)",
-        "parsing",
-        "strings"
-      ],
-      "unlocked_by": null,
-      "uuid": "d80210b8-45e8-4e5c-8b07-9e87c33c5959"
-    },
-    {
-      "core": false,
-      "difficulty": 5,
-      "slug": "anagram",
-      "topics": [
-        "arrays",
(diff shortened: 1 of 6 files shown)


patricksjackson:
This is pretty good :)

The test suite is complete, and the implementation is good enough I won't bother nitpicking it.

What you'll need to do before I can merge it is finish matching the style in this repo and adding it to our build scripts. So a short list of things you'll need to fix:

- Please don't re-format the config.json. Just add the lines necessary for this new problem.
- You'll need to add the name to the list in the root [CMakeLists.txt](https://github.com/exercism/cpp/blob/master/CMakeLists.txt), otherwise it won't build during our automated testing.
- Try to match the formatting of existing tests. Take a look at [Word Count](https://github.com/exercism/cpp/tree/master/exercises/word-count) in detail, it has similar requirements to this one. Specifically:
1. Formatting, specifically vertical spacing (blank lines).
2. Name of the `#define` in the header.
3. Placing the helper function in an anonymous namespace and not exposing it in the header.
4. Including only what you need in the header file, and moving the others to the cpp (`cctype`).

Is0metry:
changes implemented :)

I didn't consciously reformat the config (I ran everything through the configlet which might've done it), so I rolled back to a previous version and just added the necessary lines. It **should** be back to normal again.

config.json
```diff
@@ -423,6 +423,14 @@
         "pattern matching",
         "stacks"
       ]
+    },
+    {
+      "slug":"pangram",
```
patricksjackson: This needs a couple more fields (the `uuid`, which configlet can generate for you, `core` and `unlocked_by` (just leave the values the same default as the others.

exercises/pangram/example.h
```diff
@@ -0,0 +1,10 @@
+#ifndef PANGRAM_H
```
patricksjackson: `#if !defined(PANGRAM_H)` to match all the other problems.

exercises/pangram/example.h
```diff
@@ -0,0 +1,10 @@
...
+#define PANGRAM_H
+#include <string>
+
+namespace pangram
+{
+    
+   bool is_pangram(std::string s);
```
patricksjackson: The repo convention is to not indent namespace contents. So this whole namespace should be:

```
namespace pangram
{

bool is_pangram(std::string s);

}
```

exercises/pangram/example.cpp
```diff
@@ -0,0 +1,30 @@
+#include "example.h"
+#include <cctype>
+namespace
+{
+    bool found_letter(char a, std::string s)
```
patricksjackson: Please add some extra vertical spacing (blank lines between the namespaces/functions/etc (see [word-count](https://github.com/exercism/cpp/blob/master/exercises/word-count/example.cpp))

And don't indent inside of the namespace block.

config.json
Is0metry: What command do I run on the configlet to get it to add the `uuid`? I've tried `fmt` and `lint` and neither seems to be doing it.

patricksjackson:
Thanks :) I pointed out a few more details.

Thanks for letting me know the configlet tool was the one that reformatted the json. I'd like to update the formatting of that with whatever configlet does, but that should be done in a separate commit.

(Also, still need the CMakeLists.txt update I mentioned, though I know you may be working on it :) )

Is0metry:
That should be just about everything. Thanks for your patience. I'm still kinda new to this :).

config.json
patricksjackson: I should clarify this - `configlet uuid` will generate the uuid string for you, but you'll need to edit the json manually

exercises/pangram/example.h
```diff
@@ -0,0 +1,11 @@
+#if !defined(PANGRAM_H)
+#define PANGRAM_H
+#include <string>
+
+namespace pangram
+{
+    
+bool is_pangram(std::string s);
```
patricksjackson: Actually please make the string arguments for both of these functions take `std::string const& s`.

I missed this on my first pass...however this and the config.json update are the only things left :)

Is0metry: latest push should have everything.

patricksjackson:
Awesome! Thanks.

Issue #149, opened 2017-08-30
gigasecond/gigasecond_test.cpp include header?


lkskstlr:
This is a question:

Wouldn't it be nicer to include "boost/date_time/posix_time/posix_time.hpp" into the gigasecond_test.cpp as it actually needs it, instead of implicitly relying on gigasecond.h to include it. Maybe one wants to keep the header slim and thus forward declares boost::posix_time::ptime and includes the full boost header in the cpp? Additionally there is "boost/date_time/posix_time/posix_time_types.hpp" which has no i/o and would be sufficient for the implementation of advance() but then the test doesn't compile.

Please correct me if I am wrong :)

Cheers, Lukas

lkskstlr:
Thanks for the references. Should have checked myself before opening an issue. I think that a test should not rely on the solution header to include another header which is actually needed by the test itself.
If posix_time.hpp was very large, it would actually be bad software design to include it in gigasecond.h only to use a type that can be forward declared easily. On the other hand this exercise is at the beginning of the track and the users probably don't think about such things anyway.
I understand the point that giving the user the header is kind of the solution already, but I think that good design practice is more important. I am however definitely no expert on this.

As you stated in #121 it would really make sense to be consistent on this for all exercises. 

If you are interested I might have time to do this for all exercises in about 2-3 weeks. But I don't think this is a big issue right now. What do you think?

Gamecock:
I have a related issue, not sure if it's just me or an actual issue.
```/Users/Finch/projects/exercism/cpp/gigasecond/gigasecond_test.cpp:23:26: error: 
      use of undeclared identifier 'time_from_string'```
I can fix it by adding this header to the test case.  `#include <boost/date_time.hpp>`
The header belongs in the test case, not the implementation.  I dont understand, because it looks like it's part of posix_time [time_from_string](http://www.boost.org/doc/libs/1_65_1/doc/html/date_time/posix_time.html#ptime_from_string).
Let me know if:
I should fix on this issue.
I should create a new issue.
I'm doing something else wrong and should not touch the repor.

Thanks

Gamecock:
I have a related issue, not sure if it's just me or an actual issue.
```/Users/Finch/projects/exercism/cpp/gigasecond/gigasecond_test.cpp:23:26: error: 
      use of undeclared identifier 'time_from_string'```
I can fix it by adding this header to the test case.  `#include <boost/date_time.hpp>`
[...]

PR #150, opened 2017-08-30
Included posix_time header in test. Changed Example accordingly (#149).

lkskstlr:
I just wanted to propose a solution for the case that you might consider the issue I opened to be valid. I hope this was somehow helpful to you.

Cheers,
Lukas

### DIFF
--- a/exercises/gigasecond/example.h
+++ b/exercises/gigasecond/example.h
@@ -1,7 +1,9 @@
 #if !defined(GIGASECOND_H)
 #define GIGASECOND_H
 
-#include <boost/date_time/posix_time/posix_time.hpp>
+namespace boost{ namespace posix_time{
+class ptime;
+} }
 
 namespace gigasecond
 {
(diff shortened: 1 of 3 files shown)


exercises/gigasecond/example.h
```diff
@@ -1,7 +1,9 @@
 #if !defined(GIGASECOND_H)
 #define GIGASECOND_H
 
-#include <boost/date_time/posix_time/posix_time.hpp>
```
patricksjackson: We really shouldn't bother removing this here. It doesn't save anything since it always must be included to use this header anyways.

Yes it might be against some companies guidelines, but we're going for simplicity here instead of strictly following some definition of best practice. These tests just aren't going to do much to help teach best practices on a large Cpp codebase.

patricksjackson:
Thanks for this.

If you add "Closes #149" to the description (first comment?) or in a commit message it will auto close the issue once this is merged. See https://help.github.com/articles/closing-issues-using-keywords/

I'm not sure if it works for a comment like this.

patricksjackson:
Closing in favor of #201

PR #151, opened 2017-09-06
Implement exercise all-your-base

baskeboler:


config.json
```diff
@@ -1,428 +1,434 @@
...
         "stacks"
-      ]
+      ],
+      "unlocked_by": null,
+      "uuid": "ce9fd8ac-0beb-1a80-4ff6-be19ce1540d905c31b7"
+    },
+    {
+      "difficulty": 2,
+      "slug": "all-your-base",
+      "topics": null,
```
patricksjackson: Please add some topics here. The list is at https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt

patricksjackson:
You used configlet to reformat the config.json right? We need to do that in a separate commit instead of along with an exercise.

exercises/all-your-base/all_your_base_test.cpp
```diff
@@ -0,0 +1,175 @@
+#include "all_your_base.h"
+
```
patricksjackson: Super-nitpick: I think all the problems don't have any whitespace in the top includes, so please remove this line for consistency.

exercises/all-your-base/example.h
```diff
@@ -0,0 +1,13 @@
+#ifndef ALL_YOUR_BASE_H
+#define ALL_YOUR_BASE_H
+
+#include <vector>
+#define EXERCISM_RUN_ALL_TESTS
```
patricksjackson: You don't need this.

exercises/all-your-base/example.h
```diff
@@ -0,0 +1,13 @@
...
+
+namespace all_your_base {
+
+    const std::vector<unsigned int> convert(unsigned int input_base, const std::vector<unsigned int>& input_digits, unsigned int output_base);
```
patricksjackson: A `const vector` return value is just odd, and I don't know of a legitimate reason for it. Please change to just a `vector`.

The formatting for the `const` input should be `std::vector<unsigned int> const& input_digits` to match the style

patricksjackson:
You need to add this problem to the list in the root CMakeLists.txt so that it builds during CI. (And I need to update the README with better instructions about that.)

Formatting in the test file needs to be updated to match other tests. Everything should be indented with 4 spaces and the vertical spacing needs to be consistent with the other exercises. This is the file the user sees, so keeping it consistent between exercises is important.

Same with your example files as well, please match formatting with other tests. There's no indenting in namespaces, the `#ifdefs` in the header needs changing, if and for blocks are formatted differently, etc.

patricksjackson:
By the way, I was curious whether or not it was worth it to use a linked list in implementing this. For most operations on small lists and even with large lists and small elements linked lists are horrible compared to vectors.

[Here's a chart I made comparing a few different methods up through lists of 16K elements](https://imgur.com/a/GsXHE). The Linked List only becomes better than even just pushing to the front of a vector after you have 1K elements per list, and it is ~30x slower than inserting to the back of a vector and then reversing it.

baskeboler:
@patricksjackson I used clang-format for formatting. Maybe it would be a good idea to include a .clang-format file in the project root.

exercises/all-your-base/all_your_base_test.cpp
```diff
@@ -0,0 +1,170 @@
+#include "all_your_base.h"
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+
+using namespace std;
+namespace utf = boost::unit_test;
```
patricksjackson: Does this serve any purpose?

exercises/all-your-base/example.h
```diff
@@ -0,0 +1,13 @@
+#ifndef ALL_YOUR_BASE_H
```
patricksjackson: Nitpick: this should be `#if !defined(ALL_YOUR_BASE_H)` to match the other exercises

patricksjackson:
Sorry it took me a while to get back to this.

I updated the config.json in the master branch. So you should now be able to rebase and fix the conflicts. That way we have the config reformatting as a separate commit than an exercise PR.

Similarly, I am very happy to talk about setting a formatting file for the entire repo, but not in an exercise PR. That should be done separately.

Other than that it looks good outside of a few nitpicks. Thanks.

patricksjackson:
I manually did the rebase and some nitpick fixes. Thanks!

baskeboler:
awesome!

PR #152, opened 2017-09-10
Added a test that checks that the roster is not changed after queryin…

j0ran:
This test will check that you do not change the roster by adding an empty grade when querying for a grade. I think this test will make the exercise better and that you can learn something from it (const methods).

### DIFF
--- a/exercises/grade-school/grade_school_test.cpp
+++ b/exercises/grade-school/grade_school_test.cpp
@@ -86,6 +86,13 @@ BOOST_AUTO_TEST_CASE(the_student_names_in_each_grade_in_the_roster_are_sorted)
         };
     BOOST_TEST(expected == actual);
 }
+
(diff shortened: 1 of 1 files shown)


patricksjackson:
That seems oddly implementation specific...but it is an edge case I guess. This also doesn't _really_ enforce `const` methods, the real way would be to have a `const grade_school::school` object and call grade on that.

I think this would be better if you just added that check to the other test case `grade_returns_an_empty_array_if_there_are_no_students_in_that_grade`

j0ran:
It is maybe implementation specific, but many submissions change the roster because they implement it as ```return roster[grade];```, which changes the roster map by inserting an empty vector into it.

I still propose to add it as a separate test because it states in its name it is testing this edge case.

patricksjackson:
Makes sense, I didn't realize how easy that issue would show up.

Thanks.

PR #153, opened 2017-09-15
Update steps to run tests.

Gamecock:
In TDD fashion I made the smallest possible change that would have prevented me from having to ask for help on chat.  Then I noticed a lot of common words and refactored it out.

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -57,7 +57,7 @@ code using the appropriate command for your environment:
 
 Examples of running CMake for different environments are shown below.
 
-#### Linux with Make
+#### Linux\MacOS with Make
 
 The generator name for CMake is `Unix Makefiles`.
 Assuming the current exercise is `bob` and we're in the exercise folder:
@@ -71,6 +71,8 @@ $ cmake -G "Unix Makefiles" ..
 
 This example shows creating empty files for the implementation before
 running CMake.
+
+To run the tests simply type `make` in the `build` directory
 
 #### Windows with Visual Studio 2013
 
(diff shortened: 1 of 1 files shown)


docs/TESTS.md
```diff
@@ -72,6 +72,8 @@ $ cmake -G "Unix Makefiles" ..
 This example shows creating empty files for the implementation before
 running CMake.
 
+To run the tests simply type `make` in the `build` directory
```
patricksjackson: The rest of this file uses full sentences, so please add a period to the end here.

Also, is this actually a good enough explanation? There's been some confusion around how to execute the tests in the past, and I honestly don't know if we need to clarify that they run automatically after compilation or if this is clear enough.

Gamecock: This the help I needed from gitter when I couldn't figure it out from the README, possibly the output from other OSs isn't as clear.
```
$ make 
Scanning dependencies of target anagram
[ 33%] Building CXX object CMakeFiles/anagram.dir/anagram_test.cpp.o
.../anagram/anagram_test.cpp:10:5: error: use of
      undeclared identifier 'anagram'
    anagram::anagram subject = anagram::anagram("diaper");
    ^
.../anagram/anagram_test.cpp:10:32: error: use of
      undeclared identifier 'anagram'
    anagram::anagram subject = anagram::anagram("diaper");
                               ^
2 errors generated.
make[2]: *** [CMakeFiles/anagram.dir/anagram_test.cpp.o] Error 1
make[1]: *** [CMakeFiles/anagram.dir/all] Error 2
make: *** [all] Error 2
$ 
```
I could change to something like "Simply type `make` in the `build` directory to compile the tests. This should generate compile time errors.  Once the errors are fixed `make` will build and run the tests."  We already have the basic cycle of TDD described in a couple of places, I don't think we need more.  If you can think of something else let me know.

patricksjackson: Yea, I mean literally just pointing out what `make` does. i.e. just saying that it builds and runs the tests and what the output might be (build errors or test pass/fail messages) to avoid people not knowing what to expect. Doesn't have to be a whole lot.

The output from other OS's will be identical, btw.

patricksjackson:
Thanks.

patricksjackson:
I could've sworn I merged this yesterday...oops.

PR #156, opened 2017-10-15
Improve robot-name tests and remove Boost.Regex dependency.

patricksjackson:
Improves the tests for robot-name (making sure that it can generate at least 1000 unique robot names), and transition to `std::regex` instead of Boost.Regex.

The use of Boost.Regex only existed because GCC hadn't instituted `std::regex` support at the time of original writing.


patricksjackson:
Closing in deference to #159, which includes this change

PR #158, opened 2017-10-28
Implement exercise acronym

chgraef:


exercises/acronym/example.h
```diff
@@ -0,0 +1,12 @@
+#if !defined(ACRONYM_H)
+#define ACRONYM_H
+
+#include <string>
+
+namespace acronym {
+
+const std::string acronym(std::string const&);
```
patricksjackson: I'm fairly sure that returning a const value is not useful in this situation, I can't think of very many situations where it would be useful. You probably want to return just `std::string`, though if there's a reason for the const I'd love to know about it :)

exercises/acronym/example.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace acronym {
+
+const string acronym(string const& input_str)
+{
+    vector<string> words;
+    boost::split(words, input_str, boost::is_any_of(" :-,"));
+
+    string acronym_str;
+    for(vector<string>::iterator it = words.begin(); it != words.end(); ++it)
```
patricksjackson: You probably want to use a ranged for loop here. [See here, especially the examples.](http://en.cppreference.com/w/cpp/language/range-for)

I think you want:
```
for (auto const& s: words)
    acronym_str.push_back(to_upper(s.front()));
```

exercises/acronym/acronym_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+
+using namespace std;
+
+BOOST_AUTO_TEST_CASE(basic)
+{
+    const string expected{"PNG"};
+
+    const string actual = acronym::acronym("Portable Network Graphics");
```
patricksjackson: I think the `expected` and `actual` definitions should be swapped for readability. When I go through and read the tests, it's more logical for me to read it as "The acronym of Portable Network Graphics is expected to be PNG", rather than "I expect to get PNG as the acronym for Portable Network Graphics"

I literally just mean reordering the tests to be:
```
const string actual = ...;

const string expected{...};

BOOST_TEST(expected == actual);
```

Make sense?

patricksjackson:
This looks very good. I pushed a couple of tiny nitpicks just to match the oddities in the repo (adding to the central CMakeLists.txt get's it to build and test on Travis CI).

I also had a couple other fairly minor suggestions that you should look at and consider. Overall though, this is pretty solid.

exercises/acronym/example.h
chgraef: Agreed! :) Will fix this.

exercises/acronym/example.cpp
chgraef: Yeah, my version is fairly verbose, admittedly. I will change this to a range-for instead, as you suggested.

exercises/acronym/acronym_test.cpp
chgraef: Sure, will do. I borrowed the skeleton of my tests from the word-count exercise but had noticed that in some other exercises the tests are structured exactly the way you're suggesting.

chgraef:
Hi @patricksjackson,

I was wondering, is there anything else needed before the exercise can be merged?

chgraef:
Hi @patricksjackson @LegalizeAdulthood 

can you review, please?

patricksjackson:
I rebased to bring this up to date with a couple repo changes. Merging.

Thanks!

PR #159, opened 2017-10-29
WIP: Remove Boost as a required dependency

### LINKED ISSUE #96: Setting up C++ dev environment
The error that was demo'd on the Installing C++ tab has happened to me, even though I followed the instructions to the T. Stack Overflow as been less than helpful as well.

### LINKED ISSUE #116: Boost unit tests on Fedora
I had problems with Boost on Fedora and these exercises: by default, boost libraries are installed as shared libraries, because of Fedora policy, so static linking is not working. I wrote small script to fix annoying problems with exercises:

    #!/bin/bash
    set -ue
    project="${1:?ERROR: Name of directory with exercism CPP project code is required.}"
    
    project_name="${project//-/_}"
    
    cd "$project"
    
    # Create empty .cpp and .h files
    touch "$project_name."{cpp,h}
    
    # Turn off static (.a) linking for boost libraries, because only dynamic (.so) libraries are available
    sed -i 's/set(Boost_USE_STATIC_LIBS ON)/set(Boost_USE_STATIC_LIBS OFF)/' CMakeLists.txt
    
    # Add def BOOST_TEST_DYN_LINK to *_test.cpp, to generate main()
    fgrep -q '#define BOOST_TEST_DYN_LINK' "${project_name}_test.cpp" || {
      sed -i '/#define BOOST_TEST_MAIN/ a#define BOOST_TEST_DYN_LINK' "${project_name}_test.cpp"
    }
    
    # Enable all test cases and generate Makefile
    cmake -DEXERCISM_RUN_ALL_TESTS=yes .
    
    # Build project
    make

IMHO, this script can be recommended in section about setup of C++ exercises on (Fedora) Linux.

### LINKED ISSUE #123: How to set BOOST_LIBRARYDIR or BOOST_ROOT?
Hi.  I'm getting the following error messages in CMake GUI in Windows 10, when pressing configure.  I'd much prefer to use cmake in Cygwin, but that's not working either so at this point anything that works would be great since I just want to learn C++ but can't currently run the tests, unfortunately.

The error message states:
_"No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the
  directory containing Boost libraries or BOOST_ROOT to the location of
  Boost."_

How do I set BOOST_LIBRARYDIR or BOOST_ROOT?

**CMake GUI error:**
----------------------------------------------
`CMake Error at C:/Program Files/CMake/share/cmake-3.8/Modules/FindBoost.cmake:1842 (message):
  Unable to find the requested Boost libraries.

  Boost version: 1.63.0

  Boost include path: C:/Program Files/boost/boost_1_63_0

  Could not find the following static Boost libraries:

          boost_unit_test_framework
          boost_date_time
          boost_regex

  No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the
  directory containing Boost libraries or BOOST_ROOT to the location of
  Boost.
Call Stack (most recent call first):
  CMakeLists.txt:14 (find_package)

CMake Warning (dev) at CMakeLists.txt:17 (if):
  Policy CMP0054 is not set: Only interpret if() arguments as variables or
  keywords when unquoted.  Run "cmake --help-policy CMP0054" for policy
  details.  Use the cmake_policy command to set the policy and suppress this
  warning.

  Quoted variables like "MSVC" will no longer be dereferenced when the policy
  is set to NEW.  Since the policy is not set the OLD behavior will be used.
This warning is for project developers.  Use -Wno-dev to suppress it.

Configuring incomplete, errors occurred!
See also "C:/Program Files/CMake/bin/CMakeFiles/CMakeOutput.log".`
----------------------------------------------

**Cygwin error:**
----------------------------------------------
`$ cmake -G "Unix Makefiles" ..
CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles".  CMAKE_MAKE_PROGRAM is not set.  You probably need to select a different build tool.
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage
-- Configuring incomplete, errors occurred!
`
--------------------------------------------

And so, additionally, how do I set CMAKE_MAKE_PROGRAM and CMAKE_CXX_COMPILER?

Any help is much appreciated, thanks!

### LINKED ISSUE #126: Add a variable to cmake to determine static or shared linked boost
On my system I do have a shared boost and linking it statically fails for some unknown reason. Currently I change those lines in CMakeLists.txt and the *_test.cpp by hand.

But shouldn't it be possible to provide a variable that can be set by the student and then create appropriate *_test.cpp?

As far as I can remember, cmake is able to pre-process any kind of file.

### LINKED ISSUE #149: gigasecond/gigasecond_test.cpp include header?

patricksjackson:
After one too many issues with people having trouble installing Boost, I created this to get feedback about removing it as a required dependency, primarily by switching to a header-only unit testing suite.

This closes every open bug in the repo (adding the tags here for auto-closing)
(closes #96, closes #116, closes #123, closes #126, closes #149, closes #150)

The primary goal was to remove Boost as a default dependency, though some tests might still require it. This change removes it from everywhere except for the `meetup` test, as that's already a nightmare of a test.

This moves to using [Catch](https://github.com/philsquared/Catch) for testing, because as far as I can tell that is the most popular header-only unit testing suite for C++ (I have no stake in this, I've never used it before). This allows us to bundle the test suite along with the problems, so **there are no dependencies other than a C++11 compliant compiler and some toolchain** (CMake + Make, or almost any IDE can work with the files directly).

For the majority of problems, this just involved replacing the Boost.Test macros with their Catch equivalents, but a handful of problems have more substantial changes:

- `gigasecond` is the most heavily impacted test, and is **a breaking change** for existing solutions. It used `boost::posix_time` across the user interface so I rewrote the interface to use strings, the example to use `<ctime>`, and increased the difficulty. It should at some point be re-ordered later in the track.
- `robot-names` used Boost.Regex to validate the user supplied names in the test suite. I attempted to replace this with std::regex, but due to issues on Travis-CI (that I wasn't able to reproduce with the same toolchain outside Travis) I just rewrote the validation to avoid regex. Ideally this should just use `std::regex`, but I didn't want to spend more time debugging it for a test that's just a one-liner if statement. See 1ff8dba
- `anagram`, `bob`, `crypto-square`, and `word-count` all used Boost libs in the example solutions (that was not user visible in the test suites). I removed those in change 8922c1f, but I don't consider those changes to be too important, since they are not user visible.

Documentation has not yet been updated, but will be if there's support for this change.


DonBarredora:
Is this ready to be merged into the master branch?

LegalizeAdulthood:
I think it would need to coincide with a big docs update.  Is that correct, Patrick?

pacmancoder:
Bump.

I think this issue should solved, as exercism should teach how to use C++ according to its modern standard. I think that all code, which use Boost should be replaced with corresponding stl-based.

patricksjackson:
Sorry for the delay here, I assumed somebody would've finished this change without me in the past yearish this has been open. I had to step away for a little while for various reasons, but I have a little time now to help out and clean up around here.

I will be taking another look at this and updating it for merge over the next few weeks. 

@pacmancoder or others, if you would like to help please open a PR against this branch. (This is tracked as another branch in this repo, so you can checkout this branch and make a PR targeting it. If not you can comment here with a pointer to your change and I can pull the commit).

This will be a big change, as all the Docs need to be updated as well. I will probably need help for testing it at the very least. I'll update here and/or make a tracking issue to track everything that needs to happen and how you could help.

patricksjackson:
Deprecating this, using #233 to track the work needed to incrementally move to Catch.

PR #160, opened 2017-11-03
Implement exercise binary-search

chgraef:


LegalizeAdulthood:
Why does this also contain commits related to the acronym exercise?

chgraef:
Hi @LegalizeAdulthood,

apologies for committing the files related to the the acronym exercise. This must have been the result of a problem re-basing my local repository. Can these files be removed from the commit?

PR #163, opened 2017-11-11
Implement exercise isogram

chgraef:
Started implementing the isogram exercise.


exercises/isogram/example.h
```diff
@@ -0,0 +1,7 @@
+#include <string>
```
patricksjackson: This needs include guards like the other exercises

exercises/isogram/example.h
```diff
@@ -0,0 +1,7 @@
...
+
+namespace isogram {
+
+bool is_isogram(std::string const);
```
patricksjackson: Should use a constant string reference here.

(In C++17 we'd use a string_view, but we need to support earlier versions)

exercises/isogram/example.cpp
```diff
@@ -0,0 +1,30 @@
...
+bool is_isogram(string const input_str)
+{
+    std::set<char> chars;
+    std::pair<std::set<char>::iterator, bool> ret;
+    
+    for (char const ch : input_str)
+    {
+        if (isalnum(ch))
+        {
+            ret = chars.insert(tolower(ch));
```
patricksjackson: It's probably clearer here to use `auto` and remove the type definition for ret above.
```suggestion
            auto ret = chars.insert(tolower(ch));
```

exercises/isogram/example.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <utility>
+#include "isogram.h"
+
+using std::string;
+
+namespace isogram {
+ 
+bool is_isogram(string const input_str)
+{
+    std::set<char> chars;
```
patricksjackson: Nit: You don't care about the order, so an unordered_set would be better.
```suggestion
    std::unordered_set<char> chars;
```

patricksjackson:
Obviously we (well, I) am very behind here, since this has been open almost a year and a half :\ Sorry.

I am trying to slowly clean things up around here. If you still have an interest in this, please make the suggested changes. If you're not interested any more, either respond here or do nothing by say the middle of March (2-3 weeks from now) and I'll open this up for others to finish up.

Since it's been so long there are things broken that have to do with the track itself. I am **not** asking you to fix them, since your PR was current when you made this. (You _can_ do it, but I will take on doing that myself if you like). That also applies to the test suite itself, which has had some updates - https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json

exercises/isogram/example.h
chgraef: Done. Good point!

exercises/isogram/example.h
chgraef: OK, done.

exercises/isogram/example.cpp
chgraef: OK, this adds some clarity.

chgraef:
Hello @patricksjackson!

Thank you very much for your note. It's been a while, hope you're doing well! 

In fact, I would love to help with updating my pull requests to the latest problem spec and new test suite. I might have a few questions, though, as I'm not entirely clear on the full extent of required changes. 

Hope this works for you?

Best wishes,

C.

patricksjackson:
Great! Of course you are welcome to do more work :)

The updates needed for each of the exercises are:
1. Updating to the new config.json file, it was formatted at some point so the merge is just going to fail.
2. Using the new default CMakeLists.txt file.
3. Running configlet to update the documentation.
4. Updating the tests since the tests have probably changed.

I'll try to finish reviewing your other PRs this weekend.

chgraef:
Hi @patricksjackson,

all done, ready to be merged! :-)

chgraef:
Actually, would it make sense to swap out Boost for Catch for unit testing straightaway? Would be happy to look into this.

config.json
```diff
@@ -470,5 +484,9 @@
         "mathematics"
       ]
     }
-  ]
+  ],
+  "foregone": [
```
patricksjackson: I believe all the additions here are now at the top of the file, so you don't need any of this group of changes

config.json
```diff
@@ -18,6 +18,20 @@
       ]
     },
     {
+      "core": false,
+      "difficulty": 1,
```
patricksjackson: I think the difficulty should be 3 here.

(The difficulty ratings right now are purely me taking a swag guess a while back, and I don't think I've had any input ever on whether they're wildly off. So we're just guessing here).

patricksjackson:
Everything looks fine, just a couple nits in the config.json and I'll merge it.

Don't bother switching out the test suite. This test suite in particular makes it easy, and I'd rather do them all at once (the BOOST_TEST macro is actually not the best thing to use in boost, because it won't print out fancy error messages, but it's a straight switch to Catch's REQUIRE which uses magic to figure out the left and right sides of the `==` and pretty print it). The benefit is making it unnecessary to install Boost, and we don't get that unless all of them are switched out. If half the exercises don't need boost I think it'll just be confusing.

config.json
chgraef: You're right, isogram is certainly not in the same ballpark as the hello world exercise. I made it a 3 now.

config.json
chgraef: Yep, this is now fixed. Thanks for the hint.

chgraef:
Thanks for all your pointers and for the comment re tweaking the tests to be compatible with Catch. I'm keen on helping, so, will keep my eyes open for other projects to work on in this track. Until then, I'm going to give a few more of the as-of-yet unimplemented exercises a go.

PR #164, opened 2017-11-11
Implement exercise binary-search

chgraef:
Started to implement binary-search exercise.


exercises/binary-search/example.h
```diff
@@ -0,0 +1,12 @@
+#if !defined(BINARY_SEARCH_H)
+#define BINARY_SEARCH_H
+
+#include <vector>
+
+namespace binary_search {
+
+int find(std::vector<int> const&, int);
```
patricksjackson: This should return a `size_t`, since it is returning an index.

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,28 @@
...
+            rpos = mid - 1;
+        } else if (value > data[mid]) {
+            lpos = mid + 1;
+        }
+        
+        if (data[mid] == value) { 
+            return mid; 
+        }
+    }
+        return -1;
```
patricksjackson:
```suggestion
        return SIZE_MAX;
```

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,28 @@
+#include "binary_search.h"
+
+namespace binary_search {
+
+int find(std::vector<int> const& data, int value)
+{
+    int lpos = 0;
```
patricksjackson: These types are indexes so they should all be `size_t`

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,28 @@
...
+    int rpos = data.size() - 1;
+    int mid = 0;
+
+    while (lpos <= rpos)
+    {
+        mid = (lpos + rpos)/2;
```
patricksjackson: There is a bug on this line, by the way. Though once you change the variables to `size_t` it will effectively never be hit.

patricksjackson:
Copying here from what I said in #163

Obviously we (well, I) am very behind here, since this has been open almost a year and a half :\ Sorry.

I am trying to slowly clean things up around here. If you still have an interest in this, please make the suggested changes. If you're not interested any more, either respond here or do nothing by say the middle of March (2-3 weeks from now) and I'll open this up for others to finish up.

Since it's been so long there are things broken that have to do with the track itself. I am **not** asking you to fix them, since your PR was current when you made this. (You _can_ do it, but I will take on doing that myself if you like). That also applies to the test suite itself, which has had some updates - https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search/canonical-data.json

exercises/binary-search/example.h
chgraef: OK, changed.

exercises/binary-search/example.cpp
patricksjackson: Sorry I should've been clear, I said this in a 'this is an interesting thing I've come across' way and not in a 'go fix this weird bug' way.

I like binary search, and I think it's a neat fact that it was first proposed in the 1940's but the first correct implementation didn't appear until the 1960's. And [most implementations are wrong due to the exact bug here](https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html).

Also, I work in the optimization space, and I've used some of the ideas in [this paper](https://arxiv.org/abs/1509.05053) in interviews to explain some of the kinds of analysis work that I do. Particularly interesting are the interactions between the branchy and branch-free binary search algorithms with the cache.

exercises/binary-search/example.cpp
chgraef: The updated problem spec suggests to raise a "value not in array" error in the last 5 cases. I changed my example and tests such that an exception is thrown and tested for, respectively. What are your thoughts on this?

exercises/binary-search/example.cpp
chgraef: Thanks. Changed this to size_t throughout.

exercises/binary-search/example.cpp
chgraef: Thanks for the explanations, the reading suggestions and for pointing out the bug in the first place! The first short article was a really enlightening read. Will defo put some time aside for a cup of tea and the arXiv paper tomorrow.

More of this, please! :-)

exercises/binary-search/example.cpp
patricksjackson: I think throwing is probably fine. I don't like it but it does fit the problem spec best :\

config.json
```diff
@@ -201,6 +201,21 @@
...
+        "arrays",
+        "control_flow_if_else_statements",
+        "control_flow_loops",
+        "searching"
+      ],
+      "unlocked_by": null,
+      "uuid": "abdd7aa3-25f6-4d27-a4dc-6bd1ea2f718c"
+    },
+    {
+      "core": false,
```
patricksjackson: There's a merge error here where you're adding to the next problem as well. You also need to re-format your settings.

The isogram merge had the same issue, and I missed it too. (I fixed that already)

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,40 @@
...
+        throw std::domain_error("Value not in input vector!");
+    }
+    
+    size_t lpos {0};
+    size_t rpos {data.size() - 1};
+    size_t mid  {0};
+
+    while (lpos <= rpos)
+    {
+        mid = (lpos + rpos) >> 1;
```
patricksjackson: Technically this doesn't fix the issue. You need
```
mid = lpos + ((rpos - lpos) / 2);
```
The fix in the blog post that they cite for Cpp was using signed indices (lpos and rpos), so they could just cast them to unsigned and the addition was guaranteed to never overflow due to the extra bit. But that's assuming the indices even fit into a signed int.

(Since you did want more reading, we do use size_t here for indexes, but there is a good argument for using a signed type -> isocpp/CppCoreGuidelines#1115. )

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,40 @@
+#include <stdexcept>
+#include "binary_search.h"
+
+namespace binary_search {
+
+size_t find(std::vector<int> const& data, int const value)
+{
+    if(data.empty() || value < data.front() || data.back() < value)
```
KevinWMatthews: A idea on style - some prefer to extract a complex boolean into a helper function such as:
```c++
if(!is_value_in_range(data, value))
```
feeling that it adds expressiveness to the code. Others don't like the extra level of indirection. I don't know if exercism has a preference.

It works nicely as it is; just food for thought.

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,40 @@
...
+        {
+            if (value < data.at(mid)) 
+            {
+                rpos = mid - 1;
+            } 
+            else if (value > data[mid]) 
+            {
+                lpos = mid + 1;
+            }
+        }
```
KevinWMatthews: Another thought on style - I've seen a lot of conditionals get out of control, so I can't help wanting to remove the `else`. Something like:
```c++
if (data.at(mid) == value) 
{ 
    return mid; 
}

if (value < data.at(mid)) 
{
    // snip
```
Again a matter of preference; it's perfectly fine as it is, and I'm sure some would find the current way preferable. 

Definitely not worth holding up a merge for something so small. Thanks for your implementation!

chgraef: Thanks, @KevinWMatthews for your comment. This is a good point, so, I unwrapped the nested if/elses and turned these into a series of simple ifs.

exercises/binary-search/example.cpp
chgraef: Good point, thanks @KevinWMatthews. I introduced a function, as per your suggestion.

config.json
chgraef: Done. Thanks for the pointer.

exercises/binary-search/example.cpp
chgraef: Oh, I see. I changed the LOC accordingly.

NB: Thanks for the link, very interesting read! I understand we'd run into a narrowing conversion issue if we were to change size_t for ptrdiff_t, since vector.size() returns an unsigned type. Hope that the index type will trickle down into the standard not too far in the future...

chgraef:
Hi @patricksjackson, thanks for all your comments, which are now addressed. Please feel free to merge if there aren't any further open points that would require attention.

KevinWMatthews:
Style points!

Do you like `is_value_in_range()` as part of the public interface or local to the .cpp file?

I like the `mid = lpos + ((rpos - lpos) / 2);` fix. It's a rare but real bug.

I'm sure @patricksjackson will weigh in/merge, but I'll keep an eye on this issue too.

chgraef:
Hello @KevinWMatthews, good point, there's no need for the helper function to be declared in the header, so I changed this and updated the pull request.

Cheers for the hint, it's much appreciated!

exercises/binary-search/binary_search_test.cpp
```diff
@@ -0,0 +1,114 @@
+#include "binary_search.h"
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+#include <stdexcept>
+#include <vector>
+#include <cstddef>
+
+using namespace std;
```
patricksjackson: You can just remove this as you don't use it anywhere
```suggestion
```

patricksjackson:
Woo! I'll go ahead and merge.

PR #165, opened 2017-11-11
Implement exercise complex-numbers

chgraef:
Starting to implement exercise complex-numbers.


exercises/complex-numbers/complex_numbers_test.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+BOOST_AUTO_TEST_CASE(Absolute_value_of_a_positive_purely_real_number)
+{
+    const complex_numbers::Complex c {5.0, 0.0};
+    
+    const double actual = c.abs();
+
+    const double expected {5.0};
+
+    BOOST_CHECK(expected == actual);
```
patricksjackson: You need to be careful comparing doubles here.

https://www.boost.org/doc/libs/1_59_0/libs/test/doc/html/boost_test/testing_tools/extended_comparison/floating_point.html

exercises/complex-numbers/complex_numbers_test.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include <boost/test/unit_test.hpp>
+#include <complex>
+#include <cmath>
+#include "complex_numbers.h"
+
+// Complex Numbers exercise test case data version 1.0.0
+
+BOOST_AUTO_TEST_CASE(Add_purely_real_numbers)
+{
+    const complex_numbers::Complex c1 {1.0, 0.0};
```
patricksjackson: I would just add a using definition for the Complex type since you use it soo much.

patricksjackson:
What do you think about making this type generic? This is a pretty simple type to do that on, so it could be used along with binary search tree #220 so we'd have some good exercises on that.

You also could use enable_if to limit the types to the floating point types just like std::complex, though I'm not sure if there's a way to write a unit test to make people do that...

I did not review all of this yet in case this gets switched to generic and a ton changes.

exercises/complex-numbers/complex_numbers_test.cpp
Ben1980: Yes that's not an easy topic: [Comparing Floating Point](https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/)

patricksjackson:
@KevinWMatthews can you take a quick look at this? I made a large set of changes to update this.

patricksjackson:
The Github actions already caught a Windows bug here: https://github.com/exercism/cpp/pull/165/checks?check_run_id=222283456

KevinWMatthews:
@patricksjackson Just saw this notification - will look at this soon.

patricksjackson:
@KevinWMatthews No rush, just whenever you get the chance :)

exercises/complex-numbers/complex_numbers_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+// Define M_PI for platforms that don't include it
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
+// Complex Numbers exercise test case data version 1.0.0
+
+using complex_numbers::Complex;
+
+TEST_CASE("add_purely_real_numbers") {
```
KevinWMatthews: [problem-specifications #1140](https://github.com/exercism/problem-specifications/pull/1140) has since reordered the test suite with a better workflow for students. Instead of jumping straight to addition they first test real and imaginary parts, then move to more complex operations.

exercises/complex-numbers/complex_numbers_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+TEST_CASE("multiply_numbers_with_real_and_imaginary_part") {
+    const Complex c1{1.0, 2.0};
+    const Complex c2{3.0, 4.0};
+
+    REQUIRE(Complex(-5.0, 10.0) == c1 * c2);
+}
+
+TEST_CASE("imaginary_unit") {
```
KevinWMatthews: The reordered problem specifications repo currently puts the test for [imaginary unit](https://github.com/chgraef/cpp/blob/7cad6c17cb12d39ef888745b32fb7755de284291/exercises/complex-numbers/complex_numbers_test.cpp#L81) in a weird place. I think it should stay with the multiplication examples as we have it here. I've opened [an issue](https://github.com/exercism/problem-specifications/issues/1584) in the problem specs repo.

exercises/complex-numbers/complex_numbers_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+    const Complex c{1.0, 2.0};
+
+    REQUIRE(Approx(2.0) == c.imag());
+}
+
+TEST_CASE("eulers_identity") {
+    const Complex c{0.0, M_PI};
+    const std::complex<double> cx{0.0, M_PI};
+
+    REQUIRE(Complex(exp(cx).real(), exp(cx).imag()) == c.exp());
```
KevinWMatthews: This follows the pattern of other tests but is a bit dense.
It is clearer to read if we create an intermediate object? Something like this:

```c++
const std::complex<double> cx{1.0, 0.0};
const std::complex<double> expected {exp(cx)};
REQUIRE(Complex {expected.real(), expected.imag()} == c.exp());
```

exercises/complex-numbers/complex_numbers_test.cpp
```diff
@@ -0,0 +1,218 @@
...
+#include <complex>
+
+#include "test/catch.hpp"
+
+// Define M_PI for platforms that don't include it
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
+
+// Complex Numbers exercise test case data version 1.0.0
```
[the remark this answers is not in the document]
patricksjackson: I've been trying to add the version info to the test suites when they get touched, since it's easier to tell if it needs an update.

(If this comment was talking about defining `M_PI`, it isn't defined on Windows for some reason. It's supposed to be [1], but that didn't work for me so defining it ourselves is easier than fixing compat problems.)

[1] https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=vs-2019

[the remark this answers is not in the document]
KevinWMatthews: Ah, my comment was unclear; I meant to refer to the version number.

I didn't realize that you added in the version info - I thought it might've been carried over from the original commit in 2017. Does the version number need to sync with something specific (say, the problem-specifications repo) or do we just increment it as we touch the file?

About `M_PI`, defining it manually is the safer route. Who knows what other compilers will do....

exercises/complex-numbers/complex_numbers_test.cpp
patricksjackson: Updated everything to the latest test suite, which reordered all the tests and added an extra at the end.

exercises/complex-numbers/README.md
```diff
@@ -0,0 +1,55 @@
...
+Multiplication result is by definition
+`(a + i * b) * (c + i * d) = (a * c - b * d) + (b * c + a * d) * i`.
+
+The reciprocal of a non-zero complex number is
+`1 / (a + i * b) = a/(a^2 + b^2) - b/(a^2 + b^2) * i`.
+
+Dividing a complex number `a + i * b` by another `c + i * d` gives:
+`(a + i * b) / (c + i * d) = (a * c + b * d)/(c^2 + d^2) + (b * c - a * d)/(c^2 + d^2) * i`.
+
+Exponent of a complex number can be expressed as
```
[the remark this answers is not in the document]
patricksjackson: Done.

exercises/complex-numbers/complex_numbers_test.cpp
patricksjackson: That's fair. I've updated to use the latest version in the problem specifications repo, though we could wait to see if the decision is made to change the tests before merging this in.

exercises/complex-numbers/complex_numbers_test.cpp
patricksjackson: This is significantly different after changes. I implemented a constructor and `operator==` for interop with `std::complex<double>`. So the test now just reads:

``` cpp
const Complex c{0.0, M_PI};
const std::complex<double> cx{0.0, M_PI};

REQUIRE(exp(cx) == c.exp());
```

This is potentially too far though, since for this test they'll have to implement not only `exp()` but also at a minimum `bool operator==(const std::complex<double>&, const Complex&)`

exercises/complex-numbers/complex_numbers_test.cpp
KevinWMatthews: I'm a little more inclined to merge this as soon as we're done. Apparently the exercise ordering hasn't been a big problem for students/mentors? And I'll get a notification if the problem spec changes.

exercises/complex-numbers/complex_numbers_test.cpp
KevinWMatthews: Hmmm, that is nicer to read.

I wonder if we could replace entirely replace `std::complex` with `Complex`? Several other languages seem to do that, such as [Java](https://github.com/exercism/java/blob/d59537b60e1a2b7243d0043d67255ecfcbbe2ffd/exercises/complex-numbers/src/test/java/ComplexNumberTest.java#L250), [Ruby](https://github.com/exercism/ruby/blob/3e9a042c07d4297e4bc6b243e6782221aa9298b0/exercises/complex-numbers/complex_numbers_test.rb#L174), [Python](https://github.com/exercism/python/blob/c05518984eba3f9ecdc88641545ba439f5117277/exercises/complex-numbers/complex_numbers_test.py#L169), and [JavaScript](https://github.com/exercism/javascript/blob/d4cc5f5aed4dbaf5bfd80d79ad7c29f226a39ccb/exercises/complex-numbers/complex-numbers.spec.js#L201).

If we do that, then we could remove two of the `==` operators and the `std::complex` constructor.

Another idea is to dispense with `==` entirely and compare the real and imaginary parts separately, like the Python and JavaScript do. At this point in the track, do you think using `std::numeric_limits` is too complex for students or an appropriate challenge?

patricksjackson: Oh right, we don't actually need `std::complex` at all. I like keeping the `==` operator at least, but yea I'll simplify so we can avoid `std::complex`. We'll still have all the operator implementations to do for this problem.

Using epsilon and scaling to the size of the operands isn't really necessary for most solutions. I expect most students will just use `std::abs(x-y) < .005` or similar if they need to do it, and the mentors can go as far down that rabbit hole as they want.

exercises/complex-numbers/complex_numbers_test.cpp
patricksjackson: :+1:

exercises/complex-numbers/complex_numbers_test.cpp
KevinWMatthews: Much improved! Looks good.

I like the idea that the students can submit a simpler solution using `std::abs` while the example shows a more detailed implementation.

KevinWMatthews:
@patricksjackson What else remains to do? The only thing I can think of is to make the solution type generic, but that can be done later.

patricksjackson:
I think we can just not bother making this generic and keep it pretty simple, I think we already have a simple generic class somewhere anyways. So really this exercise will just be a simple class with many of the mathematical operators used, and usage of the cmath header. What do you think?

KevinWMatthews:
Sounds reasonable to me.

patricksjackson:
This is done, I'm just waiting to merge this with #319 so we only run the update script once.

PR #166, opened 2017-11-17
Implement exercise robot-simulator

chgraef:
Starting to implement robot-simulator exercise.


exercises/robot-simulator/example.cpp
```diff
@@ -0,0 +1,87 @@
+#include "robot_simulator.h"
+
+robot_simulator::Robot::Robot() : robot_position(std::make_pair(0, 0)), robot_bearing(robot_simulator::Bearing::NORTH) {}
```
patricksjackson: I would define this file inside the namespace to avoid these long names.

exercises/robot-simulator/example.h
```diff
@@ -0,0 +1,27 @@
...
+
+class Robot 
+{
+public:
+    Robot();
+    Robot(std::pair<int, int>, Bearing);
+    void turn_left(); 
+    void turn_right();
+    void advance();
+    void execute_sequence(std::string);
```
patricksjackson: Use a const string reference here.

exercises/robot-simulator/robot_simulator_test.cpp
```diff
@@ -0,0 +1,366 @@
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+#include "robot_simulator.h"
+
+// Robot-Simulator exercise test case data version 2.0.0
+
+BOOST_AUTO_TEST_CASE(Robots_are_created_with_a_position_and_direction)
+{
+    const robot_simulator::Robot r;
```
patricksjackson: This file is long enough I'd add a using declaration for both `Robot` and `Bearing` just to make it a little shorter to read.

patricksjackson:
A couple of small things here. The only other needed changes are stuff you know how to do from the other problems, or that I volunteered to do myself :)

chgraef:
Hey @patricksjackson, 

updating the tests did not seem to have triggered a new Travis build. I was wondering if this is intentional?

patricksjackson:
Travis builds the merge between the PR and the upstream branch (master). I think it's not working because there's a conflict with master. You need to do another merge with master.

exercises/robot-simulator/example.cpp
chgraef: Good point, I've done as you're suggesting.

exercises/robot-simulator/example.h
chgraef: OK, changed this. No need to copy the string, really...

exercises/robot-simulator/robot_simulator_test.cpp
chgraef: You're right. This is saving a lot of text. So, I added these two using declarations, as you're suggesting.

chgraef:
Thanks for answering my question. The was indeed a conflict with the master branch.

NB: If you're ok with the current state, I think this exercise should be ready to be merged.

patricksjackson:
I would really prefer it if the test cases were easier to read, right now they are extremely verbose.

At a minimum, we could place the blank lines only above and below the actual test. So basically have the following format to clarify the difference between setup, test, and checking.
```
void test() {
  // Start block (no blank lines)

  // Test (moving forward, turning, whatever)

  // Check final state
}
```

It might also be clearer if we clean up some of the extra definitions. I think it's fine to just have the actual/expected values inline in the checks so longs as they don't need to be reused. We do that for many other tests, but in this case it's adding a lot of lines that end up being just boilerplate. If you want to be that verbose on the checks you could instead define a helper function that checks a robot against a bearing and position since we do that in every test. Running a BOOST_CHECK in a helper function works fine.

(Don't take the obvious optimization of creating a new robot from the expected bearing and position and comparing with the actual, as that would let the student cheat too easily).

```
BOOST_AUTO_TEST_CASE() {
  const std::pair<int, int> starting_position{0, 0};
  Robot r{starting_position, Bearing::NORTH};

  r.turn_right();

  BOOST_CHECK(starting_position == r.get_position();
  BOOST_CHECK(Bearing::EAST == r.get_bearing();
}
```

chgraef:
You're right, @patricksjackson, thanks for the hint. What do you think about the updated version?

patricksjackson:
Excellent! Thanks @chgraef! Merging this now.

Sorry for being a bit nitpicky on the tests, that's the only code the students see so I try to make those somewhat consistent in style/form.

chgraef:
You're welcome. Cheers for your comments!

PR #170, opened 2017-12-31
Fix invalid uuids

sjwarner-bp:
Configlet previously could generate invalid uuids.

It looks as though `atbash-cipher`, `bracket-push` and `pangram` had invalid uuids.


patricksjackson:
Thanks.

PR #178, opened 2018-03-20
Implement an exercise: Reverse String

nidhi98gupta:


LegalizeAdulthood:
Thanks for the contribution!

IMO, it would be more idiomatic to use the standard library reverse algorithm, making this a one-liner:
```
#include <algorithm>

std::string reverse_string(std::string text)
{
    std::reverse(text.begin(), text.end());
    return text;
}
```

elyashiv:
Or just:
```cpp
std::string reverse_string(const std::string &str)
{
  return std::string(str.rbegin(), str.rend());
}
```

patricksjackson:
This was close enough to finished that I made the changes myself.

Thanks!

PR #179, opened 2018-04-10
Implement exercise: Protein-translation

Smanikonda:


LegalizeAdulthood:
Please address the travis errors.

Also, please use white space formatting that is consistent with the existing problems.  In particular, this means:
* `std::vector<std::string>` not `std::vector <std::string>` (no space before template arguments in `<>`)
* `if (condition)` not `if(condition)` (don't write control structures like function calls)

Thanks!

patricksjackson:
Closing this since none of the comments have been addressed for a while. Please reopen if you feel like picking this back up :)

Issue #183, opened 2018-06-12
Ensure C++ track is ready for v2 launch


kytrinyx:
There are a number of things we're going to want to check before the v2 site goes live. There are notes below that flesh out all the checklist items.

- [ ] The track has a page on the v2 site: https://v2.exercism.io/tracks/cpp
- [ ] The track page has a short description under the name (not starting with `TODO`)
- [ ] The "About" section is a friendly, colloquial, compelling introduction
- [ ] The "About" section follows the formatting guidelines
- [ ] The code example gives a good taste of the language and fits within the boundaries of the background image
- [ ] There are exercises marked as `core`
- [ ] Exercises have rough estimates of difficulty
- [ ] Exercises have topics associated with them
- [x] The first exercise is `auto_approve: true`

## Track landing page

The v2 site has a landing page for each track, which should make people want to join it. If the track page is missing, ping `@kytrinyx` to get it added.

### Blurb
If the header of the page starts with `TODO`, then submit a pull request to https://github.com/exercism/cpp/blob/master/config.json with a `blurb` key.  Remember to get [configlet](https://github.com/exercism/configlet/releases/latest) and run `configlet fmt .` from the root of the track before submitting.

### About section
If the "About" section feels a bit dry, then submit a pull request to https://github.com/exercism/cpp/blob/master/docs/ABOUT.md with suggested tweaks.

#### Formatting guidelines
In order to work well with the design of the new site, we're restricting the formatting of the `ABOUT.md`. It can use:

* Bold
* Italics
* Links
* Bullet lists
* Number lists

Additionally:

* Each sentence should be on its own line
* Paragraphs should be separated by an empty line
* Explicit `<br/>` can be used to split a paragraph into lines without spacing between them, however this is discouraged.

### Code example
If the code example is too short or too wide or too long or too uninteresting, submit a pull request to https://github.com/exercism/ocaml/blob/master/docs/SNIPPET.txt with a suggested replacement.

## Exercise metadata
Where the v1 site has a long, linear list of exercises, the v2 site has organized exercises into a small set of required exercises ("core").

If you update the track config, remember to get [configlet](https://github.com/exercism/configlet/releases/latest) and run `configlet fmt .` from the root of the track before submitting.

### Topic and difficulty
Core exercises unlock optional additional exercises, which can be filtered by topic an difficulty, however that will only work if we add topics and difficulties to the exercises in the track config, which is in https://github.com/exercism/cpp/blob/master/config.json

### Auto-approval
[...]

PR #185, opened 2018-06-13
Added auto_approve true for hello-world

jackhughesweb:
I've set hello-world to auto_approve for issue #183


LegalizeAdulthood:
Thanks

PR #186, opened 2018-06-14
Updated ABOUT.md to include current standard

jackhughesweb:
I've updated the about section to reference the current C++17 standard (it may be useful to reword this section completely in future to avoid needing to update this again).

[Issue #183]


LegalizeAdulthood:
Thanks!  The standard is generally on a 3 year update cycle, so this isn't going to churn very often.

PR #187, opened 2018-06-14
Added core and unlocked_by to exercises

jackhughesweb:
I've selected 10 core exercises based upon difficulties, and added unlocked_by to the rest of the exercises (apart from a couple which are neither core nor locked by default - "bonus exercises")

Issue #183


LegalizeAdulthood:
Thanks!

PR #192, opened 2018-07-21
Implementation of isogram, secret-handshake, collatz-conjecture, binary-search

SahilTara:
Hey, I decided to implement some of the exercises needing to be implemented still.
Tell me if anything needs to be changed.
Updated to conform with exercism/exercism#4110 for collatz-conjecture.

### DIFF
--- a/exercises/binary-search/example.cpp
+++ b/exercises/binary-search/example.cpp
@@ -1,0 +1,23 @@
+#include "binary_search.h"
+
+int binary_search::find(const std::vector<int>& arr, int val) {
+	int low = 0;
+	int high = arr.size() - 1;
+	
+	while (low <= high) {
+		int mid = (static_cast<unsigned int>(low) + static_cast<unsigned int>(high)) >> 1;
+		int key = arr[mid];
+
+		if (key < val) {
+			low = mid + 1;
+		}
+		else if (key > val) {
+			high = mid - 1;
+		}
+		else {
+			return mid;
+		}
+	}
+
+	return -1;
+}
--- a/exercises/isogram/example.cpp
+++ b/exercises/isogram/example.cpp
@@ -1,6 +1,4 @@
 #include "isogram.h"
-
-#include <string>
 #include <unordered_set>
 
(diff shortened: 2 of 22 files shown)


SahilTara:
Hey, @patricksjackson or @LegalizeAdulthood can one of you please tell me what I need to add/change so that this PR can be merged?

exercises/binary-search/example.cpp
```diff
@@ -0,0 +1,23 @@
+#include "binary_search.h"
+
+int binary_search::find(const std::vector<int>& arr, int val) {
```
elyashiv: shouldn't we return an iterator?

exercises/isogram/example.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+	for (auto&& ch : word) {
+		ch = tolower(ch);
+
+		// character not parsed over
+		if (letters.find(ch) == letters.end()) {
+			letters.insert(ch);
+		}
+		// character is a member of the alphabet
+		else if (alphabet.find(ch) != alphabet.end()) {
```
elyashiv: The ifs should be the other way round. Also, shouldn't we use `isalpha`?

exercises/binary-search/example.cpp
elyashiv: I don't think we need to do like other languages. The idea is to learn the language, and the best practice is to return an iterator. I don't want to just use the std function because it teaches less.

patricksjackson:
Obviously we (well, I) am very behind here, since this has been open 6+ months without me addressing it. Sorry.

There are really two larger issues here before we can move forward with this PR.
1. This has 4 exercises in it. It's a lot easier if we address each of them individually, so please break this up.
2. 2 of the exercises (isogram and binary search) already had open existing PRs (one of those I've now merged). Yes, they were old (my fault) but they were still open so I assume they take precedence.

patricksjackson:
Closing this for now, feel free to reopen/split this up.

PR #198, opened 2018-08-04
Add 'math' topic to relevant exercises in track config

kytrinyx:
The new site lets people filter optional exercises by topic, and this
makes it easy for people to skip past them.

This is important since the math-y exercises are not key to learning
the language, and a lot of people find them intimidating and demotivating.

This updates any 'mathematics' topics to be 'math' since it's shorter
and won't get cropped in the UI.

It also removes the math topics for any exercises that are not in the
canonical math-y list.

exercism/exercism.io#4110

### DIFF
--- a/config.json
+++ b/config.json
@@ -106,7 +106,8 @@
       "unlocked_by": "leap",
       "difficulty": 3,
       "topics": [
-        "control_flow_loops"
+        "control_flow_loops",
+        "math"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


kytrinyx:
@patricksjackson @LegalizeAdulthood I normalized 'mathematics' to 'math', per the discussion in the linked issue, which I think is pretty uncontroversial. This also removes the 'math' topic from _triangle_ and _roman numerals_, since while they are related to math, they don't seem to be inherently about math (though... I guess "definition of triangle" could be seen as math).

kytrinyx:
I'm going to go ahead and merge this. We can always tweak in later PRs if there are any changes you think we should have.

Issue #142, opened 2017-08-11
Gigasecond unit tests don't include the proper boost headers


hpwooten-wta:
In the test suite for gigasecond, the compiler emits errors about not finding the types from boost.dateTime. This is because the proper header isn't included. I can see two ways forward for this:

1. Don't do anything. Having to debug this issue was a good exercise, because it taught me to remember to check the headers (which have to be used in the user-provided gigasecond.h anyway). 
2. Include the headers into the test suite. The user will still need to include the header in their own solution files, so they'll still get that practice.

patricksjackson:
I don't remember this being an issue when I did the problem, but I don't usually try to compile an empty solution. I understand many people do though, which is one reason why we've been adding some of the `std` includes needed by the test suite. ( #121 )

This one is a bit different though, since reading the Boost docs (with the link given in the test suite) and then writing the obvious 1 liner is the entire exercise. Giving them the right header in the test suite lets the user avoid the issue of figuring out what include file they need.

I think it's better to leave as is. Thanks for bringing it up though :) 

(If you do feel strongly that it should be in there, reopen and we can discuss more. These types of discussions about improving the test suite are really valuable to have. )

hpwooten-wta:
I was torn on whether including the header would be better or worse, so thanks for your input. I agree with this decision.

PR #201, opened 2018-08-18
Including needed boost::posix_time.h

menaczar:
I'm getting a compiler error when trying the gigasecond exercise.
The library needed for posit_time was missing.
Please check the boost documentations: http://www.boost.org/doc/libs/1_59_0/doc/html/date_time/posix_time.html

### DIFF
--- a/exercises/gigasecond/gigasecond_test.cpp
+++ b/exercises/gigasecond/gigasecond_test.cpp
@@ -1,6 +1,7 @@
 #include "gigasecond.h"
 #define BOOST_TEST_MAIN
 #include <boost/test/unit_test.hpp>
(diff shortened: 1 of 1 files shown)


patricksjackson:
This has come up before in #142. The missing header is by design, you need to include it in your header file anyways to pick up the appropriate types.

Thanks

patricksjackson:
Nvm, after looking at how many open issues we have around this I've changed my mind.

I've been against this in the past (since you have to put in real effort to write an implementation that actually needs this include in the test suite), but given how many issues we've seen with people hitting initial compile problems with`meetup` and `gigasecond` (See #201, #142, #149, #150) I'm going to stop being stubborn and just merge it.

Thanks

Issue #193, opened 2018-07-21
Add more compilation flags


AlexLeSang:
I think that compile flags for exercises should contain at least '-Wall', maybe '-Wextra' and '-Wpedantic' and even '-Werror'. That would provide students with better diagnostics from the compiler and avoid common mistakes.

patricksjackson:
I agree with this. It's what we've done on the C track as well. This is the list we use there:

```
CFLAGS += -g
CFLAGS += -Wall
CFLAGS += -Wextra
CFLAGS += -pedantic
CFLAGS += -Werror
```

@AlexLeSang would you like to make the change? It should be a change to [this line](https://github.com/exercism/cpp/blob/6e2a2693ca7a6d609396bd61699baf73f6087247/exercises/allergies/CMakeLists.txt#L18) that needs to be changed in every CMakeLists.txt file in the exercises folder.

KevinWMatthews:
Do you find `-Werror` to be a useful teaching tool? Personally I feel that it tends to disrupt one's workflow and it can be frustrating to be at the mercy of a pedantic compiler.

Having said that, it may be more important to force students to resolve all issues. Warnings are easy to ignore, especially if you don't clean the project.

patricksjackson:
I have no strong opinions on using `-Werror` here, it's useful for me (usually), and there are not many false positives with just all/extra/pedantic enabled. Generally those flags only pop up things that really should be fixed, or that actually are bugs int he code.

I admittedly don't teach enough to know if it's helpful for beginners, so I am totally open to other opinions.

KevinWMatthews:
Thinking over this again, I wonder if I'm projecting my experience with C warnings on to C++. I'm happy to defer to others.

@AlexLeSang are you excited to add this? If so, by all means. If not, I might give it a go myself - I should have time this weekend.

AlexLeSang:
Yes, I am very excited to do this and I will do it this weekend. 

In my experience working with less experienced colleagues, it is always better to promote a warning  to an error and pedantic compiler can teach a lot.

KevinWMatthews:
@AlexLeSang For sure. Enjoy!

patricksjackson:
These are now supported in the cmake files. Thanks!

PR #204, opened 2018-08-24
Add more compilation flags (#193)

AlexLeSang:


patricksjackson:
I'm going to hold off on this until #205 is figured out to avoid churning the CMake files too much. I'm sorry for the extra effort you put in on this, but thanks for doing it.

I do want to merge your build error fixes though. Could you open a separate PR for that?

It seems like you might be new to git (because you made these changes on your `master` branch instead of a branch). This guide may help understand the basic workflow if you need it: https://github.com/exercism/docs/blob/master/contributing/git-basics.md

To set up a separate branch for just that one change you should be able to do that by [Resetting your Master branch](https://github.com/exercism/docs/blob/master/contributing/git-basics.md#resetting-master), being careful to do the `git checkout -b backup` step so that you make a separate branch with these changes saved. Then making a new branch forked from master, cherry-picking that one change, and opening a new PR. After following the steps to reset master, that's just:

```
git checkout master
git checkout -b compile_fixes
git cherry-pick 6898ee59d508d214a5d142542efdcc46cff400ae
```

That long string is the hash of that one commit.

AlexLeSang:
@patricksjackson Indeed I should have delivered on a separate feature branch instead of using master.

Sure, I can apply flags changes after #205.

I will open a separate PR with build fixes.

PR #208, opened 2018-08-30
Run tests on hello-world exercise

KevinWMatthews:
A full build skips the `hello-world` exercise entirely.

Is there any reason that we should avoid this exercise?


patricksjackson:
Wow, yea that shouldn't be the case. We need to have the tests running for everything. Thanks!

PR #215, opened 2018-12-28
moved to use enum class

elyashiv:
I needed to put the `BOOST_TEST_DONT_PRINT_LOG_VALUE` as mentioned [here](https://stackoverflow.com/questions/30860241/use-enum-classes-with-boost-test).

The error messages are still readable and contain the actual expression.

This will break previous solutions.


patricksjackson:
Interesting, I didn't know about enum classes. That's a decent update to what we have now.

Breaking previous solutions isn't really a problem. Each student's version is locked against the git commit that they first started the exercise with. This is a good enough update to warrant the change though.

Thanks!

PR #219, opened 2019-02-20
Surround math operator with space

hanneskaeufler:
All other math operators are surrounded by spaces. This one was missing a leading space.


patricksjackson:
These small formatting things tend to happen since we don't run things through a formatting tool.

Thanks!

PR #220, opened 2019-02-23
Implement exercise: binary search tree

elyashiv:


patricksjackson:
This is looking pretty good, though I won't review it until the build is passing. If you need help debugging those errors, let me know. From a quick glance it looks like the problem is that Boost doesn't know how to print one of the things you're using to compare.

I also would ask that you make this class generic using templates. We don't have that anywhere in the repo yet, but this would be an excellent opportunity to do so.

elyashiv:
1. I think there might be a problem with the versions - I have a later boost version than travis uses, and a newer compiler. I will see if I can fix that.
2. I will try, though I don't have much experience with generics. Is there a good way to make tests for the generics bonus?

exercises/binary-search-tree/binary_search_tree_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+    test_leaf(tested->left()->left(), 1, false, false);
+    test_leaf(tested->left()->right(), 3, false, false);
+
+    test_leaf(tested->right()->left(), 5, false, false);
+    test_leaf(tested->right()->right(), 7, false, false);
+}
+
+static void test_sort(const tree_ptr &tree, const std::vector<int> &expected)
+{
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(tree->begin(), tree->end(), expected.begin(), expected.end());
```
patricksjackson: The compile error is probably this line due to the custom iterator.

patricksjackson:
Nice! It looks like you fixed the compile issues :)

You've already done most of the work to make this generic. All you need to do is add a `template <T>` to the line above the class definition, and define the `tree_data_t` to be `T`. You'll also need to move the entire class definition into the header file.

I need to think about this a bit more, but I'm currently thinking that what we can do here is make the generics part of the exercise (it sort of has to be once you template the class) and make the iterator definition the bonus. We'd need to make the normal tests pass without using the iterators, then add a test or two to the bottom that does use them.

I will look more tomorrow and try to make time to review it.

patricksjackson:
*I messed up in my comment above. The way to define a template is `template <class T>` not `template <T>`.

I'll hold off on reviewing this again too, since you do still have it as WIP. I'm curious what you think about making the iterator optional.

elyashiv:
Hi!

1. About the iterator - I think that it's the most cpp way to do the sort part of the exercise. The other option is to return a std::vector or something.
2. I need to find time for the template part of this, so I marked it as WIP. I guess there is nothing urgent...

elyashiv:
Hi. I added the generics to the tests/example. I'm still not sure how to make them bonus (If we want).

There seems to be a problem with the travis - it cannot find one of the boost libs. The tests do pass on my machine :)

patricksjackson:
I see you figured out the problem with Travis :) 

Just FYI - I don't have time to review this now, and I'll be on vacation this weekend. I'll get back to you (hopefully early) next week.

chgraef:
Hi @elyashiv, 

I ran into the same issue as you did with the failing Travis build. I noticed you removed both, date_time and regex as dependencies from your CMakeLists.txt. However, in my understanding only regex should be taken out, to be coherent with the existing exercises in the repository.

Cheers and sorry for the nitpick. ;-)

elyashiv:
Hן @chgraefת Thanks!

I need just to remove regex for this to compile, but as I don't need date_time, I might as well remove it, and reduce build time (and binary size, but that's less interesting in this case)

exercises/binary-search-tree/binary_search_tree_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+    test_leaf<uint32_t>(tested->left(), 2, true, true);
+    test_leaf<uint32_t>(tested->right(), 6, true, true);
+
+    test_leaf<uint32_t>(tested->left()->left(), 1, false, false);
+    test_leaf<uint32_t>(tested->left()->right(), 3, false, false);
+
+    test_leaf<uint32_t>(tested->right()->left(), 5, false, false);
+    test_leaf<uint32_t>(tested->right()->right(), 7, false, false);
+}
+
```
patricksjackson: You should add a comment here saying 'all the tests bellow require that you create your own custom iterator for your type'. And maybe add a link to cppreference or something.

Also you should move the string test that doesn't require an iterator above the comment.

exercises/binary-search-tree/binary_search_tree_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+template<typename T>
+static void test_sort(const tree_ptr<T> &tree, const std::vector<T> &expected)
+{
+    BOOST_REQUIRE_EQUAL_COLLECTIONS(tree->begin(), tree->end(), expected.begin(), expected.end());
```
patricksjackson: I am concerned that this will give ridiculously bad errors when here are problems with the students' iterators. Can you check what might have the best error messages with a broken iterator? (an iterator missing required functions basically).

I'd guess some of the following would give better errors.

```
std::vector<T> tmp;
std::copy(tree.begin(), tree.end(), std::back_inserter(tmp));
BOOST_CHECK(tmp == expected);
```
```
size_t i = 0;
for (auto& x : tree) {
  BOOST_CHECK(x == expected[i++]);
}
```

Or maybe write out the equivalent of the ranged for loop for the tree
```
size_t i = 0;
for (auto x = tree->begin(); x != tree->end(); x++) {
  BOOST_CHECK(*x == expected[i]);
}
```

I'm just concerned that this isn't a good way to provide tests (with good failure cases) for an iterator. The test suite is just way too simple for adequately testing it.

exercises/binary-search-tree/binary_search_tree_test.cpp
```diff
@@ -0,0 +1,146 @@
...
+BOOST_AUTO_TEST_CASE(can_sort_complex_tree)
+{
+    test_sort(make_tree<uint32_t>({2, 1, 3, 6, 7, 5}), {1, 2, 3, 5, 6, 7});
+}
+
+// strings
+
+BOOST_AUTO_TEST_CASE(can_create_complex_tree_strings)
+{
+    auto tested = make_tree<std::string>({"4", "2", "6", "1", "3", "5", "7"});
```
patricksjackson: I think you should replace these with short words instead of numbers. That'll be more clear to the student that this can operate on any string.

exercises/binary-search-tree/example.h
```diff
@@ -0,0 +1,226 @@
...
+{
+    template<typename T>
+    class binary_tree final
+    {
+    public:
+        class binary_tree_iter;
+        using binary_tree_ptr = std::unique_ptr<binary_tree>;
+
+        explicit binary_tree(T data)
+            : _data(std::move(data)),
```
patricksjackson: I don't think the data needs to be explicitly moved here. It's not a reference, so I'm almost certain the compiler figures that out anyways.

exercises/binary-search-tree/example.h
```diff
@@ -0,0 +1,226 @@
...
+        class binary_tree_iter final
+        {
+            friend binary_tree_iter binary_tree::begin() const;
+            friend binary_tree_iter binary_tree::end() const;
+        public:
+            using binary_tree_iter_ptr = std::unique_ptr<binary_tree_iter>;
+        private:
+            enum class state
+            {
+                Left,
```
patricksjackson: These enum values should be all caps as a matter of convention.

exercises/binary-search-tree/example.h
```diff
@@ -0,0 +1,226 @@
...
+                Right,
+                Middle,
+                Done
+            };
+
+            const binary_tree &_tree;
```
patricksjackson: Data values should be defined at the end of the class.

exercises/binary-search-tree/example.h
```diff
@@ -0,0 +1,226 @@
...
+            state _state;
+            binary_tree_iter_ptr _side_iter;
```
patricksjackson: The name `_side_iter` was very confusing for me. Maybe `_branch_iter` would be better?

exercises/binary-search-tree/example.h
```diff
@@ -0,0 +1,226 @@
...
+            location = binary_tree_ptr(new binary_tree(data));
+        else
+            location->insert(data);   
+    }
+
+    template<typename T>
+    void binary_tree<T>::insert(const T &data)
+    {
+        binary_tree_ptr &insert_location = data > _data ? _right : _left;
+        insert(data, insert_location);
```
patricksjackson: This is the only place you call the private insert function. It's not doing anything logically different than this function so you should just place it inline.

elyashiv:
Hi. Thanks for the comment! I fixed some, and I will fix the rest when I have some time.

Should I add iterator traits?

exercises/binary-search-tree/example.h
patricksjackson: On second thought, this is just an unnecessary fix, don't bother.

patricksjackson:
No problem, there was a lot of feedback. When you're ready for me to take another look just leave a comment.

I don't think you should add iterator traits, this is large enough as it is. That may mean one of my examples of code I gave you for the iterator wouldn't work, which is fine, and if that's the case please tell me. I've never actually built a production ready iterator, so I'm still learning a few things :)

exercises/binary-search-tree/binary_search_tree_test.cpp
elyashiv: Agree, I was just lazy...

exercises/binary-search-tree/binary_search_tree_test.cpp
elyashiv: I commented out some random functions, and I get a nice error message (cannot increment, requires pointer operator) with clang. We can start like this, and see.

elyashiv:
I cannot find the comment about the std::move, so I will just write here: I don't the compiler is allowed to just move it (it's a lvalue), so I need to do it explicitly. I shouldn't be using std::move but:
```c++
using std::move;
...
move(...)
```
So if someone has his own custom function I will use it.

elyashiv:
I think I'm ready for another pass of CR.

elyashiv:
Travis has a problem with the apt-repository... Not sure what to do :sob:

KevinWMatthews:
Hi @elyashiv, that's odd. `apt` is failing but `travis.yml` is unchanged.

At first glance I can't see why it would fail. We'll need to give this a closer look.

exercises/binary-search-tree/binary_search_tree_test.cpp
patricksjackson: Well, that was a good idea :) I admittedly was just guessing and didn't actually check.

Thanks for checking! This is fine then.

patricksjackson:
Travis has been failing to install the dependencies for some reason (it's not even the extra ppa we've added for boost, so I don't know why). It seems to always be a temporary problem, I re-ran it and it worked fine.

On the std::move, yes I realized I was wrong, thanks for looking up the proper fix :)

This is awesome, I'm going to go ahead and merge :) Thanks @elyashiv!

elyashiv:
Hi. I just noticed I have a problem with the move in the example - in case there is no move constructor, or that moves are not cheap, the approach I took is a problem, because I call both the copy-ctor and the move-ctor in every case. This is fine for strings and ints, but not for every class...

We should probably add mentoring notes for this problem.

patricksjackson:
Ah, I was unaware that mentoring notes were even a thing...so I haven't written any.

I see where they'd go now, in exercism/website-copy. I may or may not get to them (probably not soon though). Please @ add me if you write them as I don't watch that repo, I'll do the same for you if I happen to get to it.

PR #223, opened 2019-03-04
Removing Boost::Regex

patricksjackson:
This removes the dependency on Boost::Regex by rewriting the one place we use it (the Robot Name test suite). Switching to `std::regex` wasn't possible because there are inconsistencies in older compilers' implementations.


KevinWMatthews:
Interesting. I'd wondered why we included `Boost::regex`.

Issue #226, opened 2019-03-08
Lots of noise in git when running tests


elyashiv:
If I run tests locally on my machine, I get a lot of noise - all the example files are renamed.

Can we copy them instead of renaming, then add them to the .gitignore?

KevinWMatthews:
That's a good idea.

Thoughts:

Need to look into how CMake supports `file(COPY)` instead of `file(RENAME)`.

I wonder if we could symlink? Hmmm.... I don't know how to test that cross-platform.

Would you update the root-level `.gitignore` or add a .gitignore to each directory?
We'd need to be sure that the user doesn't download the `.gitignore` file - that could be frustrating.

I don't know of any way that `.gitignore` supports advanced pattern matching, so we might need to configure each exercise by hand. Can you think of another way?

Are there consequences to editing the `.gitignore` that I'm not thinking about? Maybe just copying is enough? It might be nice to know that copied files are present.

KevinWMatthews:
As far as I can tell, CMake's `file(COPY)` does not support renames during a copy. This is trickier than I expected.

patricksjackson:
I'll note here, the other thing to remember is that some exercises have a skeleton header file that is delivered to the student, so we're actually overwriting those files. (One of the quirks of exercism. It downloads everything to the student except files that start with "example", as far as I remember. I think it would download a .gitignore if it was in the exercise directory.)

I feel this pain too while developing. Two thoughts (I don't have time to look myself right now).

1. Does the new cmake support running in a build/ subdirectory from the root?
2. Can we just setup the root cmake to create a parallel tree of exercises/ that we could gitignore?

I would advise against trying to symlink and deal with the cross-platform issues that would come out of that.

elyashiv:
I like the idea of another tree.

The gitignore can be contain something like:
```
*
!CmakeLists.txt
!example.*
```
But that will ignore also the skeleton files...

KevinWMatthews:
@patricksjackson Yes, CMake does support running in a `build/` subdirectory. You can run it anywhere, for that matter. I'm currently running a true out-of-tree build (`build/` is at the same level as `cpp/`) and it works quite nicely.

I'm glad you mentioned the skeleton header files. I only see one? But it's an important one: `hello_world.h`.

I like the idea of a parallel exercise tree. It might seem strange at first, but we do have an atypical use case. I wrote up a paragraph of ideas, but I'll just go try them and report back.

[...]

PR #228, opened 2019-03-09
Copy all exercises to alternate tree, build these

KevinWMatthews:
Avoid renaming the original example files to keep a clean git diff.
The alternate tree is ignored - it's only used to build.


KevinWMatthews:
@patricksjackson  @elyashiv Proof of concept to ameliorate #226 . Does it work for you?

Pros:
  * Original `exercises/` directory is untouched
  * New `build_exercises/` directory is built and ignored by git
  * Works with in-tree and out-of-tree builds

Cons:
  * If the examples are changed, must re-run CMake to recopy the exercises into the alternate directory

Nitpicks:
  * I'm hard-coding `alt_exercise_tree` into `travis_fixup()`. Should probably add an extra argument.
  * What should we name stuff?

I'm looking at having the recopy run with every build - I think that facilitate a convenient workflow. There are a few sticking points with CMake semantics, so I'll come back to it if you like this.

KevinWMatthews:
Took a stab at fixing my own nitpicks.

KevinWMatthews:
@patricksjackson You're right; this is an unconventional workflow that deserves some explanation. I'll try it out - don't mean to shirk writing documentation - but you're welcome to edit heavily.

KevinWMatthews:
Ok, there's a first draft. I added in some background information; I figure it's easy to remove if it's too verbose.

KevinWMatthews:
I still hope that there's a way to make CMake a bit smarter. It may be possible to `add_custom_command()` to each target and have this custom command copy only the files for that particular exercise. That would remove a user step and provide an extra bit of granularity.

It might get complicated, however, so if you like the current CMake system as-is I'm inclined to start with this and make a separate pull request later.

README.md
```diff
@@ -22,3 +22,35 @@ Here is a list of things that need to be done to add an exercise to this track.
...
+The current CMake build system navigates this with an unusual approach: it copies
+all example solutions from the `exercises/` tree to an alternate directory
+(which is ignored by git), renames the solutions as if they were student exercises,
+and runs a complete build in this new directory.
+
+Maintainers can largely ignore the alternate exercise directory if they recopy
+their example solution before running a build. Re-running CMake will recopy
+all exercise files.
+
+For example, a maintainer can copy, configure, compile, and test all exercises with:
```
patricksjackson:
```suggestion
For example, a maintainer can copy, configure, compile, and test all exercises by
running the following from the root directory:
```

README.md
```diff
@@ -22,3 +22,35 @@ Here is a list of things that need to be done to add an exercise to this track.
...
+```bash
+cmake . && make
+```
+
+For an individual exercise, in this example `anagram`, the maintainer should edit:
+```bash
+exercises/anagram/example.h
+exercises/anagram/example.cpp
```
patricksjackson:
```suggestion
exercises/anagram/example.cpp
exercises/anagram/anagram_test.cpp
```

patricksjackson:
I have a few small nits , but otherwise is LGTM. (Just push yourself after fixing if those are fine).

This is already pretty simple, I'm not sure how it's an extra step unless you mean you could somehow get `make test_anagram` to do the copy and then build and run the tests?

KevinWMatthews:
Good nits. Trying out the GitHub interface for resolving/committing changes.

Yes, exactly that. It seems possible, at least in principle, to to set up a custom command or target that will copy the files for just the `anagram` executable. If the copy command/target could be configured as always out of date and as a prerequisite for the `anagram` target, then the dependency should trickle through to `test_anagram` and would be run every time.

The trick is that this can't be visible to students so it has to happen in the root `CMakeLists.txt` file. It might be an unreasonable amount of work.

Issue #227, opened 2019-03-09
c++ phone number mistake.


ViralTaco:
Hi,

Here: [https://github.com/exercism/cpp/tree/master/exercises/phone-number](https://github.com/exercism/cpp/tree/master/exercises/phone-number) 
you say

> The format is usually represented as
> 
> ```text
> (NXX)-NXX-XXXX
> ```
> 
> where `N` is any digit from 2 through 9 and `X` is any digit from 0 through 9.

However in the tests : 
`BOOST_REQUIRE_EQUAL("1234567890", phone.number());`

The problem: `1` isn't a digit **from** `2` through `9`

Thank you for your time

KevinWMatthews:
Well, that is a contradiction, isn't it?

Good catch - need to look into this.

KevinWMatthews:
Looks like the [problem description](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/description.md) and [problem specification](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json) have evolved but our [test suite](https://github.com/exercism/cpp/blob/master/exercises/phone-number/phone_number_test.cpp) has not.

ViralTaco:
Thanks for replying. 
I can look at implementing the tests from the problem spec but I'm not sure when I'll be able to do it. :/ 
I'll try to remember not to forget.

ViralTaco:
Hey @KevinWMatthews 
I got [this](https://github.com/ViralTaco/cpp/tree/exercism_phone_cpp/exercises/phone-number)
please tell me if it's ok to make a pull request (especially since it's on the master branch, don't wanna f that up)

KevinWMatthews:
Hey @ViralTaco ,

Thanks for your help. Yes, it's typically safe to make a pull request. Your changes will end up on the master branch of the main exercism repo, but they don't go directly there - the maintainers have to merge them first. So hack away!

Also, you can easily reference other issues or pull requests using a '#'. Something like this: #229 . Try it out - it's a nice GitHub feature.

ViralTaco:
Nice, thank you for the tip.

patricksjackson:
Closed by #230

PR #230, opened 2019-03-11
C++ phone number tests

ViralTaco:
Current examples shouldn't pass the tests since the [problem spec](https://github.com/exercism/problem-specifications/blob/master/exercises/phone-number/canonical-data.json) changed. 

EDIT: forgot to mention this is for issue #227

### DIFF
--- a/exercises/phone-number/example.cpp
+++ b/exercises/phone-number/example.cpp
@@ -1,6 +1,7 @@
 #include "phone_number.h"
 #include <algorithm>
 #include <cctype>
+#include <stdexcept>
 #include <iterator>
 #include <sstream>
 
@@ -13,7 +14,6 @@ const int area_code_length = 3;
 const int exchange_length = 3;
 const int extension_length = 4;
 const int valid_number_length = area_code_length + exchange_length + extension_length;
-const string cleaned_invalid_number(valid_number_length, '0');
 
 string clean_number(const string& text)
 {
@@ -24,10 +24,13 @@ string clean_number(const string& text)
         if (result[0] == '1') {
             result.erase(result.begin());
         } else {
-            result = cleaned_invalid_number;
+            throw std::domain_error("Invalid number");
         }
-    } else if (result.length() < valid_number_length) {
-        result = cleaned_invalid_number;
+    } else if (result.length() != valid_number_length) {
+        throw std::domain_error("Invalid number");
+    }
+    if (result[0] < '2' || result[3] < '2') {
+      throw std::domain_error("Invalid number");
     }
     return result;
 }
--- a/exercises/phone-number/phone_number_test.cpp
+++ b/exercises/phone-number/phone_number_test.cpp
@@ -1,54 +1,95 @@
 #include "phone_number.h"
+
 #define BOOST_TEST_MAIN
 #include <boost/test/unit_test.hpp>
+#include <exception>
 
-BOOST_AUTO_TEST_CASE(cleans_parens_dashes_and_spaces_from_the_number)
-{
-    const phone_number phone("(123) 456-7890");
-
-    BOOST_REQUIRE_EQUAL("1234567890", phone.number());
+BOOST_AUTO_TEST_CASE(cleans_the_number) {
+  BOOST_REQUIRE_EQUAL("2234567890", phone_number("(223) 456-7890").number());
 }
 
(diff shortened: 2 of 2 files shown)


patricksjackson:
Hey @ViralTaco,

Thank you for spending time on this. I know it can be frustrating dealing with contributing for the first time, dealing with git, etc. And I think you and @KevinWMatthews were talking past each other a little bit, which is more confusing.

I'm going to be very verbose here just because I'm not sure where you're coming from, and I'd like us all to be on the same page :)

If you want to get a better understanding of Git, I would suggest you start [here](https://github.com/exercism/docs/blob/master/contributing/git-basics.md), which is pretty good at explaining the basics of how the normal github flow works. You might be familiar with that already. For learning the _rest_ of git, I found it easiest to just google "how to X in git", which will almost always take you to a nice StackOverflow post, it just can take a while to get the right _name_ for the thing you want.

Also don't be afraid to just start over again if it's too annoying to fix up your git branch. I encourage copying files around and making new commits in some cases, especially here. Closing PR's that would take too much time to fix up properly is absolutely fine.

@KevinWMatthews actually wanted to know what was wrong with the current example solution. Making a PR like this, starting with updating the Examples fits well with [TDD](http://butunclebob.com/ArticleS.UncleBob.TheThreeRulesOfTdd). That lets Travis (the continuous integration server) run and show us the output failures. Sometimes that is easier to review since we aren't always at a dev machine.

Now since you've pushed it, we have a record of the failing tests. I can now see more easily what went wrong without getting my own environment setup. So now you should add in your example solution. Once you push that it will trigger a new build on Travis that will hopefully pass.

The best practice is to make the smallest change necessary to make the tests pass, while keeping to good style. For example: [this commit](https://github.com/patricksjackson/xcpp/commit/7971a1df827830be609b6b1acb9f949f4bd4a95c).

(If you wanted, you actually should be able to cherry-pick that commit from my repo using the following git commands (while you have this branch checked out))

```
git remote add other https://github.com/patricksjackson/xcpp.git
git fetch other
git cherry-pick 7971a1df827830be609b6b1acb9f949f4bd4a95c
```
If you want to take that change, that'll work. And this would be all set to merge.

However if you want to have your solution in there feel free to add that to this branch and I will give you a full review of that solution.

(The only nit I have on your test suite is that you have some extra tests somehow? That's not a big deal though. "has_an_area_code" and "formats_a_number" I don't think I see in the main test suite)

ViralTaco:
Hi @patricksjackson,

>(If you wanted, you actually should be able to cherry-pick that commit from my repo using the following git commands (while you have this branch checked out))

I will, thanks.

I think your commit is fine. I just have a personal problem with `using namespace std;` and then `namespace` to put everything in the global namespace. `string` especially is a type that comes in many different frameworks but I guess if it passes the tests it's good enough. 

Also `const int area_code_length = 3;` I would have had something more along the lines of:
`static constexpr std::size_t kAreaCodeLength{ 3 };` but I guess that's a nitpick. I don't even know if C++ 11 has constexpr I didn't look at C++ before the C++14  standard…

>(The only nit I have on your test suite is that you have some extra tests somehow? That's not a big deal though. "has_an_area_code" and "formats_a_number" I don't think I see in the main test suite)

They were in the old test suite. I just keep them there to be honest. I wasn't sure if it was ok to remove them.

patricksjackson:
I agree that the existing example is not perfect, but I disagree with some of the issues mentioned.

`using namespace std` is [usually considered fine](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnonexported-entities) to use in this context since it only affects the source file. Putting it in the header would have been a problem.

Using just `namespace` makes an anonymous namespace, which is generally [a good idea for internal objects](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf22-use-an-unnamed-anonymous-namespace-for-all-internalnonexported-entities), and will work pretty much identically to `static`.

Using `constexpr size_t`, and your different name choice for those globals would definitely be better.

I'd be happy to take improvements to the example, but here we care most about making sure that what the students see while doing the tests themselves is accurate, rather than making the example solutions we have saved as clean as possible. Going back and cleaning up the old examples hasn't exactly been a priority.

Thanks :)

patricksjackson:
Closes #227

PR #231, opened 2019-03-12
Cleaning up travis settings.

patricksjackson:
* Using ninja instead of make on travis to take advantage of multicore (travis machines have 2 cores, and ninja does interleave the output unlike `make -j2` would).
* Adding a modern compiler.
* Cleaning up how the compilers are set a little bit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ os: linux
 
 matrix:
   include:
-    # GCC 4.9
     - env: COMPILER=g++-4.9
       addons:
         apt:
@@ -14,7 +13,6 @@ matrix:
           packages:
             - g++-4.9
 
-    # Clang 3.8
     - env: COMPILER=clang++-3.8
       addons:
         apt:
@@ -24,7 +22,6 @@ matrix:
           packages:
             - clang-3.8
 
-    # Clang 7 (default on Xenial)
     - env: COMPILER=clang++-7
       addons:
         apt:
@@ -33,6 +30,14 @@ matrix:
             - llvm-toolchain-trusty-7
           packages:
             - clang-7
+
+    - env: COMPILER=g++-8
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-8
 
 before_install:
   - sudo add-apt-repository -y ppa:samuel-bachmann/boost
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,21 @@
 dist: trusty
 sudo: false
 language: cpp
+os: linux
 
 matrix:
   include:
     # GCC 4.9
-    - os: linux
+    - env: COMPILER=g++-4.9
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
           packages:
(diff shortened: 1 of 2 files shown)


KevinWMatthews:
Cool!

The default ninja settings do suppress the output for compiling individual files, but that probably makes it easier to read (and faster to run). A compilation error or test failure should still be apparent.

Is it worth trying to set up a more recent version of gcc, say 7 or 8?

.travis.yml
```diff
@@ -1,38 +1,48 @@
...
 
 before_install:
   - sudo add-apt-repository -y ppa:samuel-bachmann/boost
   - sudo apt-get -y -d update
-  - eval "${MATRIX_EVAL}"
 install:
-  - sudo apt-get -y install cmake3 libboost-test1.60-dev libboost-date-time1.60-dev
+  - sudo apt-get -y install cmake3 ninja-build libboost-test1.60-dev libboost-date-time1.60-dev
+  - export CXX=${COMPILER}
+  - ${CXX} --version
```
KevinWMatthews: Good idea.

Huh, it looks like the compiler settings don't take. See output for:
  * [g++ 4.9](https://travis-ci.org/exercism/cpp/jobs/505040922#L587) => `g++ 4.8.4`
  * [clang++ 3.8](https://travis-ci.org/exercism/cpp/jobs/505040923#L591) => `g++ 4.8.5`
  * [clang++ 7](https://travis-ci.org/exercism/cpp/jobs/505040924#L629) => `g++ 4.8.4`

I expected to see something like this for Clang (from my personal Linux box):
```bash
$ clang++-3.8 --version
clang version 3.8.0-2ubuntu4 (tags/RELEASE_380/final)
```

.travis.yml
```diff
@@ -1,38 +1,48 @@
 dist: trusty
 sudo: false
 language: cpp
+os: linux
 
 matrix:
   include:
     # GCC 4.9
```
KevinWMatthews: Comments on compiler version are no longer necessary - the settings themselves are descriptive.

patricksjackson:
ninja will print out lots of errors in the event of a build failure, and it'll print the build command and all the error output together instead of interleaved with the other build messages like make. Actually, I will push up a failure as an example.

I added gcc 8 as well, as that's reasonable.

.travis.yml
KevinWMatthews: I see - I was looking for the output in the wrong place.

patricksjackson: Yea, the travis output takes a bit of getting used to, it's not terribly obvious.

patricksjackson:
@KevinWMatthews do you think this is ok to merge?

KevinWMatthews:
@patricksjackson Looks good to me.

In a case like this, do you have a preference of who merges the PR?

patricksjackson:
If it's all finished, and there aren't some outstanding issues that need to be resolved or some sensitive issue, feel free to just merge it if you see it. I'm also considering trying to use Githubs actual Review and approval system to see if that works better, at least for larger changes.

I only pinged you again to double check you were good with the changes I'd made.

Issue #225, opened 2019-03-05
Setting up Boost_ARCHITECTURE variable in CMakeLists.txt


pr0minence:
I've a suggestion to explicitly set up Boost_ARCHITECTURE variable in CMakeLists.txt
```
if (CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
    set(Boost_ARCHITECTURE "-x64")
else()
    set(Boost_ARCHITECTURE "-x32")
endif()
```
This will allow CMake to find Boost libraries with "-x64" or "-x32" suffixes. (I've compiled both architectures and this helps me to avoid "Unable to find the requested Boost libraries" error).

KevinWMatthews:
@pr0minence Interesting - I think this is a use case that we didn't consider.

A few questions to help us craft a better solution:
  * What OS(es) are you using?
  * What version of CMake are you using?

Can you explain your specific setup a little bit more? Two machines, a virtual machine, etc. I'm curious to know how you're using both 64 and 32 bit libraries.

Also, can you provide the complete CMake output when it fails to find Boost?

Thanks!

pr0minence:
@KevinWMatthews, sure!

Machine specs:
```
Processor: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
Windows Version: Microsoft Windows 10 64-bit
```

I am using Cmake 3.14 and here is CMake output with error:
```
[cmake]   CMake Error at C:/Develop/CMake/share/cmake-3.14/Modules/FindBoost.cmake:2124 (message):
[cmake]   Unable to find the requested Boost libraries.
[cmake] 
[cmake]   Boost version: 1.69.0
[cmake] 
[cmake]   Boost include path: C:/Develop/Boost/include/boost-1_69
[cmake] 
[cmake]   Could not find the following static Boost libraries:
[cmake] 
[cmake]           boost_unit_test_framework
[cmake]           boost_date_time
[cmake] 
[cmake]   No Boost libraries were found.  You may need to set BOOST_LIBRARYDIR to the
[cmake]   directory containing Boost libraries or BOOST_ROOT to the location of
[cmake]   Boost.
[cmake]   Call Stack (most recent call first):
[cmake]   CMakeLists.txt:14 (find_package)
```

As I said before, I've compiled Boost libraries. For example:
```
libboost_unit_test_framework-mgw81-mt-x32-1_69.a
[...]

PR #232, opened 2019-03-15
Setting up Boost_ARCHITECTURE variable

pr0minence:
#225 Setting up this option gives the opportunity to CMake to try to find Boost libraries with '-x64-' / '-x32-' suffixes.


pr0minence:
Waiting for better times.

PR #234, opened 2019-03-20
Fix the unused variable warning in space age

HenryRLee:
Move the declaration of the 'accuracy' variable inside the
EXERCISM_RUN_ALL_TESTS section, so that when compiling the first test
case, the unused variable warning wouldn't show up.

### DIFF
--- a/exercises/space-age/space_age_test.cpp
+++ b/exercises/space-age/space_age_test.cpp
@@ -3,11 +3,6 @@
 #include <boost/test/unit_test.hpp>
 #include <cmath>
 
-namespace
-{
-const double accuracy = 0.005;
-}
-
 BOOST_AUTO_TEST_CASE(age_in_seconds)
 {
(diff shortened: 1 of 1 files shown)


KevinWMatthews:
Hi @HenryRLee,

Good point. I wonder if the original code compiles - I thought that we added the `-Werror` flag recently.

Will test this in detail soon.

HenryRLee:
Hi @KevinWMatthews,

That's right, now it becomes a compilation error.

patricksjackson:
Ah I see, when you're actually doing the test suite it isn't used in the first problem, as every place it is used is `#ifdef`'ed out. We only ever run the full test suite in CI, that's why we didn't catch it.

That _might_ be an error we hit in other tests too.. I'll open an issue to consider fixing it generally.

Thanks for this @HenryRLee! We don't often do the problems themselves, so contributions like this is how we keep the track working well for students :)

PR #241, opened 2019-03-25
Add more test cases in the sieve exercise

HenryRLee:
When the `limit` is a prime number, test if the answer includes the `limit`.

### DIFF
--- a/exercises/sieve/sieve_test.cpp
+++ b/exercises/sieve/sieve_test.cpp
@@ -2,7 +2,26 @@
 #define BOOST_TEST_MAIN
 #include <boost/test/unit_test.hpp>
 
-BOOST_AUTO_TEST_CASE(primes_up_to_10)
+BOOST_AUTO_TEST_CASE(no_primes_under_two)
+{
+    const std::vector<int> expected{};
+
+    const std::vector<int> actual = sieve::primes(1);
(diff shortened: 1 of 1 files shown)


HenryRLee:
I found that some students' solutions fail to include the `limit` in their answer when the `limit` is a prime number, hence the PR.

patricksjackson:
The canonical test cases were changed quite a while ago to include that test actually, so this really only happened because we are behind on the test version. Can I ask instead that you add a few more tests to get us fully updated to the new spec?

The test suite is here: https://github.com/exercism/problem-specifications/blob/master/exercises/sieve/canonical-data.json

HenryRLee:
@patricksjackson Sure thing. I just updated my PR.

patricksjackson:
Ah sorry, could you make the order the same as the problem spec? So that it checks up to 1 , then 2, then 10, then 1000? That just makes a lot more sense from the perspective of the student.

Everything else looks fine, just do the reorder (making sure to keep the EXERCISM_RUN_ALL_TESTS thing after the new first test) and it'll be good to go.

HenryRLee:
Sure! Fixed it.

patricksjackson:
Awesome. Thanks!

PR #242, opened 2019-03-26
Update the test case in the leap exercise

HenryRLee:
Since I found a student's solution was using 200 instead of 400, I think this might be a common case.

The test code we are using now is largely different from the canonical test suite. Because that code base changes quite frequently, I kept our test code, adding only one extra test case which I think is needed. Would be glad to sync the whole test code from the canonical test data if that is desired.

### DIFF
--- a/exercises/leap/leap_test.cpp
+++ b/exercises/leap/leap_test.cpp
@@ -2,34 +2,34 @@
 #define BOOST_TEST_MAIN
 #include <boost/test/unit_test.hpp>
 
-BOOST_AUTO_TEST_CASE(a_known_leap_year)
+BOOST_AUTO_TEST_CASE(not_divisible_by_4)
+{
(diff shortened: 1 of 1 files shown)


patricksjackson:
I prefer updating to match the canonical tests whenever we touch the test suite unless there's a good reason why it's a specific issue for C++. They update more frequently precisely because they get to crowdsource more of the edge cases than we could individually.

(Also by the way, since you were amending and force-pushing to your PR branch the last time. You don't have to do that if you don't want as the interface will squash them to one commit when it gets merged. And with multiple commits it's easier to see what changes were made over time.)

HenryRLee:
@patricksjackson Cool! I updated the test cases according to the canonical data.

patricksjackson:
:+1: Thanks!

Issue #233, opened 2019-03-17
Tracking Issue for Deprecating Boost


patricksjackson:
I believe we should move to make Boost a dependency for only a couple non-core exercises, while leaving the core of the track Boost-free. That would let us avoid having to help configure boost for every single platform and corner case, which just isn't going to happen.

For what we really needed Boost for, the unit test suite, we will switch to a header-only library so we can just distribute it along with the exercises, https://github.com/catchorg/Catch2 seems to be the most popular. Most platforms should then be able to compile problems without even bothering with CMake if necessary, since it'll just be a handful of self-contained files + the stdlib, and many IDEs have support for Catch2 specifically.

We will use version 2.7.0 of Catch, as that is the latest release as of today, located in a `test/` subdirectory in each exercise folder.

I initially setup #159 to remove everything at once, but it ended up having to touch too much of the repo (I didn't realize how often we used boost outside of the unit test suite itself) and it's now so old as to be far too difficult to merge.

Also, I'm older and a bit smarter, so I realize we can just do it in pieces so long as we let the students know in the documentation.

What I'm proposing now is that we take the following actions in roughly this order.

- [x] Remove Boost Regex - done #223 
- [x] Make Gigasecond and Meetup non-core exercises (some discussion in #224 already). These are the only exercises that pretty much require the use of boost libs, so we should make them non-core with extra documentation noting you'll need to install boost. - See #240
- [x] Update the documentation to explain we're in the process of switching, so some exercises may use Catch and not require installing boost. - See #239 
- [x] Update the exercises from Boost to Catch via PRs. A lot of which can be stolen from #159. - See #253
- [x] (optionally) Remove the uses of Boost algorithms from our examples (just cherry-pick https://github.com/exercism/cpp/commit/8922c1f274c18925ad9af73b77f362a5b5858f6d where I've already done that). This is only to make the examples more relevant for people who cheat and end up here. - See #238 
- [x] After all exercises have been updated, update the docs again.

Most of the grunt work is in updating the test suites. I will, at a minimum, put together an example moving the Hello World exercise over to Catch to demonstrate what to do for anyone who wants to help, and figure out what to do with an old script I had that helped me do a lot of the conversion work. There isn't a need to do the problems 1 at a time, but be somewhat reasonable as I do want 2 people to actually look at each of the changes.

If you'd like to grab some work for yourself, go ahead and claim it here or just open a PR.

patricksjackson:
[...]

PR #244, opened 2019-03-27
Moving all the core exercises to Catch2.

patricksjackson:
Drops Boost for Catch 2.7.0 in the core exercises (after #240 is merged). Depends on #239 getting merged first.

Related to #233.

### DIFF
--- a/exercises/grains/CMakeLists.txt
+++ b/exercises/grains/CMakeLists.txt
@@ -7,12 +7,6 @@ cmake_minimum_required(VERSION 3.1.3)
 # Name the project after the exercise
 project(${exercise} CXX)
 
-# Locate Boost libraries: unit_test_framework and date_time
-set(Boost_USE_STATIC_LIBS ON)
-set(Boost_USE_MULTITHREADED ON)
-set(Boost_USE_STATIC_RUNTIME OFF)
-find_package(Boost 1.59 REQUIRED COMPONENTS unit_test_framework date_time)
-
 # Get a source filename from the exercise name by replacing -'s with _'s
 string(REPLACE "-" "_" file ${exercise})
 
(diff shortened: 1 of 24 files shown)


elyashiv:
Looks nice. The errors also look quite clear. 
I will like to see how hard is it to migrate more complex test case. Two I can think of are handling scoped enums in triangle, and array compression in binary_tree.

patricksjackson:
@elyashiv That's a good idea, these are mostly only the simplistic test cases. I'll setup some of the more complex cases in a different PR.

Triangle is pretty easy, there should be only limited difference from my older changes - https://github.com/exercism/cpp/blob/remove_boost/exercises/triangle/triangle_test.cpp

For all the instances of BOOST_REQUIRE_EQUAL_COLLECTIONS, Catch instead relies on using REQUIRE statements and gives pretty good error messages ( It also has [Matchers](https://github.com/catchorg/Catch2/blob/master/docs/matchers.md) which also has some overlap with that, but those aren't likely to be too useful for the existing tests at least since we're comparing against a fully known output). Most places we use that are just comparing two containers, so we can switch to just using `REQUIRE(expected == actual)`. The trickiest compare is probably going to be the comparison using the iterator in binary_search_tree, which I should be able to do by just building a container out of the iterator. I'll mess around with it to figure out a good solution.

As an example of a failure, I edited one element of a vector and did a comparison to give this output (which is colored too):

```
exercises/anagram/anagram_test.cpp:42: FAILED:
  REQUIRE( expected == matches )
with expansion:
  { "stream", "maers" }
  ==
  { "stream", "maters" }

===============================================================================
test cases: 10 | 9 passed | 1 failed
assertions: 10 | 9 passed | 1 failed
```

PR #245, opened 2019-03-29
Binary tree better constructor

elyashiv:
As we don't now whether the parameter is easily moveable, passing by value is not a real option, so instead I pass by universal reference and perfect forward the arguments.

I need to constrain the template so I don't shadow the copy constructor and move constructor. This is not really needed because I deleted them, but it's better to have it in the example.

Should I get a variable length of arguments? Not needed by tests...

### DIFF
--- a/exercises/binary-search-tree/example.h
+++ b/exercises/binary-search-tree/example.h
@@ -3,10 +3,10 @@
 
 #include <memory>
 #include <cstdint>
+#include <utility>
 
 namespace binary_tree
 {
-    using std::move;
     template<typename T>
     class binary_tree final
     {
@@ -14,8 +14,9 @@ namespace binary_tree
         class binary_tree_iter;
         using binary_tree_ptr = std::unique_ptr<binary_tree>;
 
-        explicit binary_tree(T data)
-            : _data(move(data)),
+        template <typename TParam>
+        explicit binary_tree(TParam &&data)
+            : _data(std::forward<TParam>(data)),
               _left(nullptr),
               _right(nullptr)
         {}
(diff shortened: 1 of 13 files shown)


patricksjackson:
First, you'll need to branch this off of master instead of my other PR branch, there's no need to be dependent on it. (Feel free to close this and make a new PR if it's easier for you).

Second, let's look at [this set of guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#fcall-parameter-passing). Right now this only allows a moved argument, and I'm actually kind of surprised this compiles, but I'm not an expert at all the move semantics stuff. My surprise is the string tests, that pass in a `const string&` to the insert function, which then moves (?) that into a new binary search tree object. And I have no idea if you can reference the string object a second time in the tests. My _guess_ is that it copies it on one of those calls and doesn't actually move it.

So, I think you need a `const T&` constructor too to allow copies. And users who want to move in can std::move() in their types. Except then you have the problem with insert, which would always use the copy constructor because it has a `const T&` type... You could make _that_ take in a moved parameter too so then users would be free to move parameters into it using std::move. See for example [std::vector push_back](https://en.cppreference.com/w/cpp/container/vector/push_back) which uses that pattern.

So you are welcome to go down this rabbit hole, and I will happily help out and provide some input (I have only done this a few times, with lots of references), but it is beyond the scope of what's really needed. There is a reason why the advice is that you should really just avoid move semantics unless you need the performance.

(Ugh, I am so looking forward to Concepts rather than this enable_if stuff too).

elyashiv:
close #245. Moved to #246 that branches from master. My bad.

PR #246, opened 2019-03-30
Binary tree better constructor

elyashiv:


elyashiv:
I can't easily copy you message from the previous PR, but I will still address the points raised there. It might be easier to copy it...

I'm not sure what you mean about only allowing move arguments - the constructor gets a universal reference - it can bind to lvalue or rvalue refs. This way I make a tree in both ways with one function. I can also pass a string literal if I want a string... This is better that const ref because in some cases I don't need to call the copy ctor.

Looking at std::vector::push_back you can compare to std::vector::empalce_back - it uses my approach

I forgot to modify the insert function. It should probably also be written that way.

Got to go, I will write a little more in a different time. :smile:

patricksjackson:
You're right, I wasn't understanding the universal reference.

I'd argue that we really wouldn't need this in most cases, but I suppose you are trying to make this closer to a real container :)

PR #248, opened 2019-03-31
binary-search-tree: Onboard to Catch2.

patricksjackson:
This moves the only tricky test suite over to Catch, specifically the iterator test in `test_sort()`. 

FYI - @elyashiv 

Related to #233, dependent on #244.

### DIFF
--- a/exercises/binary-search-tree/binary_search_tree_test.cpp
+++ b/exercises/binary-search-tree/binary_search_tree_test.cpp
@@ -1,6 +1,6 @@
 #include "binary_search_tree.h"
-#define BOOST_TEST_MAIN
-#include <boost/test/unit_test.hpp>
+#define CATCH_CONFIG_MAIN
+#include "test/catch.hpp"
 #include <vector>
 
 // test data version: 1.0.0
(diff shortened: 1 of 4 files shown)


exercises/binary-search-tree/binary_search_tree_test.cpp
```diff
@@ -88,38 +88,42 @@ BOOST_AUTO_TEST_CASE(can_create_complex_tree)
 template<typename T>
 static void test_sort(const tree_ptr<T> &tree, const std::vector<T> &expected)
 {
-    BOOST_REQUIRE_EQUAL_COLLECTIONS(tree->begin(), tree->end(), expected.begin(), expected.end());
+    std::vector<T> actual;
```
elyashiv: I think it will be clearer if we write:
```cpp
std::vector<T> actual(tree->begin(), tree->end);
REQUIRE(expected == actual);
```
This way it's clear what functions you need to implement.

patricksjackson: Hmm, my thinking was they're going to have to look it up anyways, so why not show them you can build something that nicely uses the language features?

You make the call on this one and I'll switch it to your preference.

elyashiv: Agree. Lets start like that, and if students complain we can change it to be more explicit.

patricksjackson:
I'm adding a commit to move this to only be unlocked at the end of the core exercises, as it fairly obviously is the trickiest exercise we have. I'd rather not have people do 2 core exercises and then encounter this.

PR #249, opened 2019-04-02
Implement exercise - luhn

junming403:


junming403:
Implement new exercise luhn, ready for review.

config.json
```diff
@@ -531,6 +531,16 @@
...
       ]
+    },
+    {
+      "slug": "luhn",
+      "uuid": "885865bc-a197-436f-94bc-b1998d5cc081",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "strings"
```
patricksjackson: Please add "math" as a topic.

exercises/luhn/luhn_test.cpp
```diff
@@ -0,0 +1,152 @@
+#include "luhn.h"
+#define BOOST_TEST_MAIN
+#include <boost/test/unit_test.hpp>
+
+using namespace std;
```
patricksjackson: I'm pretty sure you don't use this?

exercises/luhn/example.cpp
```diff
@@ -0,0 +1,44 @@
+#include "luhn.h"
+
+using namespace std;
+
+namespace luhn{
+
+    bool valid(std::string const& input_str)
```
patricksjackson: Please run `clang-format --style="{BasedOnStyle: google, IndentWidth: 4}" -i example.cpp example.h` to fix the formatting.

Clang-format is installed with clang (or possibly llvm). If you don't have access to that I can do it myself after this is finished.

exercises/luhn/luhn_test.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+BOOST_AUTO_TEST_CASE(strings_with_non_digits_is_invalid)
+{
+    const bool actual = luhn::valid(":9");
+
+    const bool expected {false};
+
+    BOOST_TEST(expected == actual);
+}
+#endif
```
patricksjackson: nit: Please add an extra line after `#endif`. The rest of the test suite looks fine though :)

exercises/luhn/example.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+        int result = 0;
+        int counter=0;
+        for(auto c:reversed)
+        {
+            if(c==' ')
+            {
+                continue;
+            } else if(isdigit(c)) {
+                int digit = (int)c-48;
```
patricksjackson: Don't use magic values. Use `'0'` instead of `48`

exercises/luhn/example.cpp
```diff
@@ -0,0 +1,44 @@
...
+using namespace std;
+
+namespace luhn{
+
+    bool valid(std::string const& input_str)
+    {
+
+        string reversed;
+        reversed.assign(input_str.rbegin(),input_str.rend());
+        reversed.erase(0,reversed.find_first_not_of(" "));
```
patricksjackson: This just makes things more complicated when you can use `counter` to check the size at the end. (you also could just iterate through the string in reverse and it would work fine, using

```
    for (auto c = input_str.rbegin(); c != input_str.rend(); c++) {
```

exercises/luhn/example.cpp
```diff
@@ -0,0 +1,44 @@
...
+        for(auto c:reversed)
+        {
+            if(c==' ')
+            {
+                continue;
+            } else if(isdigit(c)) {
+                int digit = (int)c-48;
+
+                if(counter%2==1){
+                    result=digit*2>9?result+digit*2-9:result+digit*2;
```
patricksjackson: You should add some parentheses here to make this line more clear (even after the formatting gets fixed).

siebenschlaefer:
Currently all the valid numbers in the tests have an even number of spaces or all the digits are `0`. Therefore the following (invalid) attempt passes:
```cpp
bool luhn::valid(const std::string& str) {
    int sum = 0, digit_count = 0, factor = 1 + str.size() % 2;
    for (auto i = 0u; i < str.size(); ++i) {
        if (str[i] == ' ') continue;
        if (str[i] < '0' || '9' < str[i]) return false;
        ++digit_count;
        factor = 3 - factor;
        int digit = (str[i] - '0') * factor;
        if (digit > 9) digit -= 9;
        sum += digit;
    }
    return digit_count > 1 && sum % 10 == 0;
}
```

Please add a test where the the number is valid and it has an an odd number of spaces, e.g.
```cpp
BOOST_AUTO_TEST_CASE(valid_number_with_an_odd_number_of_spaces)
{
    const bool actual = luhn::valid("095 245  88");
    const bool expected {true};
    BOOST_TEST(expected == actual);
}
```

patricksjackson:
Hold on.  @junming403 please keep the original test that matches the problem specifications https://github.com/exercism/problem-specifications/blob/master/exercises/luhn/canonical-data.json

@siebenschlaefer that does seem like it might be a problem with the tests, however we do try to keep the test suites here identical the main problem specifications. Please change the canonical test suite in the problem-specifications repo so all the other tracks can benefit too :)

siebenschlaefer:
@patricksjackson Thanks, I didn't realize the rules are that tight. I submitted a PR to the problem-specifications repo: https://github.com/exercism/problem-specifications/pull/1500

patricksjackson:
@siebenschlaefer It's not really that the rules are tight, it's that I don't want to be in a position to maintain a good set of tests, so it's easier to let the central team do it. There are actually _some_ places where we might have an extra test for a C++ specific thing, but those are the only exceptions we generally make.

@junming403 I'll come back and review your updates when I get the chance, it might take until the weekend.

exercises/luhn/example.cpp
```diff
@@ -0,0 +1,32 @@
...
+
+namespace luhn {
+bool valid(string const& input_str) {
+    int result = 0;
+    int counter = 0;
+    for (auto c = input_str.rbegin(); c != input_str.rend(); c++) {
+        if (*c == ' ') {
+            continue;
+        } else if (isdigit(*c)) {
+            int digit = (int)*c - '0';
```
elyashiv: digit should be const

exercises/luhn/example.cpp
```diff
@@ -0,0 +1,32 @@
+#include "luhn.h"
+
+using namespace std;
+
+namespace luhn {
+bool valid(string const& input_str) {
```
elyashiv: We should have a standard way of writing this... I don't mind if it's `const string` or `string const`, but there should be a single way through the whole repo.

exercises/luhn/example.cpp
elyashiv: Can we make travis fail if the formatting is wrong?

exercises/luhn/example.cpp
patricksjackson: I believe we've been pushing everything to use `string const&` for reference inputs, though `const int` is used for local variables, so this should be fine.

Yes it's somewhat inconsistent, but this style is consistent with the rest of the repo.

exercises/luhn/example.cpp
patricksjackson: Yes we can. Getting us to standardize on formatting and running it on a git hook and in travis has been on my backlog, but it's lower pri than the boost stuff.

junming403:
Updated according to review

patricksjackson:
Thanks!

PR #252, opened 2019-04-05
use perferct forwarding also for insert

elyashiv:
Insert can also benefit from the perfect forwarding.

Also, I cleaned up the code a little.

I guess the example doesn't need to be perfect, but I will want to use it for the mentoring notes.

### DIFF
--- a/exercises/binary-search-tree/example.h
+++ b/exercises/binary-search-tree/example.h
@@ -7,6 +7,11 @@
 
 namespace binary_tree
 {
+    template <bool B, class T = void>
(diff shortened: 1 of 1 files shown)


patricksjackson:
Does this change leave the code open to the same drawbacks as `emplace_back`? See https://abseil.io/tips/112

elyashiv:
You mean you might call some constructor instead of moving the object into the container? yes. The code does almost the same thing emplace_back does. The only difference is the number of arguments - emplace back takes any number, and my insert function only one.

PR #263, opened 2019-04-10
Adding test_pattern.

patricksjackson:
This adds a test_pattern to config.json so that only the test suite shows up on exercism.io rather than the full copy of Catch.

### DIFF
--- a/config.json
+++ b/config.json
@@ -2,6 +2,7 @@
   "language": "C++",
   "active": true,
   "blurb": "C++ is a general-purpose programming language that supports procedural, object-oriented, generic and functional programming styles. C++ is deployed on billions of devices from the smallest embedded microprocessor to the largest supercomputer.",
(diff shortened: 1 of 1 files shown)


patricksjackson:
Actually I have no clue how to test this without just merging it and quickly fixing it if I broke something.

(Also, I ran configlet fmt on this and it seems I had messed up the config.json file for the 2nd or 3rd time in the last month. I'll add a configlet formatting test to the CI at some point, which we can steal from the rust repo)

patricksjackson:
We successfully merged a more complicated change into the C track in exercism/c#377, so I don't expect this to have issues.

I'm headed out the door right now, but I will just merge this myself when I get a few minutes to baby sit it in case it breaks.

patricksjackson:
Looks like it works fine :+1:

KevinWMatthews:
This looks awesome. Fixes the Catch issue on my mobile phone.

PR #266, opened 2019-04-13
Adding partial builds to Travis

patricksjackson:
This switches the Travis builds so that it tests different subsets of the exercises depending on whether it is a PR build or a branch build, and on the changes made compared to the master branch. Basically, for a normal PR this will help lighten the load while it's being worked on, and we will run a full build against master once it is merged.

Some other tracks do this already, I've borrowed from the script used for Rust but heavily modified it.

If it's a branch build we run everything, as we currently do.
If it's a PR build that changes no exercises, we run nothing.
If it's a PR build that changes 1 or more exercises, _we only run the changed exercises._
If it's a PR build that changes the _root_ CMakeLists.txt file, we run everything to make sure that works. [1]

I'll be kicking off builds on Travis to test all the paths before this is ready to be merged.

[1] - This also changes the CMakeLists.txt file so it reads the directories instead of taking in a list of exercises, so it doesn't need to be edited just to add an exercise. It'll only be changed if there are real changes to it.


patricksjackson:
So this now has examples for all the different paths:

Building everything in a branch (from my fork): https://travis-ci.org/patricksjackson/xcpp/builds/519523367
PR - building nothing (first build here)
PR - successfully building subset (second build here)
PR - failure when building subset (third build here)
PR - Running everything when root CMake changed (last build here)

bin/check-exercises.sh
```diff
@@ -0,0 +1,31 @@
...
+if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+    echo "This is not a Pull Request, running all tests."
+    cmake --build "$repo"
+elif git diff --name-only master | grep "^CMakeLists.txt"; then
+    echo "This PR changes the root CMakeLists.txt."
+    echo "Running a single test and then all tests."
+    cmake --build "$repo" -- test_hello-world
+    cmake --build "$repo"
+else
+    tests="$(git diff --name-only master | grep "exercises/" | cut -d '/' -f2 | sed 's/.*/test_&/' | tr '\n' ' ')"
```
HenryRLee: Can we use `grep "^exercises/"` here, to be more precise?

Moreover, it's possible that multiple files being edited in the same exercise path, resulting in the same exercise name occurs multiple times. So I think we can add a `|uniq ` before piping into the `|tr` command.

patricksjackson: Good points all around. I've made both suggested changes :)

Neither changes the function of the script, but they do make it more clear. Thanks!

(CMake/Ninja ignore multiple targets in the build list)

patricksjackson:
I'm going ahead and merging this, at it will save us compilation time and help with #272.

PR #267, opened 2019-04-14
Boost->Catch: Final Docs update

patricksjackson:
This is a decent sized doc update adding better Windows instructions and taking into account all the changes being made with the switch to Catch2.

I'll plan to merge this once all the exercises are updated, planned for April 20th.

Related to #233

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -4,15 +4,18 @@ The C++ language track requires that you have the following software
 installed on your system:
 * a modern C++11 compiler
 * the CMake cross-platform build system
-* the Boost libraries
 
-Microsoft's Visual Studio IDE and Apple's Xcode IDEs combine the compiler
-and the
-native build system into a single IDE (integrated development environment).
-Linux environments typically expose the compiler and native build system
-as separate command-line tools.
+We use [Catch2](https://github.com/catchorg/Catch2) for testing, which is a
+header-only library that is bundled along with the exercises.
 
-### Prerequisite: A Modern C++11 Compiler
+This set of dependencies is very common for C++ projects, so if you have any
+issues setting it up you will find many helpful resources online.
+
+Many IDE's, including Microsoft's Visual Studio, Apple's Xcode, and Jetbrain's
+Clion bundle a compiler and CMake together, so you should only need to install
+the IDE. 
+
+### A Modern C++11 Compiler
 
 This language track requires a compiler with [C++11](http://en.wikipedia.org/wiki/C%2B%2B11)
 support, which was released in 2011. All major compilers released in the last few years should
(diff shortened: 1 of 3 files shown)


docs/INSTALLATION.md
```diff
@@ -74,99 +47,41 @@ improve the CMake support.
...
-Linux users can usually get Boost from your distribution's package manager.
-If a pre-built package is unavailable for your linux distribution, try
-[installing Boost from source](http://www.boost.org/doc/libs/release/more/getting_started/index.html).
+## Linux
 
-Alternately, there may be some other way to install Boost for your
-distribution without installing from source. For example, you can install
-Boost 1.60 for Ubuntu 16.04 LTS by using a PPA.
+Ubuntu 16.04 and later have compatible compilers in the package manager, so
+installing the necessary compiler can be done with `sudo apt-get install gcc`.
```
KevinWMatthews: Consider highlighting `sudo apt-get install gcc` by moving it to a separate block.
One's eye is drawn to the more complex and frequently unnecessary installations below simply because they take up more real estate.

docs/INSTALLATION.md
```diff
@@ -4,53 +4,26 @@ The C++ language track requires that you have the following software
...
-as separate command-line tools.
+We use [Catch2](https://github.com/catchorg/Catch2) for testing, which is a
+header-only library that is bundled along with the exercises.
 
-### Prerequisite: A Modern C++11 Compiler
+This set of dependencies is very common for C++ projects, so if you have any
+issues setting it up you will find many helpful resources online.
+
+Many IDE's, including Microsoft's Visual Studio, Apple's Xcode, and Jetbrain's
+Clion bundle a compiler and CMake together, so you should only need to install
```
KevinWMatthews: Capital 'B', trailing apostrophe, and a capital 'L': `JetBrains' CLion`
But no one is really watching ;)

KevinWMatthews:
One nit and one super-nit ;)
Seems pretty good on the whole. Time will tell?
Links worked for me.

docs/INSTALLATION.md
patricksjackson: I rewrote this little section to handle the common case first, since Ubuntu 16.04 is super old, and separated this command out to it's own segment. I think it's clearer now.

patricksjackson:
Thanks! Yea, most of what I'm touching here is just removing boost and explaining Catch. I expect that the installation instructions will get a lot less use anyways without Boost, there isn't really anything to install anymore.

patricksjackson:
Oh, I should also mention in here - I think the docs are supposed to move over to exercism/website-copy at some point, though I'm not entirely sure if _these_ docs are going to move. Basically though someone will tell us if that happens.

KevinWMatthews:
>  I think the docs are supposed to move over to exercism/website-copy at some point

Ok. So far I only see generic exercism docs and mentoring notes on exercism/website-copy. These docs are "copy", I guess, so I can see that we'll add them to the `tracks` directory at some point.

PR #271, opened 2019-04-20
Separate Catch Object

patricksjackson:
This adds a separate compilation object for the Catch main configuration. See https://github.com/catchorg/Catch2/blob/master/docs/slow-compiles.md for an explanation from Catch.

In short, including the Catch header is relatively heavy but not too bad, but most of the compilation time for Catch comes from compiling it with support for main, as that actually causes a compile of the Catch functionality and not just of the interface. By separating this out into it's own object, we ensure that students only need to build the object once per exercise, rather than every time they modify the test.cpp file or their exercise header. This was discussed in the comments of #233.

This also revealed an issue of a missing include in the clock test, which is fixed here.

This was done with the following script:

```fish
cd exercises/
ls */*_test.cpp | xargs sed -i '/CATCH_CONFIG_MAIN/d'
ls */CMakeLists.txt | xargs sed -i 's/${file}.h/${file}.h\ test\/tests-main.cpp)/'
for x in *
  echo "#define CATCH_CONFIG_MAIN" >> $x/test/tests-main.cpp
  echo "#include \"catch.hpp\"" >> $x/test/tests-main.cpp
end
```

### DIFF
--- a/exercises/acronym/CMakeLists.txt
+++ b/exercises/acronym/CMakeLists.txt
@@ -18,7 +18,7 @@ else()
 endif()
 
 # Build executable from sources and headers
-add_executable(${exercise} ${file}_test.cpp ${exercise_cpp} ${file}.h)
+add_executable(${exercise} ${file}_test.cpp ${exercise_cpp} ${file}.h test/tests-main.cpp)
 
 set_target_properties(${exercise} PROPERTIES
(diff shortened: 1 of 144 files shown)


KevinWMatthews:
Looks reasonable. Fixes the user experience first in a straightforward fashion.

PR #276, opened 2019-04-21
Using a common Catch main object for development.

### LINKED ISSUE #272: Speed up CI builds
I've discussed this in various messages, collecting here to track this cleanup work.

@KevinWMatthews and I discussed in #233 the problems and some solutions for the fact that Catch2 has very slow compilation speed. I also mention it in #253. I explained in https://github.com/exercism/cpp/issues/233#issuecomment-482856884 the problems in more detail.

I've already setup #271 to improve the student user experience.

To reduce our resources in CI so we don't impact exercism as a whole, we should also implement the changes proposed by @KevinWMatthews  in https://github.com/exercism/cpp/issues/233#issuecomment-483082792, using a common Catch main object for all the exercises. This, as I experimented with in https://github.com/exercism/cpp/issues/233#issuecomment-483844808, will greatly speed up the builds.

I propose then that we implement those changes, after making sure there isn't something we've overlooked, and use the common Catch main object as the _default_ for compiling from the root, and setup Travis to only compile every object separately for the modern Clang compiler setting. There may be some work to combine this logic with the partial builds implemented in #266.

This isn't really urgent, but it would be a nice improvement to be better citizens for Exercism and Travis.

patricksjackson:
This needed a few small changes from my hacked together test. Mostly I hit an issue where I couldn't create a `catchlib` object in every exercise during a full CMake build, so I had to switch how I did the individual exercise CMakeLists.txt files.

I added some comments into the students' CMakeLists files, but they may need to be phrased better to be clearer to them. I wanted to explain what it was instead of just confusing the student if they looked at it.

Closes #272


KevinWMatthews:
Ah, yes - you would need to name each library uniquely. Normally that's no issue, but in our case that is much harder to maintain.

I think that we're correct to omit special compiler flags or definitions from catchlib: `-Wall -Wextra -Wpedantic -Werror` or `_SCL_SECURE_NO_WARNINGS`. Those are useful in student code but not for this.

Looks pretty good at a read. Hope to try this in detail this evening.

KevinWMatthews:
I'm getting failures when I build individual exercises in isolation (using only the exercise-level `CMakeLists.txt`, not using the root-level `CMakeLists.txt`). Can you reproduce this:
```bash
CMake Error at CMakeLists.txt:26 (add_executable):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:catchlib>

  Objects of target "catchlib" referenced but no such target exists.
```

KevinWMatthews:
I think it's a quirk of CMake conditionals. When building from the root level, `EXERCISM_COMMON_CATCH` exists and is Boolean. When building at the exercise level, `EXERCISM_COMMON_CATCH` does not exist at all.

Reading through CMake's [if()](https://cmake.org/cmake/help/v3.1/command/if.html) documentation (no really, there's a big page), I see:

```cmake
if(NOT <expression>)
  # True if the expression is not true.
```
This fails when the expression does not exist.

I think that [variable references](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#variable-references) evaluate to an empty string if the value is unset.

A standard `if()` behaves differently:
```cmake
if(<constant>)
  # [...]  False if the constant is [...] the empty string [...].
```

KevinWMatthews:
Short story - got off in the weeds above.

I think we need to reverse the order of the conditional for isolated exercises to build properly:
```cmake
# Use the common Catch library?
if(${EXERCISM_COMMON_CATCH})
    # For Exercism track development only
    add_executable(${exercise} ${file}_test.cpp ${exercise_cpp} ${file}.h $<TARGET_OBJECTS:catchlib>)
else()
    # Build executable from sources and headers
    add_executable(${exercise} ${file}_test.cpp ${exercise_cpp} ${file}.h test/tests-main.cpp)
endif()
```
because `if (NOT <expr>)` behaves differently that `if (<expr>)`.

KevinWMatthews:
There are some CMake semantics here. I see this succeed:
```cmake
if(NOT EXERCISM_COMMON_CATCH)
```

this fail:
```cmake
if(NOT ${EXERCISM_COMMON_CATCH})
```

and this succeed:
```cmake
if(NOT "${EXERCISM_COMMON_CATCH}")
```

Do you see something similar? This is when building as a student would - only one specific `CMakeLists.txt` file.

patricksjackson:
I alwyas like to work with CMake occasionally to remind me how much I hate it. That test is identical to the one elsewhere in the file that tests the run all tests option, except it has a `NOT` in it.

I think the problem is CMake for evaluates the branch in some instances, even if it's not taken? No idea.

Regardless, I added a test to CI to run it from an individual directory, and updated the exercise CMakeLists to what you suggested actually works.

Also changing the title to WIP, I need to test this on Visual Studio because I now don't trust this will work everywhere, I'll change it back after confirming.

patricksjackson:
Oh, I need to fix how I did the testing of an individual test anyways.

patricksjackson:
Works fine in Visual Studio too.

I added running the individual test in CI to #279, since it requires a compilable hello-world anyways.

KevinWMatthews:
> I alwyas like to work with CMake occasionally to remind me how much I hate it.

There's always Autotools ;)

> That test is identical to the one elsewhere in the file that tests the run all tests option, except it has a NOT in it.

I don't know for certain, but I think that `if (NOT)` has different variable expansion rules than `if ()`. Counterintuitive, I agree.

CMakeLists.txt
```diff
@@ -19,6 +19,16 @@ function(travis_fixup exercise_dir alt_exercise_root)
 endfunction()
 
 option(EXERCISM_RUN_ALL_TESTS "Run all Exercism tests" On)
+option(EXERCISM_COMMON_CATCH "Link against a common Catch2 main lib." On)
+
+if(${EXERCISM_COMMON_CATCH})
```
KevinWMatthews: Is it worth removing the `${}` and changing this to:
```cmake
if(EXERCISM_COMMON_CATCH)
```
in the hopes of avoiding a future "gotcha"? It would match the others.

.travis.yml
```diff
@@ -22,7 +22,9 @@ matrix:
           packages:
             - clang-3.8
 
-    - env: COMPILER=clang++-7
+    - env:
+      - COMPILER=clang++-7
+      - EXERCISM_COMMON_CATCH=false
```
KevinWMatthews: Looks good. `EXERCISM_COMMON_CATCH` is enabled by default.
We use the fastest compiler to test the students' setup.

CMakeLists.txt
patricksjackson: 👍

patricksjackson:
I missed that this was actually all finished apparently. I've updated after #284 and am merging.

> > I alwyas like to work with CMake occasionally to remind me how much I hate it.

> There's always Autotools ;)

Thankfully I've never had to touch Autotools, but it scares me anyways.

PR #279, opened 2019-04-27
Improves Hello-World

### LINKED ISSUE #273: Improve hello-world
Setting up hello-world has been an issue for people. We send the header file to them, but they need to figure out how to setup a hello_world.cpp file correctly to even get it to compile.

This is a problem since the error messages are usually really bad, because we allow there to not be a cpp file and it thus fails while trying to link. New people don't have a clue how to fix that issue.

I think we should improve hello-world so that it:

1. Compiles without the student making any code changes, but fails the test.
2. Contains commented explanations for what things are.

patricksjackson:
Closes #273 

This adds lots of comments and a compilable example for the hello world exercise. It also adds a CI test to validate that assertion.

### DIFF
--- a/exercises/hello-world/hello_world.cpp
+++ b/exercises/hello-world/hello_world.cpp
@@ -1,0 +1,16 @@
+#include "hello_world.h"
+
+// Use everything from the 'std' namespace.
+// This lets us write 'string' instead of 'std::string'.
+using namespace std;
+
+namespace hello_world {
+
+// Define the function itself. This could have also been written as 'std::string
+// hello_world::hello()'.
+string hello() {
+    // Return the string we need.
+    return "Hello!";
+}
+
+}  // namespace hello_world
--- a/exercises/hello-world/hello_world.cpp
+++ b/exercises/hello-world/hello_world.cpp
@@ -6,11 +6,11 @@ using namespace std;
 
 namespace hello_world {
 
-// Define the function itself. This could have also been written as 'std::string
-// hello_world::hello()'.
+// Define the function itself. This could have also been written as:
+// std::string hello_world::hello()
 string hello() {
     // Return the string we need.
-    return "Hello!";
+    return "Fix me!";
 }
 
 }  // namespace hello_world
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ cmake_install.cmake
 
 # Ignore alternate directory of exercises that is used for building
 build_exercises/
+build_hello-world/
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,4 +50,5 @@ script:
   - bin/fetch-configlet
   - bin/configlet lint .
   - bin/check-configlet-fmt.sh
+  - bin/check-hello-world.sh
   - bin/check-exercises.sh
--- a/bin/check-hello-world.sh
+++ b/bin/check-hello-world.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+#
+# Test that the hello-world exercise compiles with no changes and fails.
+
+# Fail if any command fails
+set -e
+
+repo=$(cd "$(dirname "$0")/.." && pwd)
+
+hello_world_tmp_dir="$repo"/build_hello-world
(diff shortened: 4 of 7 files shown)


elyashiv:
I wander - are we teaching c++ from ground up, or are we improving the coding of someone that's learning c++ in some other manner? Some of the comments are very verbose - I hope I don't need to explain what's `using`, or that the implementation and declaration are separate... What do we expect the student to know already?

patricksjackson:
The vast majority of people using Exercism track seem to already be somewhat proficient in the language already. The changes here are not really for them, and they should know that. We don't even mentor the Hello World exercise AFAIK, because the point is just to demonstrate how the track works.

Many people use it as a starting point to learn a new language though, and for those people the hello world exercise demonstrates how exercises are structured and some basics of the language. I've been overly verbose in the vein of https://learnxinyminutes.com/docs/c++/ so they can learn what features of the language you really have to know to understand the problems.

Mostly though, this change is meant to improve the on-ramp for people who are entirely new to the language. The goal is to get something that compiles, so they know when they've set it up correctly, and to demonstrate how the problems are structured. There are a number of people who benefit from that. I'm pretty sure most of the issues that students open in this repo are just about setting up Hello World.

(Actually, I should really change the failing "Hello!" to "Change me!" or something to make the error messages clear.)

Other tracks have actually spent time ordering their exercises in a way that introduces new concepts in the language logically, so they're actually better for learning the language from scratch. There are some higher level efforts in that area too - see https://exercism.io/blog/track-anatomy-project

exercises/hello-world/hello_world.cpp
```diff
@@ -0,0 +1,16 @@
+#include "hello_world.h"
+
+// Use everything from the 'std' namespace.
+// This lets us write 'string' instead of 'std::string'.
+using namespace std;
+
+namespace hello_world {
+
+// Define the function itself. This could have also been written as 'std::string
+// hello_world::hello()'.
```
KevinWMatthews: It might help to put
```c++
std::string hello_world::hello()
```
all on a single line.

exercises/hello-world/hello_world.cpp
```diff
@@ -0,0 +1,16 @@
...
+string hello() {
+    // Return the string we need.
+    return "Hello!";
```
KevinWMatthews: As you suggested, `"Change me!"` sounds better.
Alternatives are `"Fix me!"` or `"Start here"`, but `"Change me!"` is direct.

exercises/hello-world/hello_world.cpp
patricksjackson: Done

exercises/hello-world/hello_world.cpp
patricksjackson: Done, I went with "Fix me!"

exercises/hello-world/hello_world.cpp
```diff
@@ -0,0 +1,16 @@
+#include "hello_world.h"
+
+// Use everything from the 'std' namespace.
+// This lets us write 'string' instead of 'std::string'.
+using namespace std;
```
Ben1980: Is that really a good idea? This is actually suggesting to use `using namespace std;` all the time. Not a good coding style

patricksjackson: People tend to disagree on this point - See [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf6-use-using-namespace-directives-for-transition-for-foundation-libraries-such-as-std-or-within-a-local-scope-only) and [Google](https://abseil.io/tips/153).

It's generally safe to use in a small, self-contained source file, though there are potential issues in large codebases. It is one of those conveniences that I see as being very helpful while learning the language and potentially dangerous later. You're going to almost never hit issues from it in a toy problem.

That said, I mostly use this in toy problems because I'm lazy and it's low- to no-risk. Do you think it would be better to encourage new learners to use `using std::string;` instead of the whole namespace?

patricksjackson:
I'm going go go ahead and merge this, I think this will be an improvement over the existing at least.

PR #281, opened 2019-05-19
Fixes #280. Includes unary number system.

### LINKED ISSUE #280: all-your-base has an incorrect test case
The [test case](https://github.com/exercism/cpp/blob/master/exercises/all-your-base/all_your_base_test.cpp#L105) `leading_zeros` is incorrect in my opinion.
Also, I think it might be better that the invalid cases throw an std::error instead of returning 0.

anupamsobti:
Kindly review changes.


patricksjackson:
I'm going to close this, we can reopen if the decision is made to add this.

PR #282, opened 2019-05-19
Fixes #280. Problem spec intact.

### LINKED ISSUE #280: all-your-base has an incorrect test case

anupamsobti:
Fixes #280 without modifying base 1 behavior. Fixes `leading_zeros` test case and requires `std::error` for test cases where the base is invalid.

### DIFF
--- a/exercises/all-your-base/all_your_base_test.cpp
+++ b/exercises/all-your-base/all_your_base_test.cpp
@@ -3,15 +3,8 @@
 
 using namespace std;
 
-TEST_CASE("test1") {
-    vector<unsigned int> in_digits{0};
-    vector<unsigned int> out_digits = all_your_base::convert(10, in_digits, 2);
-    vector<unsigned int> expected{};
+//all-your-base exercise test case data version 2.3.0
 
-    REQUIRE(expected == out_digits);
-}
-
-#if defined(EXERCISM_RUN_ALL_TESTS)
 TEST_CASE("single_bit_one_to_decimal") {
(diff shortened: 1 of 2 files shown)


exercises/all-your-base/all_your_base_test.cpp
```diff
@@ -102,50 +102,35 @@ TEST_CASE("multiple_zeros") {
...
     REQUIRE(expected == out_digits);
 }
 
 TEST_CASE("first_base_is_one") {
     vector<unsigned int> in_digits{};
-    vector<unsigned int> expected{};
-    vector<unsigned int> out_digits = all_your_base::convert(1, in_digits, 10);
-
-    REQUIRE(expected == out_digits);
+    REQUIRE_THROWS_AS(all_your_base::convert(1, in_digits, 10), std::invalid_argument);
```
patricksjackson: Please add a blank line above this line in each of the test cases, just to match the visual style of the other tests.

patricksjackson:
Oh, also you need to fix the example solution so that it passes the new tests.

anupamsobti:
Made the changes. Kindly review.

patricksjackson:
Looks good 👍  Thanks!

PR #283, opened 2019-05-19
Luhn: Add a test with an odd number of spaces

siebenschlaefer:
In version 1.6.0 of the canonical-data.json in the
problem-specifications repository a new test case was added.

This commit just incorporates that test into the C++ test suite.


patricksjackson:
Could you update the test version at the top of that file?

(Also, though it is somewhat bad practice, could you fix the formatting elsewhere in the file with a blank line between lines 95/96? It is unrelated but whenever I see updates to the test files I try to make sure there aren't minor issues with the rest of the file too.)

siebenschlaefer:
Shoot, I missed that version number in the comments.
Thanks for catching that.

patricksjackson:
Great. Thanks!

PR #287, opened 2019-06-02
grade-school: Force const on roster() and grade()

siebenschlaefer:
Currently school::roster() and school::grade() can be implemented as
non-const methods. Logically it makes sense to be able to call those two
methods on a const instance because they should not modify that
instance.

### DIFF
--- a/exercises/grade-school/grade_school_test.cpp
+++ b/exercises/grade-school/grade_school_test.cpp
@@ -6,7 +6,7 @@ using namespace std;
 
 TEST_CASE("a_new_school_has_an_empty_roster")
 {
-    grade_school::school school_;
+    const grade_school::school school_;
     REQUIRE(school_.roster().empty());
 }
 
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
I accidentally deleted my previous (identical) PR.

patricksjackson:
The CI is failing because of [this](https://stackoverflow.com/questions/7411515/why-does-c-require-a-user-provided-default-constructor-to-default-construct-a). We test against Clang 3.8 since last I checked it was still the default in something (probably an Ubuntu LTS build, 16.XX?). For now we should support that until we update the compilers, so you'd need to add a default constructor to the example.

I would suggest actually adding a brief comment to the added constructor to explain why there is otherwise unnecessary code.

This is a good change though :)

siebenschlaefer:
Good catch!
I could add a defaulted default constructor (as you suggested) or I could could use value initialization for those three `const grade_school::school` instances (see https://godbolt.org/z/Mrx2uh).
Which do you prefer?

patricksjackson:
Wow that's...interesting.  Let's do value initialization as you suggested so we don't force the students to deal with that.

Issue #237, opened 2019-03-23
Exercise Ordering


patricksjackson:
I'm not sure if the exercise ordering makes much sense. The core exercises and the exercise ordering was setup in #187, and they may need to be updated or changed. (I know of at least a couple changes I want to see happen, which I'll list here).

We're looking for someone to take on this task, figuring out if the ordering makes sense now after it's been active for several months, and possibly reordering many of the exercises. This would work best if we got feedback from the mentors for the track, and possibly from current or recent students.

I can help and provide input, but I may not have time to take this on myself. And I haven't mentored or done any exercises on exercism.io since they launched the new version of the website, so I don't personally have great context for what would make sense.

The one thing I want to happen is to make Gigasecond and Meetup not part of the core track to help with #233. I may end up making just that change myself if necessary, but the rest of the exercises need to be looked at as well.

KevinWMatthews:
I agree with @patricksjackson 's concerns about Gigasecond and Meetup - they become much more difficult without Boost libraries. They could be discouraging/blocking for students.

Gigasecond is currently the [third core exercise](https://github.com/exercism/cpp/blob/a5a0983b205304a714f25e457630d7c035ccd01c/config.json#L54) in the track.

Meetup looks to be an [extra exercise](https://github.com/exercism/cpp/blob/a5a0983b205304a714f25e457630d7c035ccd01c/config.json#L212) that is unlocked by Space Age.

Mentor feedback would be excellent. You are the ones with experience!

NobbZ:
Before doing random changes to the track, it might be worth to ask Maud in the #project-track-anatomy channel on the slack, whether or not she can take you into her beta program.

She is developing some tool that should help with big track restructurings.

KevinWMatthews:
@NobbZ Thanks! That's worth looking into.

patricksjackson:
I'm setting up a list for the Catch conversions, and it might help here as well. This is the current list of exercises and how we have them organized.

Unlocked by nothing:
- acronym
- all-your-base
- binary
- hexadecimal
- isogram
- pascals-triangle
- robot-simulator
- scrabble-score

Core track and what they unlock
- hello-world
    - grade-school
    - hamming
    - nucleotide-count
    - rna-transcription
- leap
    - difference-of-squares
    - nth-prime
    - queen-attack
    - say
    - sieve
- reverse-string
    - allergies
    - gigasecond
    - prime-factors
    - robot-name
    - sum-of-multiples
- space-age
    - beer-song
[...]

PR #288, opened 2019-06-04
Implement exercise - circular-buffer

marvelou-s:
Hi!
Ported [circular-buffer](https://github.com/exercism/problem-specifications/tree/master/exercises/circular-buffer) to C++ track.


exercises/circular-buffer/circular_buffer_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+    circular_buffer::circular_buffer buffer(1);
+
+    REQUIRE_NOTHROW(buffer.write(1));
+
+    int expected = 1;
+    REQUIRE(expected == buffer.read());
+
+    REQUIRE_NOTHROW(buffer.write(2));
+    expected = 2;
+    REQUIRE_NOTHROW(expected == buffer.read());
```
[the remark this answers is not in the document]
marvelou-s: You're absolutely right, thank you! I missed that.

config.json
```diff
@@ -555,6 +555,18 @@
         "loops",
         "math"
       ]
+    },
+    {
+      "slug": "circular-buffer",
+      "uuid": "92788dbc-feb2-43a6-99c0-af21ce318b01",
+      "core": false,
+      "unlocked_by": null,
```
patricksjackson: Let's change `unlocked_by` to "raindrops" instead of null. That will be a decent place for it until #237 is addressed.

Everything else should be the same.

exercises/circular-buffer/circular_buffer_test.cpp
```diff
@@ -0,0 +1,187 @@
+#include "test/catch.hpp"
+#include "circular_buffer.h"
+#include <stdexcept>
+
+// Circular-buffer exercise test case data version 1.1.0
+
+using value_type = int;
```
patricksjackson: I would not bother doing this, just use `int` directly.

exercises/circular-buffer/circular_buffer_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+    expected = 3;
+    REQUIRE(expected == buffer.read());
+
+    expected = 4;
+    REQUIRE(expected == buffer.read());
+
+    expected = 5;
+    REQUIRE(expected == buffer.read());
+}
+#endif  // !EXERCISM_RUN_ALL_TESTS
```
patricksjackson: Add a new test for using something other than an int. Easiest thing is to just use a `string` type and repeat the final test case.

Instead of using integers, just steal some of the words from here: https://github.com/exercism/cpp/blob/master/exercises/anagram/anagram_test.cpp

exercises/circular-buffer/example.cpp
```diff
@@ -0,0 +1 @@
+#include "circular_buffer.h"
```
patricksjackson: You actually don't need this file at all, just delete it.

exercises/circular-buffer/example.h
```diff
@@ -0,0 +1,54 @@
...
+    void clear() { head_ = 0; tail_ = 0; }
+
+private:
+    std::vector<ValueType> buffer_;
+    std::size_t head_;
+    std::size_t tail_;
+
+    void insert(ValueType item);
+    bool is_empty() const { return head_ == tail_; }
+    bool is_full() const { return head_ == (tail_ + 1) % (buffer_.size() + 1); }
```
patricksjackson: There's a bug here, though it's not caught in the test suite. This test should pass: 

``` cpp
TEST_CASE("custom_test") 
{
    circular_buffer::circular_buffer<value_type> buffer(1);

    REQUIRE_NOTHROW(buffer.write(1));
    buffer.overwrite(2);
    REQUIRE_THROWS_AS(buffer.write(2), std::domain_error);
}
```

Detecting if it's full is tricky just using the head and tail pointers. See https://embeddedartistry.com/blog/2017/4/6/circular-buffers-in-cc which has an explanation.

marvelou-s: @patricksjackson thank you for pointing me the obvious ;)
`tail_ - head_ >= buffer_.size()`will solve it.

I have a question. Potentially, students may implement `overwrite()` method through `write()` which throws an exception, but `overwrite()` mustn't follow this logic. Should I modify test cases to prevent such scenario?

patricksjackson: I don't think just that will solve it. The problem (or one problem) is that `overwrite()` increments `head_` by constraining the value to valid indexes, but `head_` and `tail_` are incremented in other places without doing that. It really is not obvious, it's somewhat tricky to get right.


On `overwrite()` exceptions:

Writing `overwrite()` in terms of `write()` isn't necessarily incorrect. If you wanted `overwrite()` to be `noexcept` you can still make that guarantee by correctly handling the exception case. And it's not clear to me that `overwrite()` should even be `noexcept`, since it should probably throw if the buffer is size 0.

So no, I don't think it's necessary (though maybe I'm misunderstanding?). I am curious how you actually would test that though, because I can't think of a way to do it.

patricksjackson:
This looks good, thanks!

PR #289, opened 2019-06-04
Update anagram exercise test cases

### LINKED ISSUE #218: Anagram: Tests need an update
I find that the second and the last test cases of Anagram exercise are nearly the same, only a minor difference about the order of string. Is it a duplication or are there any reasons to include both tests?
Also, the tests here are quite behind the problem specifications repo.

Here are the 2 test cases I mentioned:

```
BOOST_AUTO_TEST_CASE(detects_simple_anagram)
{
    auto subject = anagram::anagram("ant");
    auto matches = subject.matches({"tan", "stand", "at"});
    vector<string> expected{"tan"};

    BOOST_REQUIRE_EQUAL_COLLECTIONS(expected.begin(), expected.end(), matches.begin(), matches.end());
}

BOOST_AUTO_TEST_CASE(matches_accepts_string_arguments)
{
    auto subject = anagram::anagram("ant");
    auto matches = subject.matches({"stand", "tan", "at"});
    vector<string> expected{"tan"};

    BOOST_REQUIRE_EQUAL_COLLECTIONS(expected.begin(), expected.end(), matches.begin(), matches.end());
}
```

marvelou-s:
This closes #218.

### DIFF
--- a/exercises/anagram/anagram_test.cpp
+++ b/exercises/anagram/anagram_test.cpp
@@ -3,6 +3,8 @@
 
 using namespace std;
 
+// Anagram exercise test case data version 1.4.0
+
 TEST_CASE("no_matches")
 {
     // 'anagram::anagram' defines a class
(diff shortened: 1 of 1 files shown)


patricksjackson:
Thanks!

PR #295, opened 2019-07-02
Finish update to Cpp14

KevinWMatthews:
We missed an exercise that was added after the initial upgrade.


KevinWMatthews:
@patricksjackson Finishes #292

patricksjackson:
My bad, I even remembered to ping on the new problem but forgot about this one. Thanks!

PR #296, opened 2019-07-09
Circular buffer - add test case for initial clear

marko213:
This adds a test case to Circular buffer, testing for a quite specific but important case when `circular_buffer::clear()` partially affects the use of the buffer.

In the solution where I came across this, the underlying `vector` was being `clear()`ed as well, preserving the allocated space while messing with the start and end iterators. This passed all the tests, however, as there was no test for `clear()` followed by `overwrite()` nor a failing `read()`.

[Link to the solution](https://exercism.io/mentor/solutions/dabad4b50b764378afb4e2cf976e584d?iteration_idx=2)

### DIFF
--- a/exercises/circular-buffer/circular_buffer_test.cpp
+++ b/exercises/circular-buffer/circular_buffer_test.cpp
@@ -2,7 +2,7 @@
 #include "circular_buffer.h"
 #include <stdexcept>
 
-// Circular-buffer exercise test case data version 1.1.0
+// Circular-buffer exercise test case data version 1.2.0
 
 TEST_CASE("reading_empty_buffer_should_fail") 
 {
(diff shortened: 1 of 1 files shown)


patricksjackson:
Good catch, I don't think I would've seen that issue in that solution.

I'd normally like to push changes to the test suites to the central list where they make sense - https://github.com/exercism/problem-specifications/blob/master/exercises/circular-buffer/canonical-data.json. I think this may be worth trying to add over there, then we could add this here and bump the test suite version :). They may have suggestions for changes so it'd be better to do it in that order.

This issue is actually pretty specific to C++ though. Note that with a small change the students solution will pass your test as well. I don't think you can test for this, since it appears that if you only use iterators for accesses it will just continue to work no matter how many times you call `clear()`, and neither ASAN or UBSAN will even warn against it.

I'm honestly a little more scared of C++ after this, because I assumed UBSAN would catch that...A note about this should be added to the mentor notes here: https://github.com/exercism/website-copy/tree/master/tracks

[1] - Add another iterator `end` as a member variable. Initialize it in the constructor as `buffer.end()`. Use it instead of `buffer.end()` inside of `increment()`

marko213:
What kind of a small change could pass the test? If you mean [1], then... well, that is technically valid (as in it shouldn't run into unallocated memory).

Quote from [here](https://en.cppreference.com/w/cpp/container/vector/clear):

> Leaves the capacity() of the vector unchanged

But then again, just above:

> Invalidates any references, pointers, or iterators referring to contained elements. Any past-the-end iterators are also invalidated. 

which seems to still keep the iterators valid for most cases.

***

I'll try adding it over there and maybe in the mentor notes as well. Should I close this PR in the meantime?

marko213:
This should probably not be merged until exercism/problem-specifications#1550 is merged.

patricksjackson:
Ah, yes I meant to add a reference to [1] in my writing, but forgot that. That is the small change I was referring to.

The iterators are all invalidated after a `clear()`, but that's only in writing and the library code doesn't check. The sanitizers probably have a hard time checking for that case.

Yes, let's wait to see if it goes into the problem-specificiations repo, then we can update here (or just add this as a custom test for us anyways, this is probably worth it for us at least since it restricts the potential issues).

marko213:
Also WIP as the changes made there must be incorporated into this PR.

patricksjackson:
Sorry, I don't think I got a notification for this when you updated(?). This looks great.

Thanks!

PR #299, opened 2019-07-30
Word-count - update tests specification to version 1.4.0

marko213:
Adds three test cases:

* substrings_from_the_beginning (exercism/problem-specifications#1557)
* multiple_spaces_not_detected_as_a_word (exercism/problem-specifications#1023)
* alternating_word_separators_not_detected_as_a_word (exercism/problem-specifications#1446)

### DIFF
--- a/exercises/word-count/word_count_test.cpp
+++ b/exercises/word-count/word_count_test.cpp
@@ -4,6 +4,8 @@
 
 using namespace std;
 
+// Word-count exercise test case data version 1.4.0
+
 TEST_CASE("counts_one_word")
 {
     const map<string, int> expected{{"word", 1}};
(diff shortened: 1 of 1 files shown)


patricksjackson:
So, it looks like the other tests are rather out of date as well...Can you update all of the tests to match the latest problem specifications and then add a note at the top with the new version?

This is fine, I just prefer to fully update to the latest test suite version when we do these :)

Example: https://github.com/exercism/cpp/blob/70eaf41f3320f13a15a361347b2619eb79d0b131/exercises/collatz-conjecture/collatz_conjecture_test.cpp#L5

marko213:
Ok, I'll close this PR, then create another one once I've worked through all the exercises.

patricksjackson:
Wait, I just meant the tests for word-count that are in this same file, not for all the exercises. I meant only updating the word-count tests to v1.4.0.

If you just mean you don't have time to do that, I will still take these changes.

marko213:
Ok, I'll do that.

Is there a need for *all* the exercises to be updated, as I misunderstood? I think I could give that a try too.

patricksjackson:
For this change, this is exactly what I was looking for, thanks!

Right now most of the test suites are out of date to the latest on problem-specifications, but there isn't really a good way to tell how out of date they are unless we go through all the individual tests. We've started adding the note about what test version it matches to the top of the test file so that there's an easier way to check that quickly.

Whenever a test suite is modified or created I have been pushing for the test suite to always do the full update to the latest version and add the version line. I haven't really pushed for an effort to update all of them, I've just been making sure that whenever we do have an update that it gets updated fully.

https://github.com/exercism/exercism/issues/4390
If the tool mentioned in this linked thread is ever built, this will be a lot easier to keep fully up to date. I've been holding off on pushing for a full update of the test suites since there's no way to keep them in sync yet.

(I should add this to the README I think, but I don't have the time this second)

PR #308, opened 2019-09-08
Use Github Actions for CI

patricksjackson:
Replaces Travis CI with Github Actions. This change also adds testing on Windows and Mac in addition to Linux.

I could leave in Travis, but this change is already a super-set of Travis functionality.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,98 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - master
+  pull_request:
+    branches:
+    - master
+
+jobs:
+  configuration-test:
+    name: Check Configuration
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Configlet
+      run: |
+        bin/fetch-configlet
+        bin/configlet lint .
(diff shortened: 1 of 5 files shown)


KevinWMatthews:
This is interesting.
I imagine I could use this to test compiler warnings on Windows?

patricksjackson:
See this as an example of this actually running: https://github.com/patricksjackson/xcpp/pull/1

Yes, this runs all the tests on Windows, Mac and Linux, and reports all the output, so this will let us test on all 3 OS's. There are some warnings in the Windows run actually  - https://github.com/patricksjackson/xcpp/pull/1/checks?check_run_id=218201746

I had to fix armstrong-numbers due to a compile error on Mac.

I think the actions may not be enabled until it's merged into the actual repo, so it won't actually run. https://github.com/exercism/julia has already added some and they appear to be running, so these should work once they're merged.

@KevinWMatthews If you can't see the checks that ran in my repo then we will need to wait to remove Travis until Github Actions comes out of Beta at least.

patricksjackson:
I'm merging this without any changes to the existing infrastructure. We can see if having it in the repo enables it or if we need to wait.

PR #317, opened 2019-10-06
Implementing exercise secret-handshake

mrfelipenoronha:


mrfelipenoronha:
So... I don't know the cause of the two errors, can someone help me?

NobbZ:
Travis as well as GitHub Actions tell you the very same thing. You haven't properly formatted the tracks config file.

They even tell you how to solve your problem:

> ```plain
> configlet fmt: changes needed
> Run 'bin/fetch-configlet && bin/configlet fmt .' to apply these changes.
> ```

KevinWMatthews:
Hi @mrfelipenoronha , thank you for your contribution! We appreciate you taking the time and effort to write and submit this.

Patrick and/or I will review this before we merge it into the master branch. You should hear from one of us soon!

KevinWMatthews:
The content of the tests looks solid and your solution is straightforward and simple. The exercise works well as it is - thanks!

I do have two comments that might help us polish it up nicely.

First some context - we try to copy the [problem-specifications](https://github.com/exercism/problem-specifications) repo as closely as is reasonable. The hope is that if we improve the exercise specs in a single location, then all tracks can benefit from the group's knowledge and consensus. It's a challenge to stay up-to-date, but we try!

With that in mind, one or two of our tests are out of order compared to the problem specification (they put the empty test last?). While I think I personally prefer your new order, they might have chosen the order for a reason.

Second, the `description` and `property` fields of our new exercise don't seem to match the spec.

For example, the [description](https://github.com/exercism/problem-specifications/blob/dab442f119d7a7196e27840cf8b5c7dc5c6e57fa/exercises/secret-handshake/canonical-data.json#L22) field for the first test case is `"wink for 1"` so we would write the test name as:

```cpp
TEST_CASE("wink for 1")
```

A side note - most of our existing tests use underscores instead of spaces: `wink_for_1`. Our new test library, `Catch2`, allows spaces in test names. You can use either convention, but personally I rather like using spaces - easier to read!

The [`property`](https://github.com/exercism/problem-specifications/blob/dab442f119d7a7196e27840cf8b5c7dc5c6e57fa/exercises/secret-handshake/canonical-data.json#L23) field, in this case `commands`, is typically the function name:

```cpp
secret_handshake::commands(1)
```

That would give a complete first test as:

```cpp
TEST_CASE("wink for 1")
{
    REQUIRE(vector<string>{"wink"} == secret_handshake::commands(1));
}
```

Neither of these suggestions have an impact on the logic, syntax, or style of the tests or of the solution but they do help us keep in sync with the spec.

Of course, the specs aren't perfect. If you would like to suggest a change I will encourage you to raise an issue in the problem specifications repo. They frequently make updates so there's a good chance that your improvements could help all the language tracks.

Thoughts? Are you are interested in making a revision?

Thanks again for your work!

mrfelipenoronha:
Hi,

Thank you for all of yours suggestions. I will review the code and make sure that it attends to the specifications, I think that all the *mistakes* are *first time here* mistakes haha.

To write this exercise I used the implementation of `collatz-conjecture` as base. Hence I'm not very confident on using the library `Catch2`. 

In this scenario I'm going to review all the pointed topics, except the part of using space instead of underline in the `description`, is that ok?!

KevinWMatthews:
Hi @mrfelipenoronha ,

Nice! Those changes are exactly what I was hoping for.

Yes, using underscores instead of spaces in test names is fine.

Don't worry too much about making mistakes. If you try to do good work and are open to feedback, then it's ok. Sometimes we correct things, other times a mistake might actually be a good idea.

I often feel like the hardest part of open-source is giving and receiving feedback. You have been very easy to work with. Thanks!

KevinWMatthews:
@patricksjackson I'm happy with this exercise looks and it seems straightforward, so I'll go ahead and merge it into master. If you see anything out of sorts, just let us know.

@mrfelipenoronha ok, this should be live!

PR #319, opened 2019-10-07
Implementing exercise protein-translation

mrfelipenoronha:


mrfelipenoronha:
Sorry for the mess in the history of this PR, I kind of mixed up stuff from the previous one. How can I avoid commits from one PR mixing with commits from another PR?

Anyway, here is another implementation, with everything according to the specs.

KevinWMatthews:
Hi @mrfelipenoronha , thanks for making another contribution!

I see what you mean - the commits from your previous PR are lingering here. Was your local master branch up-to-date before you submitted this PR? I squashed the commits from the other PR before I merged them, which put your master branch out of date.

The easiest way to solve this might be with a rebase and a force push. You're welcome to do that, but if you haven't done that before don't worry about it too much - we'll be sure to set things straight before we merge. In the future, it might be easier to submit PRs by creating a new branch first; a squash might not be as confusing in that context.

Am I explaining things at the right level of detail? Maybe I can ask how much you've been exposed to git - that will help us figure out what the best way forward.

Thanks again for a new exercise! Patrick or I will review it as soon as we can.

KevinWMatthews:
Ah, I see - that often a good strategy, but in this specific case it makes the history more complicated.

If you're relatively new to git, it's best to avoid a `force push` unless you are *sure* of what you are doing - this can permanently delete history!

If you haven't seen it already, the [Git Book](https://git-scm.com/book/en/v2) is an excellent resource. It's a long read but it is rewarding. The section on branching relates to what we're doing here.

patricksjackson:
https://github.com/exercism/docs/blob/master/contributing/git-basics.md#branches

That guide helped lay out the flow for me when I was figuring things out. The main thing that will help is if you only make changes in a branch that is not master for each separate change. Squashing commits isn't necessary here either, since we will (usually) squash all of your changes in a PR when we merge it.

Also, before we've started reviewing it it is perfectly ok to just close a PR and open a new one if that helps clean up some git issue.

(I usually don't get time to review things till the weekend, so no review just yet).

config.json
```diff
@@ -589,6 +589,18 @@
         "arrays",
         "bitwise_operations"
       ]
+    },
+    {
+      "slug": "protein-translation",
+      "uuid": "fb73e822-4831-4d33-a049-44a514bfa46a",
+      "core": false,
+      "unlocked_by": null,
```
patricksjackson: Unlocked by reverse-string maybe?

Several of the exercises are unlocked by nothing, but we're trying not to do that anymore.

exercises/protein-translation/example.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <vector>
+#include <string>
+#include <algorithm>
+#include <map>
+
+using namespace std;
+
+vector<string> protein_translation::proteins(string rna) {
+
+    map<string, string> dic;
```
patricksjackson: Nit: Should probably use an `unordered_map` here. It's more explicit about the fact that you don't need them to be in order.

Even better would be a `static const unordered_map`.

exercises/protein-translation/example.cpp
```diff
@@ -0,0 +1,29 @@
...
+    dic["UUU"] = dic["UUC"] = "Phenylalanine";
+    dic["UUA"] = dic["UUG"] = "Leucine";
+    dic["UCU"] = dic["UCC"] = dic["UCA"] = dic["UCG"] = "Serine";
+    dic["UAU"] = dic["UAC"] = "Tyrosine";
+    dic["UGU"] = dic["UGC"] = "Cysteine";
+    dic["UGG"] = "Tryptophan";
+    dic["UAA"] = dic["UAG"] = dic["UGA"] = "STOP";
+
+    vector<string> polypeptide;
+    for (int i = 0; i < (int)rna.size()-2; i += 3) {
```
patricksjackson: Nit: Should probably use `size_t` instead of `int` here for `i`'s type.

exercises/protein-translation/example.cpp
```diff
@@ -0,0 +1,29 @@
...
+        string codon = rna.substr(i, 3);
+        string protein = dic[codon];
```
patricksjackson: The tests for some reason don't check, but this will cause problems if the value in the map doesn't exist. You want to use `dic.find(codon)` instead.

exercises/protein-translation/example.cpp
mrfelipenoronha: I didn't make this change because I'm afraid we can get some king of overflow error

exercises/protein-translation/example.cpp
mrfelipenoronha: We will only get an empty string "", aren't we?

exercises/protein-translation/example.cpp
patricksjackson: Oops, yes you're correct here :) There'd be a pretty bad underflow error if the array was small.

exercises/protein-translation/example.cpp
patricksjackson: With this solution the empty string would be included in the output, which is probably not the desired behavior.. For some reason the tests don't check the error case where the string doesn't exist, I'd assume we'd treat it as either a "STOP" or as an error.

Also to be clear, this line will also add an entry to the map with key = `codon` and the value = the empty string. It's a quirk of c++ maps that you need to be careful with.

(I will probably just fix this myself in a few minutes and push the change, everything else looks good)

patricksjackson:
I pinged @iHiD on Slack to run the update script, since that is still a manual process.

iHiD:
Run :)

PR #328, opened 2019-12-22
complex-numbers: Fixing operator== and adding skeleton files.

### LINKED ISSUE #326: complex-number: Don't require an operator==
The exercise "complex-numbers" requires the students to implement an `operator==` that returns `true` iff the two complex numbers are *roughly* the same.

That's what the [example solution](https://github.com/exercism/cpp/blob/master/exercises/complex-numbers/example.cpp#L49-L65) does, and I see no other way to pass the [tests](https://github.com/exercism/cpp/blob/master/exercises/complex-numbers/complex_numbers_test.cpp#L199-L222) of the `.exp()` method.

I don't think that's the best approach. There's a reason why in C++ `some_double1 == some_double2` compares the exact values.

Also, the [instructions](https://github.com/exercism/cpp/blob/master/exercises/complex-numbers/README.md) don't mention the `operator==()`.

I looked what tests other language tracks do:

- [Python](https://github.com/exercism/python/blob/master/exercises/complex-numbers/complex_numbers_test.py) uses exact comparisons for most of the tests, uses an absolute delta of `1e-7` for `.exp()`.
- [JavaScript](https://github.com/exercism/javascript/blob/master/exercises/complex-numbers/complex-numbers.spec.js) uses exact comparisons for most of the tests, uses an absolute delta of `1e-2` for `.exp`.
- [Swift](https://github.com/exercism/swift/blob/master/exercises/complex-numbers/Tests/ComplexNumbersTests/ComplexNumbersTests.swift) uses exact comparisons for most of the tests, uses an absolute delta absolute delta of `1e-3` for `.exponent()`.
- [Java](https://github.com/exercism/java/blob/master/exercises/complex-numbers/src/test/java/ComplexNumberTest.java) and [Kotlin](https://github.com/exercism/kotlin/blob/master/exercises/complex-numbers/src/test/kotlin/ComplexNumberTest.kt) use an absolute delta of `1e-15` for all tests.
- [C#](https://github.com/exercism/csharp/blob/master/exercises/complex-numbers/ComplexNumbersTest.cs) uses exact comparisons for some tests, but many tests use an absolute delta of `1e-7`.
- [Haskell](https://github.com/exercism/haskell/blob/master/exercises/complex-numbers/test/Tests.hs) and [F#](https://github.com/exercism/fsharp/blob/master/exercises/complex-numbers/ComplexNumbersTest.fs) use exact comparisons for the getters and `abs`, the other tests AFAIK use an absolute delta of `1e-2`.
- [Scala](https://github.com/exercism/scala/blob/master/exercises/complex-numbers/src/test/scala/ComplexNumberTest.scala) uses some exact comparisons (?), but most tests use an absolute(?) delta of `1e-15`.
- [Erlang](https://github.com/exercism/erlang/blob/master/exercises/complex-numbers/test/complex_numbers_tests.erl) requires the students to implement `equal`.
- I don't understand what [Clojure](https://github.com/exercism/clojure/blob/master/exercises/complex-numbers/test/complex_numbers_test.clj) does.

So the only other language track that expects students to implement some "rough comparison" is Erlang. In all the other tracks the tests are responsible for exact or inexact comparisons of floating point values.

I'd suggest that we follow the majority of the language tracks, and do the "rough comparisons" in (some of) the tests. Catch2 has some built-in tools for that (see [assertions.md](https://github.com/catchorg/Catch2/blob/master/docs/assertions.md#floating-point-comparisons) and [matchers.md](https://github.com/catchorg/Catch2/blob/master/docs/matchers.md#floating-point-matchers)). We could even implement a custom matcher (see [matchers.md](https://github.com/catchorg/Catch2/blob/master/docs/matchers.md#custom-matchers)) but I think that would confuse some of our students.

patricksjackson:
Closes #326


patricksjackson:
@siebenschlaefer @heitorPB - let me know if this looks good :)

heitorPB:
In line 21:
`static const double m = 0.005;`

In Phsyics/Maths, I've always seen this as `eps` (or similar), not `m`.  It's a reference to [(ε, δ)-definition of limit](https://en.wikipedia.org/wiki/(%CE%B5,_%CE%B4)-definition_of_limit). Is there a reason to use `m` instead?

patricksjackson:
It was named `m` because I was lazy and didn't think about it :). Renamed it to eps and modified the comments to clarify.

siebenschlaefer:
@patricksjackson Looks good!

patricksjackson:
Thanks for taking a look!

PR #333, opened 2020-01-10
fix inconsistency in exercise topics

### LINKED ISSUE #329: Inconsistency in exercise topics
There are interesting topic names here: `algorithms` and  `algorithm`.
There's also `Control Flow If Else Statements` and `Control Flow Conditionals`.
And also `Control Flow loops` and `Loops`.

 In [problem-specifications/TOPICS.txt](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) there are `algorithms`, `loops`, `conditionals`. 

Is there a reason for this? I'd like to submit a PR to fix it, if it was not intentional.

heitorPB:
Are there other topics that need changes?

fix #329

### DIFF
--- a/config.json
+++ b/config.json
@@ -25,8 +25,8 @@
       "unlocked_by": null,
       "difficulty": 3,
       "topics": [
-        "control_flow_if_else_statements",
-        "control_flow_loops",
+        "conditionals",
+        "loops",
         "strings"
       ]
     },
@@ -37,7 +37,7 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "control_flow_conditionals",
+        "conditionals",
         "logic"
       ]
     },
@@ -69,7 +69,7 @@
       "unlocked_by": "hello-world",
       "difficulty": 3,
       "topics": [
-        "control_flow_loops",
+        "loops",
         "strings"
       ]
     },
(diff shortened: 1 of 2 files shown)


config.json
```diff
@@ -26,7 +26,7 @@
       "difficulty": 3,
       "topics": [
         "control_flow_if_else_statements",
```
patricksjackson: I think this can also become "conditionals"

patricksjackson:
Other topic fixes I noticed:
"mathematics" should be "math"
"data_structure" should be "data_structures" (plural)

You need to run `bin/fetch-configlet && bin/configlet fmt .` to reformat the config.json file before we can merge it.

Also if you have the time, adding a brief pointer to the topics file in the README under 'Adding an Exercise' would be very useful :)

Thanks!

heitorPB:
Oh, i missed those two. Thanks for spotting them!

Oh, didn't know configlet is needed to update the topics as well. Travis is happy now :)

I updated the README as suggested. What do you think about it?

patricksjackson:
Perfect! The configlet tool is used for several things, including (in this case) just making sure that the formatting is correct in the json file. It was just making sure that the list of Topics was in alphabetical order.

Thanks!

PR #345, opened 2020-03-06
Modify fetch-configlet script

marvelou-s:
https://github.com/exercism/configlet/issues/173: Travis CI are fails because of fetch-configlet script failure while getting latest release URL. As described in https://github.com/exercism/haskell/pull/898#pullrequestreview-368719612 the problem was that  `curl` started to output headers in lowercase

Few days ago I was created PR which simply fixed the case (https://github.com/exercism/cpp/pull/344)

This PR provides possibility to get latest release of configlet script using GitHub API.

### DIFF
--- a/bin/fetch-configlet
+++ b/bin/fetch-configlet
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-LATEST=https://github.com/exercism/configlet/releases/latest
-
 OS=$(
 case $(uname) in
     (Darwin*)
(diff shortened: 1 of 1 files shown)


patricksjackson:
Should this file just be tracking a central fetch-configlet script that can be shared by all the tracks? Maybe https://github.com/exercism/configlet/blob/master/scripts/fetch-configlet?

It appears that the one in the configlet repo does a few things differently from this but it doesn't have the fix from https://github.com/exercism/configlet/issues/173 yet.

marvelou-s:
Fully agree. Script should be centralized, but it isn't. Each track maintaining it by itself.
The idea I came with to stop script being depended on case-sensitivity `curl` output and use standard GitHub API. Perhaps it makes sense to propose it to exercism/configlet#173.

patricksjackson:
Yea, I think the better solution here is to try to get the version in the configlet repo fixed and merge all the updates there into here.

This is a good change so I'll go ahead and merge, but the long term solution should be trying to herd everybody into sharing a single version. (You don't necessarily have to be the one to do that, but I'm just stating it here for future reference).

Thanks!

PR #347, opened 2020-04-25
Add test for double and integer argumetns

elyashiv:
Some solutions use templates, and this test won't compile if they have just one template argument.

A sample compilation error:
no matching function for call to ‘kind(double, int, int)’

### DIFF
--- a/exercises/triangle/triangle_test.cpp
+++ b/exercises/triangle/triangle_test.cpp
@@ -72,4 +72,9 @@ TEST_CASE("larger_triangles_violating_triangle_inequality_are_illegal")
 {
     REQUIRE_THROWS_AS(triangle::kind(7, 3, 2), std::domain_error);
 }
+
+TEST_CASE("double_and_integer_arguments")
+{
+    REQUIRE(triangle::flavor::scalene == triangle::kind(5.5, 4, 2));
+}
 #endif


elyashiv:
It seems like the windows test can't pass. I think this is because boost is not installed. I will like if someone can confirm this is unrelated to my change and open an issue.

patricksjackson:
I think that you should also re-write the example using templates to demonstrate.


The CI issue is because of https://github.com/actions/virtual-environments/issues/319, not because of the changes here of course. They removed BOOST_ROOT which Cmake requires to find boost on Windows, and is normally set by a Boost install on Windows.

I will either fix this weekend or just disable the Windows build because I have wasted days of my life dealing with just Boost in this repo and I can't be bothered to care anymore.

patricksjackson:
I've pushed the change to fix the CI issue. Rebasing the changes on top of master should fix it here.

elyashiv:
> I think that you should also re-write the example using templates to demonstrate.

I'm not sure what you mean. You want me to fail the test? Or add an example with a few template arguments?

Can I add more than one example to the CI?

siebenschlaefer:
> > I think that you should also re-write the example using templates to demonstrate.
> I'm not sure what you mean.

There are example implementations for each exercise in the files `example.h` and `example.cpp`.

elyashiv:
@siebenschlaefer  
I knew about the example, but there is already an example to pass the tests.

Also, I added a test that should not pass in the simple template solution.

siebenschlaefer:
@elyashiv Sorry, I misunderstood.

elyashiv:
What else is needed for this to be merged?

elyashiv:
Hi, is this ready to be merged?

PR #348, opened 2020-05-01
Fix Boost issues in Github Actions

patricksjackson:
Removes/Adds workarounds for Boost running on Github Actions.


patricksjackson:
I'll go ahead and merge this myself, as all this change does is remove the two Boost exercises from being tested on Windows due to the regression caught in #347.

Apologies for the noise.

Issue #274, opened 2019-04-20
Implement and Enforce Code Formatting


patricksjackson:
We should standardize on a code format convention, and enforce it throughout this repo. We won't force the students to use it, but we will format our own files in this way so that the test suites are in a consistent style for the students.

[Clang Format](https://clang.llvm.org/docs/ClangFormat.html) is the standard way of formatting cpp files, so this would entail adding a `.clang-format` file to the root with the settings we want, and then either removing or modifying the .editorconfig file to represent our convention. We'd also need to add a check to CI, and preferably just format the whole repo in that style. We could do it incrementally to not mess up git blame, as many projects do, but that's not really a useful history to preserve in this repo.

First we need to agree on a format to use though. I have always felt that the details don't matter too much, but it's important to have a consistent style everywhere. Conversations about the right formatting often devolve into [bikeshedding](https://en.wikipedia.org/wiki/Law_of_triviality), which I find exhausting and unnecessary.

So I propose we use the following clang-format settings, which are reasonably close to what most of the repo is already in. If anyone has an objection I would want to see a strong reason why.

```
BasedOnStyle: Google
IndentWidth: 4
```

KevinWMatthews:
No strong opinions on formatting style, but Clang Format seems like a good tool to use.

patricksjackson:
I tried this out in one of my branches - https://github.com/patricksjackson/xcpp/tree/fmt

The problem is that clang-format is not stable. The Styles themselves change over time, and even dumping the settings to a file doesn't work over time, presumably because non-specified settings get different defaults with new clang versions. Version pinning isn't great since we still need developers to install specific versions locally.

I think what needs to happen is that rather than attempting to format the whole repo on every change and forcing that config, we do a one-time format of everything and then setup git-clang-format for each PR, but let us override it if necessary.

PR #352, opened 2020-06-03
Implemented enforcement of code formatting style

silvanocerza:
This should solve #274.
I've not formatted any code in this PR just implemented a way to enforce it, I was thinking of creating a separate PR after this one is merged with a single commit that formats all C++ code.

### DIFF
--- a/.github/workflows/format-code.yml
+++ b/.github/workflows/format-code.yml
@@ -1,0 +1,49 @@
+name: "Format C++ code"
+
+on:
+  issue_comment:
+    types: [created]
+
+jobs:
+  format:
+    name: "Format C++ code"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/format-code.yml
```diff
@@ -0,0 +1,49 @@
+name: "Format C++ code"
```
patricksjackson, line 1: What does this workflow do? I _think_ it lets us comment using either of those two format strings on a pull request and have the CI make a formatting commit, is that right?

silvanocerza: Yes, it does exactly that. Something similar is already in use for the C# track and for Markdown and JSON files in the V3 repo.

For example in PR exercism/v3#1788 you can see the one for Markdown and Json in action.

silvanocerza:
No worries about the delay. :)

By the way as soon as I can I'll make another PR formatting the whole repository.

patricksjackson:
Awesome thanks :)

PR #354, opened 2020-07-06
Added debug symbols to executable

elyashiv:


elyashiv:
Inspired by https://exercism-team.slack.com/archives/CASG1FXUJ/p1594048624091900

patricksjackson:
There is probably a way to have CMake add this flag, but this will work for now.

PR #462, opened 2021-06-24
Fix bug in clock_test.cpp

gerymate:
Fix equality check in test - the previous version was passing when both == and != was implemented to always return true.

### DIFF
--- a/exercises/practice/clock/clock_test.cpp
+++ b/exercises/practice/clock/clock_test.cpp
@@ -206,11 +206,11 @@ TEST_CASE("equal_tests")
 
         if (e.expected) {
             INFO(errorMsg(string(clock1), string(clock2), e.msg));
-            REQUIRE(clock1 == clock2);
+            REQUIRE(!(clock1 != clock2));
         }
         else {
             INFO("[" << string(clock1) << " == " << string(clock2) << "] test case: " << e.msg);
-            REQUIRE(clock1 != clock2);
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
With those changes the `operator==` and `operator!=` could just always return `false` to pass the tests. 

If you want to catch those incorrect implementations you could change the tests to something like
```cpp
INFO(errorMsg(string(clock1), string(clock2), e.msg));
REQUIRE(e.expected == (clock1 == clock2));

INFO("[" << string(clock1) << " == " << string(clock2) << "] test case: " << e.msg);
REQUIRE(!e.expected == (clock1 != clock2));
```

But in my *personal* opinion the current tests are not incorrect. They are not meant to be exhaustive. Without randomized tests and assignments/casts to known types it's always possible to write code that passes the tests without solving the problem. E.g. in the leap and armstrong-number exercises it's possible to compare the parameter with just a few fixed numbers. IMHO the tests should guide you and test a reasonable sample of input data, not be bullet-proof.

patricksjackson:
I agree that the test cases won't ever be exhaustive, but this is a pretty basic issue with the test suite itself that I think we should probably fix.

I agree with @siebenschlaefer that the current change has the same problem though. The suggested changes are better, though they should be placed into [separate scopes for the logging statements to work correctly](https://github.com/catchorg/Catch2/blob/devel/docs/logging.md) and we should add some documentation in there since this is visible to students.

PR #466, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/.meta/config.json
+++ b/exercises/practice/hello-world/.meta/config.json
@@ -12,8 +12,7 @@
   ],
   "files": {
     "solution": [
-      "hello_world.cpp",
-      "hello_world.h"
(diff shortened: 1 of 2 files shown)


exercises/practice/hello-world/.meta/config.json
```diff
@@ -12,8 +12,7 @@
   ],
   "files": {
     "solution": [
-      "hello_world.cpp",
-      "hello_world.h"
+      "hello_world.cpp"
```
ErikSchierboom, line 15: I've removed the header file as the student could in this case just as easily work with just the source file. It makes things a bit simpler.

PR #484, opened 2021-09-09
Include hello_world.h as part of the solution

### LINKED ISSUE #482: 'Hello World' exercise is missing header file when downloaded via the CLI
When downloading via the CLI, the project is missing hello_world.h but CMakeLists.txt is still looking for it.

ref: https://github.com/exercism/cpp/blob/26352eb5dccf8553228add904ae6aca227e3b07f/exercises/practice/hello-world/CMakeLists.txt#L35

I think it could be due to changes to config.json in commit [7e75421](https://github.com/exercism/cpp/commit/7e75421ab1c195981401ea6a5666315d90a7b5dc#diff-5ab184438a2f2595277344b75c576e1471c27226ad6e0f030df90c0380526f01)

Completing the challenge via the online editor required a 1 one line change, but via the CLI I had to create a header file in order to build and run the tests.

patricksjackson:
Fixes #482 

The skeleton solution downloaded for the user needs to include hello_world.h, otherwise the student needs to create that file themselves.

As a sidenote, does anyone know if there is any documentation on what the fields in this json file actually do? I haven't been able to find any and am only assuming from context.

### DIFF
--- a/exercises/practice/hello-world/.meta/config.json
+++ b/exercises/practice/hello-world/.meta/config.json
@@ -12,7 +12,8 @@
   ],
   "files": {
     "solution": [
-      "hello_world.cpp"
+      "hello_world.cpp",
+      "hello_world.h"
     ],
     "test": [
       "hello_world_test.cpp"


patricksjackson:
I answered my own question. The documentation for the json files is at https://github.com/exercism/docs/blob/main/building/tracks/config-json.md

patricksjackson:
@exercism/cpp I confirmed that `hello-world` is broken on the cli without this change. If someone has the time please review, thanks.

vinothpandian (approved):
I encountered the same issue with the missing header file `hello_world.h`. The PR solves the problem 👍🏾

patricksjackson:
@ErikSchierboom @iHid Can you approve this change? It's been broken since you merged #466 back in July. The header file is required to compile the tests, so as a result the students need to figure out how to make their own header file just to finish `hello world`

I can't merge this change myself as no one else is active to review my PRs.

ErikSchierboom (approved):

iHiD:
So I'm clear, what's the actual issue that a student has? The `.h` should still be downloaded regardless of whether it's in this field or not, as I understand things. It missing from the `solution` field will cause the issue that it's not editable in the editor, but that doesn't seem to be the issue reported here. So could you give me a little more info in case there's a wider issue pls.

@patricksjackson Also, just as an FYI, you can tag `@exercism/reviewers` into any PR that needs reviewing in any repo where you're the only active maintainer, and they will have permissions to approve/merge.

patricksjackson:
Thanks :)

@iHiD The `.h` file is not being downloaded which the students are reporting and I confirmed as well.

Following [the docs here](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md#file-stub-implementation), I _think_ the issue is that this field needs to point to the exercise stub files for them to be downloaded.

However it doesn't seem to be fixed ~30 minutes after submitting. I'm not sure if there is any caching going on, so I'll double check tomorrow to see if it's fixed.

iHiD:
Hmm, I think the issue here is actually that the header file is listed in the example files in the config, which are excluded from download. A file should not be included in both so needs to be removed from one or the other. Do we know why it was added to the example files? Can it be removed from there?

Config LOC: https://github.com/patricksjackson/xcpp/blob/0a91e6a82e27b42a6ea75c9494188b10bff693a5/exercises/practice/hello-world/.meta/config.json#L23

ErikSchierboom:
Yep, that must be it. It needs to be removed from the example files.

patricksjackson:
Ah, that would explain it then :)

I've opened #505 just to make that change then. Thanks!

PR #485, opened 2021-09-11
Remove the String concept exercise.

patricksjackson:
This concept exercise only consists of stub data and an author file, it is not worthwhile to keep this data around.


patricksjackson:
Ah, I missed the implementation is separate in exercises/concepts, so there is at least something there. I will take a more holistic look to understand what's going on with the current version of Concepts before removing, since requirements have changed about 3 or 4 times since this was made.

PR #486, opened 2021-09-11
Modified CMakeLists to add flags and fixed code flagging warnings.

### LINKED ISSUE #302: Warninig Levels for Visual Studio's C++ compiler
Copying an issue raised on [exercism/exercism](https://github.com/exercism/exercism/issues/4976) for better visibility:

> In the C++ track, exercises come with a CMakeLists.txt file that configures CMake for each exercise, and I noticed that when building C++ solutions to exercises certain warnings I'd expect to see aren't appearing.
>
> It looks like warning levels are set to be verbose for Clang and GNU compilers but there's no setting that addresses the default compiler MSVS ships with.
>
> https://stackoverflow.com/questions/2368811/how-toset-warning-level-in-cmake/50882216#50882216
>
> This SO answer describes how that could be addressed for visual studio, and it would be a great help to new programmers such as myself if this configuration option could be added.

CryptoToxin:
Hopefully fixes #302 

For all of the exercise CMakeLists.txt I've added ``/W4`` (Level 4 warnings) and ``/WX`` (Warnings as errors). I've also gone through and fixed code which currently fails due to this addition. I've tested these changes in both VS2019 and VS2022 toolsets.

Note: currently don't have a working gcc/clang setup so will await CI to check the changes.

As per the thread, I didn't add ``/Wall`` as I believe this is somewhat heavy-handed for what is the desired outcome. I was toying with suggesting ``/permissive-`` however I kept strictly to the main point of the issue.

Warning Levels: https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-160
WX Flag: https://docs.microsoft.com/en-us/cpp/build/reference/wx-treat-linker-warnings-as-errors?view=msvc-160

### DIFF
--- a/exercises/practice/atbash-cipher/.meta/example.cpp
+++ b/exercises/practice/atbash-cipher/.meta/example.cpp
@@ -10,7 +10,7 @@ const char ALPHA_END = 'z';
 
 char atbash_transform(char c)
 {
-    return static_cast<char>(std::isalpha(c)) ? ALPHA_END - static_cast<char>(std::tolower(c)) + ALPHA_START : c;
+    return static_cast<char>(std::isalpha(c) ? ALPHA_END - std::tolower(c) + ALPHA_START : c);
 }
 
 std::string encode(std::string const& plaintext)
--- a/exercises/practice/atbash-cipher/.meta/example.cpp
+++ b/exercises/practice/atbash-cipher/.meta/example.cpp
@@ -10,7 +10,7 @@ const char ALPHA_END = 'z';
 
 char atbash_transform(char c)
 {
-    return std::isalpha(c) ? ALPHA_END - std::tolower(c) + ALPHA_START : c;
+    return static_cast<char>(std::isalpha(c)) ? ALPHA_END - static_cast<char>(std::tolower(c)) + ALPHA_START : c;
 }
 
 std::string encode(std::string const& plaintext)
--- a/exercises/practice/acronym/.meta/example.cpp
+++ b/exercises/practice/acronym/.meta/example.cpp
@@ -15,7 +15,7 @@ string acronym(string const &input_str) {
             delimiter_state = FOUND;
         } else if (delimiter_state == FOUND) {
             delimiter_state = NOT_FOUND;
-            acronym_str.push_back(std::toupper(c));
+            acronym_str.push_back(static_cast<char>(std::toupper(c)));
         }
     }
     return acronym_str;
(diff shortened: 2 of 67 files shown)


patricksjackson:
Thanks!

The CI failure is not related to this change, and is because we are using an old version of Ubuntu in our CI that is temporarily disabled in Github actions. I will rerun the CI in a few hours and most likely merge it in then. I expect the CI will pass when it actually runs.

patricksjackson (approved):

exercises/practice/atbash-cipher/.meta/example.cpp
```diff
@@ -10,7 +10,7 @@ const char ALPHA_END = 'z';
 
 char atbash_transform(char c)
 {
-    return std::isalpha(c) ? ALPHA_END - std::tolower(c) + ALPHA_START : c;
+    return static_cast<char>(std::isalpha(c)) ? ALPHA_END - static_cast<char>(std::tolower(c)) + ALPHA_START : c;
```
patricksjackson, line 13: Atbash-cipher is failing on older versions. I think the first static cast is somehow causing it to always return false.

Maybe it needs to be `static_case<bool>(std::isalpha(static_cast<unsigned char>(c)))`?

See cppreference on isalpha: https://en.cppreference.com/w/cpp/string/byte/isalpha

patricksjackson (changes requested):

siebenschlaefer:
From a human perspective: Is it really necessary to write `static_cast<bool>(isalpha(...))` instead of `isalpha(...)`?
*Personally* I would hate that in my code because IMHO it just makes the code tedious to write and harder to read.

CryptoToxin:
> From a human perspective: Is it really necessary to write `static_cast<bool>(isalpha(...))` instead of `isalpha(...)`?
> _Personally_ I would hate that in my code because IMHO it just makes the code tedious to write and harder to read.

I somewhat agree however as per [cppref](https://en.cppreference.com/w/cpp/string/byte/tolower) it would appear that this should probably be further modified to include not only the return type being ``static_cast<char>`` but also the input ``static_cast<unsigned char>`` (similar for all cctype functions - cctype is included via the string header). The first example in the notes section is a pretty good example of the use.

siebenschlaefer:
> this should probably be further modified to include [...] input `static_cast<unsigned char>`

I agree completely, that's one of my pet peeves.

> cctype is included via the string header

I don't think that's required by the standard.

exercises/practice/atbash-cipher/.meta/example.cpp
CryptoToxin: To clarify - wrapped the line with a static_cast<char> in order to fix the issue.

CryptoToxin:
Just to quickly clarify - there's nothing outstanding on this PR from my point of view?

CryptoToxin:
Closing, this PR could well be out of date by now, I haven't checked.

Issue #491, opened 2021-09-25
Gigasecond compilation error on submission


CSjianbel:
I solved the problem locally and was able to pass all testcases, but during submission through cli my solution fails and I get the following compilation error from the tests tab: 

Results locally: 
```bash
[100%] Built target gigasecond
===============================================================================
All tests passed (5 assertions in 5 test cases)

[100%] Built target test_gigasecond
```

Error from exercism's iteration page on tests tab:
```bash
make[2]: *** No rule to make target '/usr/lib/libboost_date_time.a', needed by 'gigasecond'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:77: CMakeFiles/gigasecond.dir/all] Error 2
make: *** [Makefile:84: all] Error 2
```

AmjadHD:
I get that when testing.

mculaway:
pano mo to nasolve tol

PR #495, opened 2021-10-07
Fix compile error for exercises using boost:date_time lib

peterschussheim:
## Overview

This PR is following up on [cpp: gigasecond does not compile](https://github.com/exercism/cpp-test-runner/issues/37) and [a fix for cpp-test-runner](https://github.com/exercism/cpp-test-runner/pull/38) which updated the test runner dockerfile to add a missing `boost::date_time` lib used in "gigasecond" and "meetup" CPP exercises.

[PR 38](https://github.com/exercism/cpp-test-runner/pull/38) did the following:

- install boost in test runner docker container
- add minimal tests for the test runner ensuring the library can be used by the runner

Unfortunately, this didn't fix the problem completely because the existing two exercises that use `boost::date_time`  require the static versions of boost in their `CMakeLists.txt` files.

Also related is #491.

## Proposed solution

This pull request updates those two exercises by removing the static lib requirement and I expect this will resolve the problem.

### DIFF
--- a/exercises/practice/gigasecond/CMakeLists.txt
+++ b/exercises/practice/gigasecond/CMakeLists.txt
@@ -8,7 +8,6 @@ cmake_minimum_required(VERSION 3.5.1)
 project(${exercise} CXX)
 
 # Locate Boost date_time library
-set(Boost_USE_STATIC_LIBS ON)
 set(Boost_USE_MULTITHREADED ON)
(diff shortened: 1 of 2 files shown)


patricksjackson (approved):

patricksjackson:
Thanks! I think this will do the trick.

PR #502, opened 2021-11-13
Update tests.toml for acronym exercise

CyanCoding:
- Formatted the file to look like the example provided on #362 
- Referred to `acronym_test.cpp` and added missing tests
- Set tests to true/false based on whether they were used or not


patricksjackson:
Unfortunately, it appears that they did change this file to a new format, and the instructions you used were for the old one.

The up to date info is here: https://github.com/exercism/problem-specifications#track-test-data

PR #506, opened 2022-01-17
Fixing error in instructions.md examples (remove `and` from examples)

ntmartinez:
This PR fixes a typo in the examples in the instructions.

The instructions have examples that include the string `"and"` in the output for numbers like `120` (output stated in instructions: `"one hundred and twenty"`). However, the actual test cases in the code don't include the word `"and"` for numbers like `120`, and instead use output like `"one hundred twenty"` (for example, the test case for `123` in the code uses the output `"one hundred twenty-three"`; `"and"` is not present in the output).

This PR removes the `"and"` from the instructions' examples to make their syntax correct.

Let me know if I missed anything. Thanks :slightly_smiling_face:

### DIFF
--- a/exercises/practice/say/.docs/instructions.md
+++ b/exercises/practice/say/.docs/instructions.md
@@ -58,6 +58,6 @@ Use _and_ (correctly) when spelling out the number in English:
 
 - 14 becomes "fourteen".
 - 100 becomes "one hundred".
-- 120 becomes "one hundred and twenty".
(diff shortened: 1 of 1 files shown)


ntmartinez:
Fixes https://github.com/exercism/cpp/issues/507

patricksjackson:
I think I've now spotted the issue, the instructions that include adding `and` are an exercise Extension and not part of the core exercise. I've explained more in #507.

Thanks!

PR #508, opened 2022-01-22
Update binary_test.cpp

bobahop:
Change TEST_CASE("carrot_is_decimal_0") to pass in "carrot" instead of "convert".


patricksjackson (approved):
Nice find! Thanks!

PR #509, opened 2022-02-15
updated V3  #421

rakshitdabral:
gave prerequisites to practice while removing all the topics also added slag to concepts

### DIFF
--- a/config.json
+++ b/config.json
@@ -53,674 +53,441 @@
         "slug": "hello-world",
         "name": "Hello World",
         "uuid": "3eeadff6-82b6-43be-8834-d9ef258e454d",
-        "practices": [],
-        "prerequisites": [],
-        "difficulty": 1,
-        "topics": [
-          "strings"
-        ]
+        "practices": ["string","input and output"],
+        "prerequisites": ["baics"],
+        "difficulty": 1
       },
       {
         "slug": "leap",
         "name": "Leap",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -53,674 +53,441 @@
...
         "name": "Hello World",
         "uuid": "3eeadff6-82b6-43be-8834-d9ef258e454d",
-        "practices": [],
-        "prerequisites": [],
-        "difficulty": 1,
-        "topics": [
-          "strings"
-        ]
+        "practices": ["string","input and output"],
+        "prerequisites": ["baics"],
```
KevDi, line 57: Found a typo here i think you mean basic or?

KevDi (commented):
Found just one typo in the beginning .
Rest looks from my point of view good to go.

config.json
rakshitdabral: Yep thanks..will update it by tomorrow

kytrinyx:
Thank you so much for taking the time to do this work.

From the documentation (https://exercism.org/docs/building/tracks/practice-exercises)

> The practices key should list the slugs of Concepts that this Practice Exercise actively allows a student to practice.

> The prerequisites key lists the Concepts that a student must have completed in order to access this Practice Exercise.

Since this track only has one `wip` concept, we can't merge this yet. I'm going to put it on hold so that we can bring in these values if and when concepts do get implemented.

PR #511, opened 2022-02-27
Implement the exercise "isbn-verifier"

KevDi:
Implementation of the Exercise ISBN-Verifier.
This is my first Exercise which i ported to the CPP Track.
If wanted i will port more exercises to the CPP Track


config.json
```diff
@@ -721,6 +721,28 @@
...
+      {
+        "slug": "isbn-verifier",
+        "name": "Isbn Verifier",
+        "uuid": "5fcc6fd8-1dc7-4493-970a-731c7602b229",
+        "practices": [],
+        "prerequisites":[
+            "basics",
+            "bools",
+            "conditionals",
+            "lists",
```
siebenschlaefer, line 734: Why "lists"?

exercises/practice/isbn-verifier/example.cpp
```diff
@@ -0,0 +1,24 @@
+#include "example.h"
+#include <regex>
+#include <vector>
+
+namespace isbn_verifier {
+    bool is_valid(const std::string& isbn) {
+        auto raw_string = std::regex_replace(isbn, std::regex("-"), "");
+        
+        if(!std::regex_match(raw_string, std::regex("\\d{9}[0-9|xX]"))) {
```
siebenschlaefer, line 9: `[0-9|xX]` includes the character `'|'`.
Also, I'm curious: Why the lowercase `'x'`?

exercises/practice/isbn-verifier/example.cpp
```diff
@@ -0,0 +1,24 @@
+#include "example.h"
+#include <regex>
+#include <vector>
```
siebenschlaefer, line 3: `<vector>` is not used in this solution.

exercises/practice/isbn-verifier/isbn_verifier_test.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+TEST_CASE("empty_isbn") {
+    REQUIRE_FALSE(isbn_verifier::is_valid(""));
+}
+
+TEST_CASE("input_is_9_characters") {
+    REQUIRE_FALSE(isbn_verifier::is_valid("134456729"));
+}
+
+TEST_CASE("invalid characters are not ignored after checking length") {
```
siebenschlaefer, line 65: This is the only name with spaces. Everywhere else they are replaced with dashes.

exercises/practice/isbn-verifier/isbn_verifier_test.cpp
```diff
@@ -0,0 +1,76 @@
+#include "isbn_verifier.h"
+#include "test/catch.hpp"
```
siebenschlaefer, line 2: In the `_test.cpp` files of the other exercises that `#include` looks like this:
```cpp
#ifdef EXERCISM_TEST_SUITE
#include <catch2/catch.hpp>
#else
#include "test/catch.hpp"
#endif
```

exercises/practice/isbn-verifier/example.h
```diff
@@ -0,0 +1,10 @@
+#if !defined(EXAMPLE_H)
+#define EXAMPLE_H
```
siebenschlaefer, line 2: I think in the other exercises we use the name of the exercise for the include guard, so `ISBN_VERIFIER_H` instead of `EXAMPLE_H`

siebenschlaefer:
Most of it looks good to me.
I checked the tests cases (`isbn_verifier_test.cpp` and `tests.toml`), they are an exact translation of the `canonical-data.json`. The solution seems to be correct (except for one small detail in a regular expression).
I did not check whether the `config.json` is correct.

siebenschlaefer:
Looks good to me. Now let's wait for the maintainers.

> Is there a better way?

I don't know. There's some tooling that performs validation checks but I never got that to work on my machine.
Usually I just hope that @patricksjackson and the other maintainers will perform those integration and system-level tests.

KevDi:
I moved here also the Example Files to the `.meta` Folder

kytrinyx (approved):
Thank you so much for implementing the exercise @KevDi.

I'm sorry it's taken so long to get this approved. I'm going to spend some time this week going through all the open PRs in this repository, to see if we can get things into shape.

I made a few tiny tweaks for consistency with other exercises in the track, and did a new `configlet sync` to bring in changes that were made upstream since you opened the PR.

Once CI is happy I'll go ahead and merge this.

PR #515, opened 2022-03-03
Largest series product

KevDi:
Implementation of the Exercise Largest Series Product.

I'm a little bit unhappy with the cast for checking if the size of the `series` is smaller then the size of the `sequence`.

### DIFF
--- a/exercises/practice/largest-series-product/example.cpp
+++ b/exercises/practice/largest-series-product/example.cpp
@@ -1,0 +1,36 @@
+#include "example.h"
+#include <numeric>
+#include <algorithm>
+#include <stdexcept>
+#include <regex>
+namespace largest_series_product {
+
+    int largest_product(const std::string& series, int size) {
+        if (size == 0) {
+            return 1;
+        }
+
+        if(size < 0) {
+            throw std::domain_error("span must not be negative");
+        }
+
+        if ((int)series.size() < size) {
+            throw std::domain_error("span must be smaller than string length");
+        }
+
+        if (!std::regex_match(series, std::regex("^\\d*$"))) {
+            throw std::domain_error("digits input must only contain digits");
+        }
+
+        int max = 0;
+        for(size_t i = 0; i < series.size() - size + 1; ++i) {
+            auto start_iter = series.begin() + i;
+            auto end_iter = series.begin() + i + size;
+            auto val = std::accumulate(std::next(start_iter), end_iter, (*start_iter) - '0', [](int value, char current_val){
+                return value *= (current_val - '0');
+            });
+            max = std::max(max, val);
+        }
+        return max;
+    }
+} // largest_series_product
--- a/exercises/practice/largest-series-product/largest_series_product.cpp
+++ b/exercises/practice/largest-series-product/largest_series_product.cpp
@@ -1,0 +1,5 @@
+#include "largest_series_product.h"
+
+namespace largest_series_product {
+
+} // largest_series_product
--- a/exercises/practice/largest-series-product/largest_series_product.h
+++ b/exercises/practice/largest-series-product/largest_series_product.h
@@ -1,0 +1,8 @@
+#if !defined(LARGEST_SERIES_PRODUCT_H)
+#define LARGEST_SERIES_PRODUCT_H
+
+namespace largest_series_product {
+
+} // largest_series_product
+
+#endif // LARGEST_SERIES_PRODUCT_H
--- a/exercises/practice/largest-series-product/largest_series_product_test.cpp
+++ b/exercises/practice/largest-series-product/largest_series_product_test.cpp
@@ -1,0 +1,67 @@
+ #include "largest_series_product.h"
+ #ifdef EXERCISM_TEST_SUITE
+ #include <catch2/catch.hpp>
+ #else
+ #include "test/catch.hpp"
+ #endif
+
+// largest-series-product test case data version 1.2.0
+
+TEST_CASE("finds_the_largest_product_if_span_equals_length") {
+    REQUIRE(largest_series_product::largest_product("29", 2) == 18);
+}
+
+#if defined(EXERCISM_RUN_ALL_TESTS)
+TEST_CASE("can_find_the_largest_product_of_2_with_numbers_in_order") {
+    REQUIRE(largest_series_product::largest_product("0123456789", 2) == 72);
+}
+
+TEST_CASE("can_find_the_largest_product_of_2") {
+    REQUIRE(largest_series_product::largest_product("576802143", 2) == 48);
+}
+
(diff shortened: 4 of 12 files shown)


exercises/practice/largest-series-product/example.cpp
```diff
@@ -0,0 +1,36 @@
...
+    int largest_product(const std::string& series, int size) {
+        if (size == 0) {
+            return 1;
+        }
+
+        if(size < 0) {
+            throw std::domain_error("span must not be negative");
+        }
+
+        if ((int)series.size() < size) {
```
siebenschlaefer, line 17: Please uses a named cast instead, in this case a `static_cast`. Either `static_cast<int>(series.size()) < size` or since `size` is non-negative at that point `series.size() < static_cast<std::size_t>(size)`.

exercises/practice/largest-series-product/example.cpp
```diff
@@ -0,0 +1,36 @@
...
+            throw std::domain_error("span must be smaller than string length");
+        }
+
+        if (!std::regex_match(series, std::regex("^\\d*$"))) {
```
siebenschlaefer, line 21: I don't think you need `^` and `$` because `std::regex_match()` matches the whole sequence.

exercises/practice/largest-series-product/example.cpp
```diff
@@ -0,0 +1,36 @@
+#include "example.h"
+#include <numeric>
+#include <algorithm>
+#include <stdexcept>
+#include <regex>
+namespace largest_series_product {
+
+    int largest_product(const std::string& series, int size) {
```
siebenschlaefer, line 8: Here and in the `.h` file the parameters are named `series` and `size`. In the canonical-data.json the parameters are named `digits` and `span`, I'd suggest keeping the names in the example solution (BTW: the error messages use the names `digits` and `span`).

exercises/practice/largest-series-product/largest_series_product.cpp
```diff
@@ -0,0 +1,5 @@
+#include "largest_series_product.h"
+
+namespace largest_series_product {
+
+} // largest_series_product
```
siebenschlaefer, line 5: I'd suggest `// namespace largest_series_product` instead of `// largest_series_product`.

exercises/practice/largest-series-product/largest_series_product.h
```diff
@@ -0,0 +1,8 @@
+#if !defined(LARGEST_SERIES_PRODUCT_H)
+#define LARGEST_SERIES_PRODUCT_H
+
+namespace largest_series_product {
+
+} // largest_series_product
```
siebenschlaefer, line 6: I'd suggest `// namespace largest_series_product` instead of `// largest_series_product` (same as above).

exercises/practice/largest-series-product/largest_series_product_test.cpp
```diff
@@ -0,0 +1,67 @@
+ #include "largest_series_product.h"
+ #ifdef EXERCISM_TEST_SUITE
+ #include <catch2/catch.hpp>
+ #else
+ #include "test/catch.hpp"
+ #endif
```
siebenschlaefer, line 6: These preprocessor instructions in lines 1-6 have a leading space. Is that on purpose?

exercises/practice/largest-series-product/largest_series_product_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+// largest-series-product test case data version 1.2.0
```
siebenschlaefer, line 8: IIRC we don't use those version numbers anymore.

exercises/practice/largest-series-product/largest_series_product_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+TEST_CASE("can_find_the_largest_product_of_3") {
+    REQUIRE(largest_series_product::largest_product("1027839564", 3) == 270);
+}
+
+TEST_CASE("can_find_the_largest_product_of_5_with_numbers_in_order") {
+    REQUIRE(largest_series_product::largest_product("0123456789", 5) == 15120);
+}
+
+TEST_CASE("can_find_the_largest_product_of_a_big_number") {
```
siebenschlaefer, line 35: In the [canonical-data.json](https://github.com/exercism/problem-specifications/blob/main/exercises/largest-series-product/canonical-data.json#L76) this test is called "can get the largest product of a big number" with a "get".

exercises/practice/largest-series-product/largest_series_product_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+TEST_CASE("reports_zero_if_the_only_digits_are_zero") {
+    REQUIRE(largest_series_product::largest_product("0000", 2) == 0);
+}
+
+TEST_CASE("reports_zero_if_all_spans_include_zero") {
+    REQUIRE(largest_series_product::largest_product("99099", 3) == 0);
+}
+
```
siebenschlaefer, line 46: One test is missing: [rejects span longer than string length](https://github.com/exercism/problem-specifications/blob/main/exercises/largest-series-product/canonical-data.json#L104-L115).

exercises/practice/largest-series-product/largest_series_product_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+TEST_CASE("reports_1_for_empty_string_and_empty_product_(0_span)") {
+    REQUIRE(largest_series_product::largest_product("",0) == 1);
```
siebenschlaefer, line 48: I'd suggest separating parameters with comma and space, not just comma.

exercises/practice/largest-series-product/largest_series_product_test.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+TEST_CASE("reports_1_for_nonempty_string_and_empty_product_(0_span)") {
+    REQUIRE(largest_series_product::largest_product("123",0) == 1);
```
siebenschlaefer, line 52: I'd suggest separating parameters with comma and space, not just comma (see above).

siebenschlaefer (commented):
Most of it looks good to me, IMHO just a few details could be improved.
Also, shouldn't `example.h` and `example.cpp` be in the `.meta/` directory?
I did not check the `.json` files.

KevDi:
I have made the mentioned changes.
Many thanks for them =)

kytrinyx (approved):
@KevDi Thanks so much for doing the work to add this exercise! There was a conflict with `main` which I've fixed up, and CI was failing, so I'm tweaking a few things to get that smoothed out.

Once CI is happy I'll make sure the exercise is up-to-date with the most recent changes on problem-specifications and then get this merged.

PR #517, opened 2022-03-06
Implementation of the Darts Exercise

KevDi:
My Implementation of the Exercise Darts.


exercises/practice/darts/example.cpp
```diff
@@ -0,0 +1,20 @@
+#include "example.h"
+#include <cmath>
+
+namespace darts {
+
+    int score(float x, float y) {
+        auto dart_location = sqrt(x*x + y*y);
```
siebenschlaefer, line 7: - Officially you need the prefix `std::` when you call a function from `<cmath>` like `sqrt()`. The major implementations of the standard library also define `::sqrt()` in `<cmath>` but that's not guaranteed by the standard.
- `dart_location` is not really a "location" but a "distance" between the center and the dart. I'd suggest a variable name like `distance`.
- Instead of `std::sqrt(a * a + b * b)` you can call `std::hypot(a, b)`. IMHO that would express the intent a little bit better.

siebenschlaefer (commented):
Looks good to me.

I checked `darts_test.cpp` and `tests.toml`, they are an exact translation of the `canonical-data.json`.
The initial solution is just that: the expected blank solution, the example solution passes the tests and looks good.
I checked `test/catch.hpp`, it is the same catch2-v2.13.6 as everywhere else.
I did not check the JSON files.

I have a few suggestions but nothing major:

siebenschlaefer:
I forgot one thing: Shouldn't `example.h` and `example.cpp` be in the `.meta/` directory?

KevDi:
That's what I was unsure. I checked your pull request for the `two-fer` and there it looks like the `example.h/.cpp` are in the top level but of course I can move them to the `.meta` folder. I will do this also for the two other exercises I implemented

KevDi:
I have make the mentioned changes thank you again =)

siebenschlaefer:
Looks good to me.  
Let's wait what the maintainers say.

PR #518, opened 2022-03-14
Implementation of exercise Bank account

KevDi:
Implementation of the Bank Account Exercise. 
As always i'm happy to hear some improvements =)

### DIFF
--- a/exercises/practice/bank-account/.meta/example.cpp
+++ b/exercises/practice/bank-account/.meta/example.cpp
@@ -1,4 +1,4 @@
-#include "example.h"
+#include "bank_account.h"
 
 #include <stdexcept>
 
--- a/exercises/practice/bank-account/.meta/example.cpp
+++ b/exercises/practice/bank-account/.meta/example.cpp
@@ -1,9 +1,54 @@
 #include "example.h"
 
-void Bankaccount::open() const { return; }
+#include <stdexcept>
 
-void Bankaccount::deposit(size_t amount) { balance_ += amount; }
+void Bankaccount::open() {
+    if (open_) {
+        throw std::logic_error{"account already open"};
+    }
+    open_ = true;
+    balance_ = 0;
+}
 
-void Bankaccount::withdraw(size_t amount) { balance_ -= amount; }
+void Bankaccount::deposit(int amount) {
+    std::lock_guard<std::mutex> guard(mutex_);
+    check_account_open();
+    check_amount_greater_zero(amount);
 
-size_t Bankaccount::balance() const { return balance_; }
+    balance_ += amount;
+}
+
+void Bankaccount::withdraw(int amount) {
+    std::lock_guard<std::mutex> guard(mutex_);
+
+    check_account_open();
+    check_amount_greater_zero(amount);
+
+    if (amount > balance_) {
+        throw std::logic_error{"amount must be less than current balance"};
+    }
+    balance_ -= amount;
+}
+
+void Bankaccount::close() {
+    check_account_open();
+    open_ = false;
+}
+
+int Bankaccount::balance() {
+    std::lock_guard<std::mutex> guard(mutex_);
+    check_account_open();
+    return balance_;
+}
+
+void Bankaccount::check_account_open() const {
+    if (!open_) {
+        throw std::logic_error{"account not open"};
+    }
+}
+
+void Bankaccount::check_amount_greater_zero(int amount) const {
+    if (amount <= 0) {
+        throw std::logic_error{"amount must be greater than 0"};
+    }
+}
--- a/exercises/practice/bank-account/.meta/example.cpp
+++ b/exercises/practice/bank-account/.meta/example.cpp
@@ -2,16 +2,17 @@
 
 #include <stdexcept>
 
+namespace Bankaccount {
 void Bankaccount::open() {
     if (open_) {
-        throw std::logic_error{"account already open"};
+        throw std::runtime_error{"account already open"};
     }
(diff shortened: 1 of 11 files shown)


exercises/practice/bank-account/.meta/example.cpp
```diff
@@ -0,0 +1,54 @@
...
+void Bankaccount::open() {
+    if (open_) {
+        throw std::logic_error{"account already open"};
+    }
+    open_ = true;
+    balance_ = 0;
+}
+
+void Bankaccount::deposit(int amount) {
+    std::lock_guard<std::mutex> guard(mutex_);
```
siebenschlaefer, line 14: Since C++17 this can be written as `std::lock_guard guard(mutex_);`, the template parameter will be deduced by the compiler.

exercises/practice/bank-account/.meta/example.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+void Bankaccount::withdraw(int amount) {
+    std::lock_guard<std::mutex> guard(mutex_);
+
+    check_account_open();
+    check_amount_greater_zero(amount);
+
+    if (amount > balance_) {
+        throw std::logic_error{"amount must be less than current balance"};
+    }
```
siebenschlaefer, line 29: The condition and the error message are slightly contradicting: Is `amount == balance_` allowed?

exercises/practice/bank-account/.meta/example.h
```diff
@@ -0,0 +1,24 @@
+#if !defined(BANK_ACCOUNT_H)
+#define BANK_ACCOUNT_H
+
+#include <cstddef>
```
siebenschlaefer, line 4: `<cstddef>` is not needed for this solution.

exercises/practice/bank-account/bank_account_test.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+TEST_CASE("deposit_into_closed_account_throws_error") {
+    Bankaccount account{};
+    account.open();
+    account.close();
+
+    REQUIRE_THROWS_AS(account.deposit(50), std::logic_error);
+}
+
+TEST_CASE("withdraw_into_closed_account_throws_error") {
```
siebenschlaefer, line 71: "withdraw_from" instead of "withdraw_into"
(Disclaimer: I'm not a native English speaker.)

siebenschlaefer (commented):
Most of it looks good to me.

Just a few observations/comments/suggestions:

- There's no `namespace`. All the other exercises use their own `namespace`, e.g. `namespace space_age` or `namespace binary_search_tree`. I don't know whether that's required, let's hear from the maintainers.
- Is `amount == 0` an error in `deposit()` and `withdraw()`? From a mathematical or technical point of view: no, it's a NOP operation that doesn't violate any constraints. But from the point of view of the domain: probably, because that operation doesn't make sense.
- Is `std::logic_error` the correct exception? Since the methods can be called concurrently the caller has no way to check whether an account is currently open or if the balance is sufficient, between the check and the actual operation there might be method call from another thread that closes the account or withdraws everything. So IMHO the exception should not be a `std::logic_error` which is commonly used for errors that are "preventable" (see [CppReference.com](https://en.cppreference.com/w/cpp/error/logic_error)) and caused by "programmer mistakes" (see [Microsoft](https://docs.microsoft.com/en-us/cpp/standard-library/stdexcept?view=msvc-170)). I'd rather use a `std::runtime_error` (or a subtype).
- Should `open()` and `close()` be guarded? Id say yes definitely, but that's much harder to test.

This is one of the few exercises that don't have a language-independent set of tests (`canonical-data.json`). There are just the instructions and they are rather short. I've read the tests, they make sense to me and seem to cover everything (maybe except mixed `.deposit()` and `.withdraw()` transactions, the amount `0`, and whether .open()` and `.close()` are properly guarded.)

Just to make that explicit, here are some requirements of the tests that are not included in the instructions:
- The instructions only require `close()`. Here the tests additionally require `open()`, that a `Bankaccount` is initially not open, and that an account can be re-opened after it has been closed.
- The tests require that a re-opened account has a balance of `0`.
- The tests require the account to be non-negative while the instructions don't mention that. But I think that's a reasonable additional requirement.

I've checked the `CMakeLists.txt`, `test/catch.hpp`, and `test/tests-main.cpp`, they are exactly the same as in the other exercises.  
The example solution looks correct, compiles and passes the tests.  
I did not check the JSON files.

KevDi:
any news on that?

kytrinyx (approved):
Thanks @KevDi. I've tweaked a few minor things to make all the various tools happy, and will get this merged as soon as CI is confirmed to be green.

kytrinyx:
Hm, CI isn't happy, but I'm not sure why. It's the end of day for me here; I'll take another look at this tomorrow if nobody else has gotten to it in the meantime.

KevDi:
@kytrinyx it seems like there is a issue with missing reference to `pthread`. Do i have to link the Program against the `pthreads`?

kytrinyx:
@KevDi That's a good guess, and worth trying.

KevDi:
@kytrinyx i added the link to the `CMakeLists.txt` let's see if it will run. I tested it on my system (Linux) and it works don't know how it is on Windows.

KevDi:
@kytrinyx  Seems to be working 😁

kytrinyx:
@KevDi Woo! ✨ 

If someone has problems on Windows, we'll ask them to help us getting working :)

PR #521, opened 2022-04-05
[WIP] Implement exercise affine-cipher

chgraef:
Starting to implement affine-cipher exercise.


kytrinyx:
I'm going to go ahead and close this, as we've not heard back from the original author.

@chgraef if you come back and want to work on it, and nobody has implemented affine-cipher in the meanwhile, please holler, and we can open the PR back up.

PR #523, opened 2022-04-19
feature: added test case to acronym exercise to cover missed case

Llcoolsouder:
The current iteration of the acronym exercise missed the case where the first letter of the phrase is lowercase.

I had a mentee start their solution with 
```cpp
std::string result(1, input[0]);
```
and all tests passed.

My new test accounts for this situation.

### DIFF
--- a/exercises/practice/acronym/acronym_test.cpp
+++ b/exercises/practice/acronym/acronym_test.cpp
@@ -62,4 +62,13 @@ TEST_CASE("hyphenated")
     REQUIRE(expected == actual);
 }
 
+TEST_CASE("all_lowercase")
+{
+    const string actual = acronym::acronym("you only live once");
+
+    const string expected{"YOLO"};
+
+    REQUIRE(expected == actual);
+}
+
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
The C++ track (usually) just translates the test cases from the language-agnostic [problem specification](https://github.com/exercism/problem-specifications/blob/main/exercises/acronym/canonical-data.json) to C++.  
The best place to suggest changes like this is at [github.com/exercism/problem-specifications/issues](https://github.com/exercism/problem-specifications/issues).

BTW: Two somehow related suggestions already exist: [#1431](https://github.com/exercism/problem-specifications/issues/1431) and [#1432](https://github.com/exercism/problem-specifications/issues/1432).

kytrinyx:
@Llcoolsouder Thanks for the suggestion!

If you would like to propose these changes in problem-specifications, please open an issue there.
If they get accepted upstream, then we will update the test suite here to include them.

I'm going to go ahead and close this.

PR #534, opened 2022-08-23
Renamed concept strings to log levels

krish40786:


kytrinyx (approved):
Since this doesn't change an C++ code, only renames files, I'm going to go ahead and merge this. We can iterate if we find that we need to.

PR #535, opened 2022-08-27
Add tests for valid queen positions

citizen428:
The tests for the C++ version of this exercise do not follow the problem specification and don't validate that the position is within [0, 8).

This PR adds two tests, one verifying the row with a value < 0 and one verifying the column with a value > 7.


kytrinyx:
@citizen428 Thanks for the contribution, and I'm sorry it's taken so long to get back to you.

All the practice exercises in this track are implemented by translating the specifications in the [problem-specifications repository](https://github.com/exercism/problem-specifications).

This exercise has undergone significant changes in the years since it was implemented in the C++ track, so I've gone ahead and overhauled the entire exercise in #554 to reflect the new data in the canonical specification, which now includes tests for positions off the board in either direction.

citizen428:
OK, I'll close this PR then.

PR #537, opened 2022-09-02
Update instructions.md

tranzzik:
Typo at lines 11/12, black and white queens were the other way around (considering the bottom left corner is the point with (1, 1) coordinates)


tranzzik:
> I don't think this is an error, the origin `(0, 0)` is at the upper left corner. Look at the test [string_representation](https://github.com/exercism/cpp/blob/main/exercises/practice/queen-attack/queen_attack_test.cpp#L39-L49): The white queen with the coordinates `(2, 4)` is on the row with index `2` and the column with index `4`. But I agree, that's counterintuitive for anybody who plays chess.

You're right, my bad, sorry!

PR #539, opened 2022-10-03
Fix nucleotide names

ludeum:
I noticed that in the tests nucleotide names are incorrect, so I am opening a small pull request to fix them :)


kytrinyx:
Thanks for the fix, @giorgospetkakis.

The practice exercises in this track use the data in the [problem-specifications](https://github.com/exercism/problem-specifications) repository for the descriptions, metadata, and test cases.

The exercise was out of date, and I've submitted a PR to bring it into sync with the upstream specification.

This PR will be closed when https://github.com/exercism/cpp/pull/550 is merged.

PR #542, opened 2022-10-19
Update instructions.md

0zAND1z:
Remove the word `one` from the instruction, which is redundant while reading.


kytrinyx:
Thanks for the fix @0zAND1z.

The practice exercises in this track use the data in the [problem-specifications](https://github.com/exercism/problem-specifications) repository for the descriptions, metadata, and test cases.

The exercise was out of date, and I've submitted a PR to bring it into sync with the upstream specification.

This PR will be closed when https://github.com/exercism/cpp/pull/547 is merged.

Issue #370, opened 2020-10-18
Update each tracks tests.toml file


patricksjackson:
In #362 a tests.toml file was added to each track in order to track the list of implemented tests for that exercise. By default every test was listed as existing, which I doubt is the reality. We need to go through each test to match the implementation here with the problem specifications repo.

More details in #362

- [ ] acronym
- [ ] all-your-base
- [ ] allergies
- [ ] anagram
- [ ] armstrong-numbers
- [ ] atbash-cipher
- [ ] beer-song
- [ ] binary
- [ ] binary-search
- [ ] binary-search-tree
- [ ] bob
- [ ] circular-buffer
- [ ] clock
- [ ] collatz-conjecture
- [ ] complex-numbers
- [ ] crypto-square
- [ ] difference-of-squares
- [ ] etl
- [ ] food-chain
- [ ] gigasecond
- [ ] grade-school
- [ ] grains
- [ ] hamming
- [ ] hello-world
- [ ] hexadecimal
- [ ] isogram
- [ ] leap
- [ ] luhn
- [ ] matching-brackets
- [ ] meetup
- [ ] nth-prime
- [ ] nucleotide-count
- [ ] pangram
- [ ] pascals-triangle
- [ ] phone-number
- [ ] prime-factors
- [ ] protein-translation
- [ ] queen-attack
- [ ] raindrops
- [ ] reverse-string
- [ ] rna-transcription
- [ ] robot-name
- [ ] robot-simulator
- [ ] roman-numerals
- [ ] say
- [ ] scrabble-score
- [ ] secret-handshake
- [ ] series
- [ ] sieve
- [ ] space-age
- [ ] sum-of-multiples
- [ ] triangle
- [ ] trinary
- [ ] two-fer
- [ ] word-count

PR #550, opened 2022-11-15
Sync nucleotide-count with problem-specifications

kytrinyx:
This brings the nucleotide-counts exercise up to date with problem-specifications.

For clarity, it makes the change in several commits:

1. Delete entries from `tests.toml` that are not reflected in the test suite (per #370)
2. Delete tests from the nucleotide-count test suite that do not have equivalents in the `canonical-data.json`
3. Reorder nucleotide-count tests to reflect the order in `canonical-data.json`
4. Run `configlet fmt` on nucleotide-count to normalize the `config.json`
5. Normalize nucleotide-count tests to use the inputs and test descriptions from the canonical data.
6. Run `configlet sync` for the exercise and implementing the missing test.

Closes #544
Closes #539

### DIFF
--- a/exercises/practice/nucleotide-count/.docs/instructions.md
+++ b/exercises/practice/nucleotide-count/.docs/instructions.md
@@ -1,10 +1,12 @@
 # Instructions
 
-Each of us inherits from our biological parents a set of chemical instructions known as DNA that influence how our bodies are constructed. All known life depends on DNA!
+Each of us inherits from our biological parents a set of chemical instructions known as DNA that influence how our bodies are constructed.
+All known life depends on DNA!
 
 > Note: You do not need to understand anything about nucleotides or DNA to complete this exercise.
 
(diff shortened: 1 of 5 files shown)


ErikSchierboom (approved):

siebenschlaefer:
I completely missed this back in November.  

Previously implementing this task with a class made sense. A `counter` object gets initialize once, then its member functions `nucleotide_counts()` and `count()` can be called a number of times. I often discussed with my students where the actual counting should happen, in the constructor or in the member functions.

Now that the member function `count()` is gone, (conceptually) it's a two-step function call: Construct the instance (which performs some validation), then call `nucleotide_counts()` with returns the counter. I find that a little bit strange.

To me that looks now like a task for a regular function that performs the validation (it might `throw`) and then returns the `std::map`.
Is there a reason to implement this as a class?

kytrinyx:
> Is there a reason to implement this as a class?

Probably not! I was likely just following what was already there, and then tweaking it to follow the current canonical data. I don't actually write C++, and I don't even remember making this PR (which is a bit disconcerting).

Please feel free to change the implementation, if you have something better in mind.

PR #554, opened 2022-11-16
Sync queen-attack with problem-specifications

kytrinyx:
The tests implemented in the queen-attack test suite had no overlap
with the tests listed in tests.toml.

I've therefore done a complete overhaul in one fell swoop rather than
trying to break it into step-by-step commits.

Closes #535

### DIFF
--- a/exercises/practice/queen-attack/.docs/instructions.md
+++ b/exercises/practice/queen-attack/.docs/instructions.md
@@ -1,27 +1,25 @@
 # Instructions
 
-Given the position of two queens on a chess board, indicate whether or not they
-are positioned so that they can attack each other.
+Given the position of two queens on a chess board, indicate whether or not they are positioned so that they can attack each other.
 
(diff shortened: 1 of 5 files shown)


ErikSchierboom (approved):

siebenschlaefer:
[Previously](https://github.com/exercism/cpp/blob/faf4b1c29dc9fb36386040c5f54ef11693b6b4e9/exercises/practice/queen-attack/queen_attack_test.cpp#L27-L31) the tests did require the solution to throw an exception if the two queens were placed on the same square.  
IMHO that makes sense, and the [canonical-data](https://github.com/exercism/problem-specifications/blob/main/exercises/queen-attack/canonical-data.json#L7-L10) even allows that explicitly.  
I'd love to have that test back, to me as a (low-skilled) chess player a board with two queens on the same square make just as much sense as queens placed outside of the board.

kytrinyx:
That seems like a reasonable extension to me.

PR #560, opened 2022-12-05
Update hello_world.cpp

Lemale13:
Solving « print hello world »


siebenschlaefer:
Hi! These exercises are not meant to be solve via Pull Request.
Have you tried solving it on https://exercism.org/tracks/cpp, either directly in the browser or by using the CLI to download the exercise and submit your solution?

PR #568, opened 2023-03-02
Binary Search approach

bobahop:
An approach for Binary Search

### DIFF
--- a/exercises/practice/binary-search/.approaches/config.json
+++ b/exercises/practice/binary-search/.approaches/config.json
@@ -1,0 +1,15 @@
+{
+  "introduction": {
+    "authors": ["bobahop"],
+    "contributors": []
+  },
+  "approaches": [
+    {
+      "uuid": "68a494d9-964c-4b2f-a31b-5063dfe0ffe7",
+      "slug": "while-with-if-statements",
+      "title": "While with if statements",
+      "blurb": "Use a while loop with if statements to return the answer.",
+      "authors": ["bobahop"]
+    }
+  ]
+}
--- a/exercises/practice/binary-search/.approaches/introduction.md
+++ b/exercises/practice/binary-search/.approaches/introduction.md
@@ -6,6 +6,19 @@ One approach can use a `while` loop with `if` statements.
 ## Approach: `while` loop with `if` statements
 
 ```cpp
+// binary_search.h
+#if !defined(BINARY_SEARCH_H)
+#define BINARY_SEARCH_H
+#include <vector>
+#include <cstddef>
+namespace binary_search {
+    std::size_t find (const std::vector<int> data, int value);
+}  // namespace binary_search
+#endif // BINARY_SEARCH_H
+
+//------------------------------------------------------------------------
+
+// binary_search.cpp
 #include "binary_search.h"
 #include <stdexcept>
 
--- a/exercises/practice/binary-search/.approaches/introduction.md
+++ b/exercises/practice/binary-search/.approaches/introduction.md
@@ -5,8 +5,8 @@ One approach can use a `while` loop with `if` statements.
 
 ## Approach: `while` loop with `if` statements
 
+**binary_search.h**
 ```cpp
-// binary_search.h
 #if !defined(BINARY_SEARCH_H)
 #define BINARY_SEARCH_H
 #include <vector>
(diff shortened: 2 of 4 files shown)


bobahop:
@ErikSchierboom: C++ Binary Search approach ready for review.

siebenschlaefer:
Some small suggestions:

- The function takes its argument by value, so it's (potentially) a copy. Creating a copy in linear time kind of defeats the purpose of the binary search. I'd suggest taking the argument by reference to `const`.
- The return type is `std::size_t` (with the prefix `std::`) but the type of the three index variables is `size_t` (without that prefix). That works on all platforms I know but officially `size_t` is not guaranteed to be defined (unless you include one of the C headers like `<stddef.h>`). I'd suggest adding the prefix `std::`.
- `mid` is defined and initialized outside of the loop. But since it's only *used* inside the loop you could define it in the first line inside the loop where you assign a "real" value to it. That's what [ES.21](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-introduce) ("*Don’t introduce a variable (or constant) before you need to use it*") and [ES.22](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-init) ("*Don’t declare a variable until you have a value to initialize it with*") of the C++ Core Guidelines recommend.
- The `mid` is calculated as `(left + right) / 2`. That's a [quite famous bug](https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html) that was present in many binary search implementations for many years. The problem is that for a very large vector that sum of two integers values might overflow and then `mid` is too small. Use `start + (end - start) / 2` to avoid the overflow.

In the explanations: 

- "the middle value" -> "the middle index", to be precise about indexes and values.
- "if an `if` statement can return" -> "if an `if` statement returns", because "can" sounds optional.

This approach uses a [half-open interval](https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints). I like that very much because it can represent an empty range and therefore doesn't need special cases. You could consider explaining that in a short paragraph.

bobahop:
@siebenschlaefer: Thank you for your comprehensive feedback! All good points, but they are rather nitpicky for my level, so I changed the mid to be defined in the loop, as that was already nagging at me. The rest you are free to modify, preferably after it's merged, so there won't be a bunch of comment changes.

siebenschlaefer:
Yes, the `left + (right - left) / 2` calculation is a little bit nitpicky.  
But I strongly recommend changing the type of the parameter from `const std::vector<int> data` to `const std::vector<int>& data`. Otherwise calling the function will create a copy (in linear time), defeating the whole purpose of its existence.  
And `mid` is now an `int`, I'd suggest changing it back to a `size_t`.

But I'll happily submit a PR after it has been merged, if that's preferred.

ErikSchierboom (approved):

bobahop:
@siebenschlaefer: Sorry, all things look better in the morning (as opposed to being weary just before going to sleep.)

Yes, the lack of referencing the vector was an oversight on my part, so thank you again for that.
I fixed the mid calculation per your recommendation and tweaked the explanation regarding it. I originally got the algorithm from Sedgewick, but it was old.
I put in the `std::` for `size_t`.

Thanks again!

bobahop:
@ErikSchierboom or @siebenschlaefer: I think this might be ready to be merged now?

ErikSchierboom:
@siebenschlaefer I'm happy with this, you? We can always do a follow-up PR to further tweak things.

bobahop:
Okay, I'm thinking the number _times_ itself would be more performant and would work. Let me make that change...

vaeng:
> The rest you are free to modify, preferably after it's merged, so there won't be a bunch of comment changes.

When we squash-merge the whole branch into one commit, it does not really matter how many interaction have happened here, right?

vaeng (approved):

bobahop:
@vaeng: If there are a lot of changes (and the amount of changes suggested here wasn't _that_ many) then they can become "hidden" and unwieldy to deal with. With optimistic merging, only show-stoppers should hold up the merging. Any finessing is encouraged to be done after the merge.

Optimistic merging is described [here](https://exercism.org/blog/freeing-our-maintainers):

>The one exception is that we will continue to accept PRs for Approaches and Articles, and will implement an org-wide optimistic merging policy that aims to populate a baseline of Approaches across exercism, and allow for incremental improvements, using the following rules:

>If the code solves the exercise and is syntactically and semantically idiomatic (i.e. it looks like $LANG code) it should be merged. If not, it should be fixed by the PR author.
If a maintainer wants to make changes to the content (e.g. to improve the advice, tweak things, highlight better/alternative/more-idiomatic approaches) then that should be done in a follow up PR.

siebenschlaefer:
Looks good!

PR #569, opened 2023-03-02
Sieve approach

bobahop:
An approach for Sieve.

### DIFF
--- a/exercises/practice/sieve/.approaches/config.json
+++ b/exercises/practice/sieve/.approaches/config.json
@@ -1,0 +1,15 @@
+{
+  "introduction": {
+    "authors": ["bobahop"],
+    "contributors": []
+  },
+  "approaches": [
+    {
+      "uuid": "49609966-f2d7-42e6-89f5-3e567cc20713",
+      "slug": "vector-of-bools",
+      "title": "Vector of bools",
+      "blurb": "Use a vector of bools to keep track of composite numbers.",
+      "authors": ["bobahop"]
+    }
+  ]
+}
--- a/exercises/practice/sieve/.approaches/introduction.md
+++ b/exercises/practice/sieve/.approaches/introduction.md
@@ -9,6 +9,18 @@ Something to consider is to keep the number of allocations and iterations at a m
 
 ## Approach: `vector` of `bool`s
 
+**sieve.h**
+```cpp
+#if !defined(SIEVE_H)
+#define SIEVE_H
+#include <vector>
+namespace sieve {
+    std::vector<int> primes(int);
+}  // namespace sieve
+#endif // SIEVE_H
+```
+
+**sieve.cpp**
 ```cpp
 #include "sieve.h"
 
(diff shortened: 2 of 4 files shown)


bobahop:
@ErikSchierboom: C++ Sieve approach ready for review.

siebenschlaefer:
Some small suggestions:

- The `std::vector` is named `composites`. *Personally* I'd rather use the name `sieve` because that's the eponymous Sieve of Eratosthenes. But naming is subjective and I can see why you chose `composites`.
- The function reserves a capacity of `limit / 2`. That's not wrong but are you sure that's worth the effort? Even for moderately large limits the number of primes is much smaller that `limit / 2`. If you really want to avoid any resize operation you could use one or multiple of the inequalities that the Wikipedia lists as [Approximations for the nth prime number](https://en.wikipedia.org/wiki/Prime_number_theorem#Approximations_for_the_nth_prime_number) or [Non-asymptotic bounds on the prime-counting function](https://en.wikipedia.org/wiki/Prime_number_theorem#Non-asymptotic_bounds_on_the_prime-counting_function). But *personally* I would omit the `reserve()` completely. All `std::vector` operations increase the capacity by some factor between 1.5 and 2, guaranteeing an amortized linear runtime and it's an optimization beyond the level of explanation.
- The inner `for` loop marks all composites of the current prime, starting with `number + number`. But you can choose a larger starting point: `number * number` because all multiples below that have been marked in a previous iteration. With this small optimization the runtime complexity becomes Θ(n log log n) (see [The Genuine Sieve of Eratosthenes](https://web.archive.org/web/20130514030554/http://www.cs.tufts.edu/~nr/comp150fp/archive/melissa-oneill/Sieve-JFP.pdf) by Melissa O'Neill).

In the explanation:

- "the number being iterated" -> "the current `number`", because it's a little bit simpler

bobahop:
@siebenschlaefer: Thank you again for your comprehensive feedback! As mentioned in the forum, we're looking for optimistic merging for these. I encourage you to bring your expertise to improving them after merging. However, if these are too far below an acceptable quality level, I won't take offense if you prefer I not do them for C++. I admittedly have limited stamina for some details, especially after the burnout of having done so many for so many languages already. Thanks again!

siebenschlaefer:
Sorry, it seems I misunderstood "ready for review".  
If you prefer merging it first and then improving it through PRs that's fine with me.

And no, I do not think the code or explanation is unacceptable.
The code is easy to read, idiomatic, and reasonably efficient, I had no problems understanding the explanation.

ErikSchierboom (approved):

bobahop:
I'm okay with leaving it as is, but re starting with number + number, when we're processing `2`, if we don't set `4` to true for being a composite, won't it get skipped? As processing `3` won't handle it, and then when we iterate to `4` it will be false for being a composite and will be added to the primes. But math isn't my forte, so maybe I'm not seeing something?

bobahop:
@ErikSchierboom or @siebenschlaefer: I think this might be ready to be merged now?

ErikSchierboom:
@siebenschlaefer Any objections to merging this as is and improving things in a follow-up PR?

bobahop:
Okay, I'm thinking the number _times_ itself would be more performant and would work. Let me make that change...

siebenschlaefer:
Looks good!

ErikSchierboom:
Thanks both!

PR #570, opened 2023-03-02
Add simple-linked-list exercise

vaeng:
The #12in23 challenge asks to do the linked-list exercise, which is currently not implemented and should be added during the month of March. As a preliminary task, we include a simpler version that can be built upon.

- [x] Basic file system
- [x] Write tests 
- [x] Write exercise stub
- [x] Write an example solution
- [x] Write optional documentation files


vaeng:
I wonder how we can test that the student made sure that no resources have leaked.

vaeng:
I would like to merge this soon, so the users can get their March badges. Would you have a look @siebenschlaefer or @exercism/maintainers-admin

config.json
```diff
@@ -1,848 +1,863 @@
-{
```
vaeng, line 1: The large change here comes from the configlet format.

config.json
```diff
@@ -1,848 +1,863 @@
...
+        ]
+      },
+      {
+        "slug": "simple-linked-list",
+        "name": "Simple linked list",
+        "uuid": "ffccabe9-9779-4914-9cff-c6f5696c8afe",
+        "practices": ["pointers"],
+        "prerequisites":[],
+        "difficulty": 4,
+        "status": "wip",
```
vaeng, line 790: Should we publish with `wip` or `beta`?

ErikSchierboom: Students won't see the exercise with `wip`.

ErikSchierboom (commented):
Thanks a ton for adding this! I'll leave the C++ review to @siebenschlaefer.

config.json
vaeng: Then I should reverse that to keep the commit small.

config.json
vaeng: That is how I understood that. Should we publish as beta, so more people can see it? Or should we keep it to the (cpp?) maintainers first? that is what wip would do, right?

config.json
vaeng: Strange. I cannot see the changes in my IDE. Even with a revert to main.

config.json
ErikSchierboom: I think it is fine to have this as `beta`

ErikSchierboom:
I'd like to get this PR merged sooner than later, as this is one of the recommended exercises for Mechanical March

@siebenschlaefer do you have time to do a review of this PR?
@vaeng could you look into fixing the diff for the `config.json` file to not include any formatting changes?

vaeng:
> I'd like to get this PR merged sooner than later, as this is one of the recommended exercises for Mechanical March
> 
> @vaeng could you look into fixing the diff for the `config.json` file to not include any formatting changes?

I had to change the end of line sequence. I uploaded with CRLF but it was LF.

ErikSchierboom (approved):
LGTM

siebenschlaefer:
I'm late to the party, sorry.

I have one tiny suggestion: Use `std::size_t` instead of `size_t`. The latter will work on all major compilers but technically that's implementation-specific. After including `<cstddef>` the standard only guarantees the existence of `std::size_t`.

PR #571, opened 2023-03-02
Secret Handshake approaches

bobahop:
Approaches for Secret Handshake. Please let me now if there are any show-stoppers. Anything else is encouraged to be done after the initial merge.

### DIFF
--- a/exercises/practice/secret-handshake/.approaches/config.json
+++ b/exercises/practice/secret-handshake/.approaches/config.json
@@ -1,0 +1,22 @@
+{
+  "introduction": {
+    "authors": ["bobahop"],
+    "contributors": []
+  },
+  "approaches": [
+    {
+      "uuid": "d0cfe906-5844-4bcc-b30a-08341057deb4",
+      "slug": "if-statements-with-reverse",
+      "title": "if statements with reverse",
+      "blurb": "Use if statements with reverse if needed.",
+      "authors": ["bobahop"]
+    },
+    {
+      "uuid": "616c4db1-392f-4033-a37d-e0c9af4d995f",
+      "slug": "iterate-once",
+      "title": "Iterate once",
+      "blurb": "Iterate once even if reverse is needed.",
+      "authors": ["bobahop"]
+    }    
+  ]
+}
--- a/exercises/practice/secret-handshake/.approaches/if-statements-with-reverse/content.md
+++ b/exercises/practice/secret-handshake/.approaches/if-statements-with-reverse/content.md
@@ -37,7 +37,7 @@ std::vector<std::string> commands(unsigned int signal)
 The `commands` function starts by defining the `vector` to hold the returned actions.
 
 It then has a series of `if` statements, each of which uses the [bitwise AND operator][bitwise-operators] to check if the input signal contains a particular action.
-Each action is represented with a binary literal, but could just as well be represented by its decimal value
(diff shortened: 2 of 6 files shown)


vaeng:
Thank you Bob, great work. Good details. The only thing I see is, that not [every sentence is on a single line](https://exercism.org/docs/building/markdown/markdown#h-one-sentence-per-line), but that is not a showstopper.

vaeng (approved):

siebenschlaefer:
`if-statements-with-reverse/content.md`

Lines 67-74 talk about member functions `.begin()` and `.end()` vs. free functions `begin()` and `end()`. To me lines 73-74 sound like the free functions return const-iterators. That's not true, in this context `result.begin()` returns exactly the same as `begin(result)`. Also, `std::reverse()` would not be able to modify the range if those iterators were const-iterators.  
These lines 67-74 did either confuse me or they are incorrect.  
I'd suggest something like "You can either call `result.begin()` and `result.end()` as member functions, or `begin(result)` and `end(result)` as free functions, the result will be the same (at least in this exercise.)"

`iterate-once/content.md`

Line 28 names the constant `REVERSE` in "ALL_CAPS".  
[ES.9 of the C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-not-CAPS) recommends: "*Avoid ALL_CAPS names*".  
By convention macros use identifiers in ALL_CAPS, if a user has their own macro with the same name or uses a third-party library with that macro the resulting error can be hard to find.

Line 47 says that `actions` and `REVERSE` are in an anonymous namespace so that they are only accessible within the current file. Technically that's not necessary because in C++ (other than in C) global `const` variables have internal linkage. I still like the unnamed namespace because I think it communicates the intent better, just the explanation seems to imply an incorrect thing.
BTW: I think the official term is "unnamed namespace".

And a few tiny nitpicks:

- Does Exercism's [Markdown style guide](https://exercism.org/docs/building/markdown/style-guide) apply? It wants one sentence per line.
- On `contents.md` lines 27-31 you made the bitmasks line up vertically, but not all `result.emplace()` and `std::reverse()`.  
    On lines 54-58 you made `result.empty()` and `std::reverse()` line up vertically but not the bitmasks.  
    And lines 3-7 in `snippet.txt` format the code in a slightly different third way.  
    Is the on purpose?
- Is it "represented with" or "represented by"? You used both. (I'm not a native English speaker.)
- In line 61 it's `vector` (with backticks), in line 65 it's vector (without backticks).
- I'd prefer spaces around operators like `+=`.

bobahop:
It's quite possible I misunderstood the difference

>Returns exactly c.begin(), which is typically an iterator to the beginning of the sequence represented by c. If C is a standard [Container](https://en.cppreference.com/w/cpp/named_req/Container), this returns C::iterator when c is not const-qualified, and C::const_iterator otherwise.

and 

>Returns exactly std::begin(c), with c always treated as const-qualified. If C is a standard [Container](https://en.cppreference.com/w/cpp/named_req/Container), this always returns C::const_iterator.

Since the issue doesn't particularly interest me at the moment, I'll just point out the two ways to do it without further explanation. I know Kate Gregory explained in one of her videos why she prefers `begin(whatever)`, but I don't remember why and don't feel like digging for it now.

I'll make some other tweaks and edit this comment when I'm done.

PR #574, opened 2023-03-06
Update min required Ubuntu version to 20.04

vaeng:
> CI: .github#L1
> The ubuntu-18.04 environment is deprecated, consider switching to ubuntu-20.04(ubuntu-latest), or ubuntu-22.04 instead. For more details see https://github.com/actions/virtual-environments/issues/6002

Ubuntu 18.04 will be phased out in GitHub actions. We should not wait too long to change along. That would also free the track from using clang 6 with C++14 to clang 10 with C++17.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   linux-min:
     name: Linux Min Config
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         compiler: [clang++-6.0, g++-7]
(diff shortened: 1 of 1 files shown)


vaeng:
We have to move very soon, tasks are failing.
> We have started the deprecation process for Ubuntu 18.04. While the image is being deprecated, You may experience longer queue times during peak usage hours. Deprecation will begin on 2022/08/08 and the image will be fully unsupported by **2023/04/03**

I am not sure, if the line that I found is the only thing we have to change.

@ErikSchierboom

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   linux-min:
     name: Linux Min Config
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         compiler: [clang++-6.0, g++-7]
```
vaeng, line 17: Does this need updating?

ErikSchierboom (approved):

PR #579, opened 2023-03-14
nucleotide-count: convert the class to a function

siebenschlaefer:
Before PR #550 the constructor was required to validate the input, the member function `count(char)` had to return the count for a specific nucleotide, the member function `nucleotide_counts()` had to return a `std::map` with all four counts.

But PR #550 has changed that significantly. After syncing the tests with the canonical data there's no call to `count(char)` anymore, so each test looks like a two-step function call:

```
const nucleotide_count::counter dna("G");
const auto actual = dna.nucleotide_counts();
```

I'd call that unnecessarily complicated and unidiomatic, there's no longer a good reason to require a `class`. Let's turn the class `counter` into a function:

```
const auto actual = nucleotide_count::count("G");
```

### DIFF
--- a/exercises/practice/nucleotide-count/.meta/example.cpp
+++ b/exercises/practice/nucleotide-count/.meta/example.cpp
@@ -4,16 +4,17 @@
 namespace nucleotide_count
 {
 
-counter::counter(std::string const& sequence)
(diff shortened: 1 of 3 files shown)


vaeng (approved):
I really like that change. I remember that felt forced. As if you had to use OOP no matter what.

PR #598, opened 2023-04-24
Concept docs for code organization (closes #597)

### LINKED ISSUE #597: Docs for "Code Organization" Concept
Apart from the exercise material, the Code Organization part also needs docs:

- [ ] about.md
- [ ] introduction.md
- [ ] links.json
- [ ] .meta/config.json

vaeng:


concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,161 @@
...
+namespace my_other_ns {
+    int foo() {
+        return -2;
+    }
+}
+
+int myresult{my_ns::foo() + my_other_ns::foo() * my_ns::my_inner_ns::baz()};
+```
+
+Deeply nested namespaces might be too verbose.
```
b8horpet, line 126: nesting namespace will have the problem of relative namespaces
```cpp
namespace bar
{
    int baz() { return 2; }
}

namespace foo {
    namespace bar{
        int baz() { return 1; }
    }
    int which()
    {
        //return bar::baz();
        return ::bar::baz();
    }
}
```

b8horpet: also unnamed namespaces help hiding stuff

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,161 @@
...
+```
+
+## Standard Library Header Files
+ 
+The standard library offers many common functions, algorithms, and data structures.
+They are grouped into header files.
+They are included with angled braces `<>` instead of double quotes `"`.
+The difference with the braces is the location, where the compiler searches for the header files.
+The search in the current project is skipped for the angled braces version, and it directly starts in the system's include directories.
+The standard library headers are located in the `std` namespace.
```
b8horpet, line 149: the old C std lib is also available both in namespace std or globally
`#include <stdio.h>`
vs
`#include <cstdio>`

b8horpet (approved):

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,161 @@
...
+}
+```
+
+Attention: the `;` is needed after the declaration, but not necessarily after the definition.
+
+## Header Files
+
+Often, declarations are bundled together in header files. 
+The most common file extension for header files is `.h`.
+Some projects use `.hpp` or skip the extension completely.
```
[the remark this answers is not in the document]
vaeng: I'm not sure we should go into that much detail. Especially since I aim to keep the C heritage very low-key. The difference between `stdio.h` and `cstdio` would also require some kind of history lesson to be understandable, right?

I would put this concept as the second total the students would see and it is already at  161 lines.

Would this information have been crucial during your first C++ projects?

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,161 @@
...
+It is possible to include the content of other files with the `include` directive.
+Includes should be stated at the top of the file.
+The snippet below shows how to use the `example()` function that was declared in the `myProject.h` file.
+
+```cpp
+#include "myProject.h"
+example();
+```
+
+If the `myProject.h` is located in the `some_folder` and not in the root directory, you would have to use `#include "some_folder/myProject.h"`.
```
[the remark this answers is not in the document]
vaeng: Setting up the compiler can be very tricky for some students. We should add a tutorial at some point that covers Windows, Mac and Linux distros.

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Include Guards
+
+It does not matter if the same file is included multiple times within a project.
+Header files should not contain definitions.
+The complete project cannot have the same definition more than once.
```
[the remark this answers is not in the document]
vaeng: Templates will come a lot later in the syllabus. I wonder if we should add footnotes everywhere with "there are exceptions to this rule"

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,161 @@
...
+The syntax can be seen below with `MY_HEADER_FILE_H` as a variable.
+
+```cpp
+#ifndef MY_HEADER_FILE_H /* any name uniquely mapped to file name */
+#define MY_HEADER_FILE_H
+// file content
+#endif
+```
+A modern form, that will achieve the same result is the #pragma directive `#pragma once`.
+The problem is, that pragmas are not an official part of the C++ language and the implementation differs from compiler to compiler.
```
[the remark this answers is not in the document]
vaeng: I'll add a sentence to put it in a more positive light.

concepts/code-organization/introduction.md
vaeng: Should we be more specific here? It's already a lot of content. I should talk about the global namespace though.

concepts/code-organization/introduction.md
vaeng: I would not talk too extensively about the finer differences. If the student understands this chapter well it should be easier to look up what will change when they use .h std files. Right?

concepts/code-organization/.meta/config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "blurb": "C++ code is often organied in header and source files and managed via include directives.",
```
meatball133, line 2:
```suggestion
  "blurb": "C++ code is often organized in header and source files and managed via include directives.",
```

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,163 @@
...
+#include "myProject.h"
+example();
+```
+
+If the `myProject.h` is located in the `some_folder` and not in the root directory, you would have to use `#include "some_folder/myProject.h"`.
+
+## Include Guards
+
+It does not matter if the same file is included multiple times within a project.
+Header files should not contain definitions.
```
meatball133, line 77: It isn't described here what a definition is, I would guess it is a function definition but that is a guess.

meatball133 (commented):
Very informative concept! ✨

concepts/code-organization/introduction.md
vaeng: There are several entities that can be defined. Classes, enumeration, constants, functions. Would "not saying what can be defined" confuse you, or is that okay?

meatball133: What I have understood is this concept is very much about definition and declaration.
I mean I think definition and declaration are quite c/c++ "heavy" words.
And I think to fully understand the concept an explanation of what these words mean could be helpful, and it doesn't have to tell what you can define and what not but say more about what they mean.
It doesn't have to be long.

In the language where I come from most of the time you would do the definition and declaration at the same time. There is some expectation but in general, this rule stays true.

concepts/code-organization/introduction.md
ErikSchierboom, line 1: This file is quite big. This seems to imply that either:

- The contents are on the verbose side
- There are multiple concepts here

Ideally, the introduction should be fairly short.

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,163 @@
...
+The problem is, that pragmas are not an official part of the C++ language and the implementation differs from compiler to compiler.
+Many big projects have switched to the simpler pragma method, but a few are still cautious.
+
+## Namespaces
+
+An important method for code organization are namespaces.
+Two functions `int foo()` can be used side-by-side if they are defined in different namespaces.
+They can be called with the scope resolution operator `::`.
+Namespaces can be nested.
+They are kept as they are after includes.
```
silvanocerza, line 105: What do we mean with this sentence? It's not clear to me. 🤔

silvanocerza (commented):
Left some comments but overall it looks nice to me.

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,163 @@
...
+int also_my_result{m::foo() + o::foo() * m::my_inner_ns::baz()};
+```
+
+## Standard Library Header Files
+ 
+The standard library offers many common functions, algorithms, and data structures.
+They are grouped into header files.
+They are included with angled braces `<>` instead of double quotes `"`.
+The difference with the braces is the location, where the compiler searches for the header files.
+The search in the current project is skipped for the angled braces version, and it directly starts in the system's include directories.
```
clechasseur, line 150: I think this will feel pretty strange for most people coming from other languages. I don't really know if it's worth mentioning, but on the other hand it can be a source of error... I'm torn about this.

How about mentioning that angle brackets are used to include library files (external to the project) while double-quotes are used to include project files, with perhaps a link to an external source (like cppreference)?

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,163 @@
...
+The standard library headers are located in the `std` namespace.
+
+One example is the C numerics library, with its "cmath" header.
```
clechasseur, line 153: Very nitpicky, but using double-quotes here could be confusing since you're mentioning angle brackets before for standard library headers. Perhaps do it like this:
```suggestion
One example is the C numerics library, with its `<cmath>` header.
```

clechasseur (commented):
I think it's an important topic, but also one that is difficult to approach. Code organization in C++ projects feels alien to most people coming from other languages. On the other hand, perhaps some topics should be kept for later (or removed entirely) to simplify the content. Good start! 🎉

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,163 @@
...
+The snippet below shows how to use the `example()` function that was declared in the `myProject.h` file.
+
+```cpp
+#include "myProject.h"
+example();
+```
+
+If the `myProject.h` is located in the `some_folder` and not in the root directory, you would have to use `#include "some_folder/myProject.h"`.
+
+## Include Guards
```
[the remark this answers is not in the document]
ErikSchierboom: > This is a pretty advanced topic since it involves explaining the workings of the preprocessor, in a way. Did you plan to have a preprocessor topic at some point? If so I'd simplify this entry to avoid talking about the specifics and defer to that later topic.

Agreed!

concepts/code-organization/introduction.md
```diff
@@ -0,0 +1,163 @@
+# Introduction
+
+Big projects often span over several files to keep the code organized in logical components.
+
+## Forward Declarations
```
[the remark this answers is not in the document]
silvanocerza: I agree on this. I would talk about forward declarations even before the code organization topic.

vaeng:
This concept will be split up. See #605, #606

PR #602, opened 2023-04-26
Concept docs for "Numbers" (closes #600) 

vaeng:
See https://github.com/exercism/cpp/issues/600

PR #594, opened 2023-04-19
Adding a basic concept (closes #589)

vaeng:

PR #603, opened 2023-04-27
Concept exercise for "Numbers" (closes #599)

### LINKED ISSUE #599: Exercise for a "Numbers" concept
I would like to have an exercise that enables the user to use doubles, ints and the most common math operations and the conversions.

I think the following topics should be covered:

- Ints
- Doubles
- Limits
- *, /. +, -, ceiling from cmath?

The related concept will cover how and why. The exercise can be taken from another track. I suggest the freelancer rates exercise, as it does not need knowledge of control structures
---

- [ ] Add code files
- [ ] Add documentation files
- [ ] Update the list of implemented exercises
- [ ] Get feedback
- [ ] Implement Feedback

vaeng:
See #599

Configlet fails, has to merged after #602, #598  and #594

### DIFF
--- a/exercises/concept/freelancer-rates/.meta/config.json
+++ b/exercises/concept/freelancer-rates/.meta/config.json
@@ -13,9 +13,9 @@
       ".meta/exemplar.cpp"
     ]
   },
-  "icon": "freelancer-rates",
   "forked_from": [
     "elixir/freelancer-rates"
   ],
+  "icon": "freelancer-rates",
   "blurb": "Learn about integers and floating point numbers by helping a freelancer communicate with a project manager about billing."
 }
--- a/exercises/concept/freelancer-rates/.meta/config.json
+++ b/exercises/concept/freelancer-rates/.meta/config.json
@@ -1,0 +1,21 @@
+{
+  "authors": [
+    "vaeng"
+  ],
+  "files": {
+    "solution": [
+      "freelancer_rates.cpp"
+    ],
+    "test": [
+      "freelancer_rates_test.cpp"
+    ],
+    "exemplar": [
+      ".meta/exemplar.cpp"
+    ]
+  },
+  "icon": "lasagna",
+  "forked_from": [
+    "elixir/freelancer-rates"
+  ],
+  "blurb": "Learn about integers and floating point numbers by helping a freelancer communicate with a project manager about billing."
+}
--- a/exercises/concept/freelancer-rates/.meta/config.json
+++ b/exercises/concept/freelancer-rates/.meta/config.json
@@ -13,7 +13,7 @@
       ".meta/exemplar.cpp"
     ]
   },
-  "icon": "lasagna",
+  "icon": "freelancer-rates",
   "forked_from": [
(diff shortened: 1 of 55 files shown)


exercises/concept/freelancer-rates/.meta/config.json
```diff
@@ -0,0 +1,21 @@
...
+      "freelancer_rates.cpp"
+    ],
+    "test": [
+      "freelancer_rates_test.cpp"
+    ],
+    "exemplar": [
+      ".meta/exemplar.cpp"
+    ]
+  },
+  "icon": "lasagna",
```
clechasseur, line 16: I'm assuming this is a copy-paste of the lasagna exercise - does Exercism have a good icon for this exercise already?

exercises/concept/freelancer-rates/.meta/exemplar.cpp
```diff
@@ -0,0 +1,35 @@
+// INFO: Headers from the standard library should be inserted at the top via
+#include <cmath>
+
+// daily_rate calculates the daily rate given an hourly rate
+double daily_rate(double hourly_rate) {
+    auto per_hour{8.0};
```
clechasseur, line 6: I think it'd be better to call this variable `hours_per_day` or something similar; it's not really a per-hour value.

exercises/concept/freelancer-rates/freelancer_rates.cpp
clechasseur, line 1: Since this exercise has the code organization topic as prerequisite, I believe it'd be best to use a namespace to group the functions here. This helps drive the idea that namespaces should be used pretty much everywhere.

exercises/concept/freelancer-rates/freelancer_rates_test.cpp
```diff
@@ -0,0 +1,62 @@
+#include "freelancer_rates.cpp"
```
clechasseur, line 1: Students seeing this will most likely find it incredibly strange. Since the code organization concept is a prerequisite, why not add a header file which declares the functions?

clechasseur (commented):
I like the exercise, it's a nice addition for beginners. Just a few nitpick comments.

exercises/concept/freelancer-rates/freelancer_rates.cpp
vaeng: My plan was to give the students some options after the basics. So it can be done in parallel and does not need namespaces. So it feels less linear and they are not directly forced to do the "boring" code organization lessons.

exercises/concept/freelancer-rates/freelancer_rates_test.cpp
vaeng: The test file will look overwhelming for them even if they understand headers and namespaces. There is so much going on here. I want to ease the process as much as possible.

To be honest I felt the other way around and was really confused why everyone is using headers and nested namespaces on exercises to define two small functions. It felt almost "java-level-code-bloated" to me when I started out.

ErikSchierboom (approved):

PR #608, opened 2023-05-14
Concept docs for "Includes" (closes #607)

### LINKED ISSUE #607: Concept docs for "Includes"
The code organization concept had too many topics, so this is a strategy to remove complexity.

vaeng:
This is the second smaller package to deliver the same content as the previous "Code Organization" concept.
The wording is almost the same as the previous pr https://github.com/exercism/cpp/pull/598

In the tree, it will follow the namespace concept as the std libs use the std namespace and the namespace resolution operator.

There will be some more for headers and forward declarations.

### DIFF
--- a/concepts/includes/about.md
+++ b/concepts/includes/about.md
@@ -1,3 +1,32 @@
 # About
 
-[comment]: # (Copy content from introduction, once it is agreed upoon)
+## Includes
+
+To include the content of other files C++ uses the `include` directive.
+These includes should be stated at the top of the file.
+During compilation, the line with the directive is replaced by the content of the respective file.
+Namespaces are kept as they are in the original file after includes.
+
+## The Standard Library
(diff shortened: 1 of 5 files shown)


concepts/includes/about.md
```diff
@@ -0,0 +1,3 @@
+# About
+
+[comment]: # (Copy content from introduction, once it is agreed upoon)
```
silvanocerza, line 3:
```suggestion
[comment]: # (Copy content from introduction, once it is agreed upon)
```

silvanocerza (approved):
Just a small typo to fix and the rest is good to go. 👍

concepts/includes/about.md
silvanocerza: Ah, it was a comment. 🤦

PR #612, opened 2023-05-23
Concept docs for "Headers" (closes #611)

### LINKED ISSUE #611: Docs for "Headers" concept

vaeng:


clechasseur:
I've skimmed through it and it looks good. Will comment a bit more later.

How is it presented on the Exercism website? What's the difference between `about.md` and `introduction.md`?

I wonder if we should go into more details on how C++ files are handled by the build system, compiler and linker. It's perhaps too advanced but would help explain why the use of header files is necessary, even if it's a bit legacy...

vaeng:
> How is it presented on the Exercism website? What's the difference between `about.md` and `introduction.md`?

`about.md` is shown after you completed the concept. It's like a summary after you have solved everything.
`introduction.md` is shown on first look and also for every exercise. That uses the concept.


> I wonder if we should go into more details on how C++ files are handled by the build system, compiler and linker. It's perhaps too advanced but would help explain why the use of header files is necessary, even if it's a bit legacy...
I wondered about that as well and think we should put it in the docs of the track. It would be useful to talk about the phases of building, but we have no exercises on this and I don't know how we would achieve good results to do any.

ErikSchierboom (approved):

PR #614, opened 2023-05-25
Concept docs and exercise for "strings"

vaeng:
I want to use the existing strings exercise to combine the strings concept and the includes concept.

I added the missing information in the existing docs as a first step.

I would not yet introduce header files. So in the second step, I would rewrite the existing exercise to run without a header file.

This would be the last one to work without headers, as the headers concept would be the next in line. Afterwards, the existing training exercises can be unlocked.


concepts/strings/introduction.md
```diff
@@ -1,5 +1,42 @@
 # Introduction
 
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
+A `string` in C++ is a mutable object that represents text as a sequence of Unicode characters (letters, digits, punctuation, etc.).
```
silvanocerza, line 3: I'd expect someone that is trying to learn C++ to know what Unicode is but what if someone doesn't? Shall we maybe add in the `links.json` an explanation of what Unicode is?

vaeng: Or take it out? Because to have real Unicode compatibility we might need to talk about the mess behind the different boundaries https://stackoverflow.com/a/50407375/4919081

silvanocerza: Yeah, taking it out works too.

ErikSchierboom (approved):

PR #620, opened 2023-05-29
[Concept] Docs and Exe - If Statements

vaeng:
Adapted from the Go version of [conditionals-if](https://github.com/exercism/go/blob/main/concepts/conditionals-if).


vaeng:
Added the exercise that combines if statements and comparisons.

Thus this has to be merged after comparison docs

ErikSchierboom (approved):

PR #628, opened 2023-06-02
[Concept] Docs - Loops

vaeng:
Will be accompanied by the [interest-is-interesting](https://github.com/exercism/csharp/tree/main/exercises/concept/) exercise.

PR #629, opened 2023-06-02
[Concept] Docs - Switch

vaeng:
See #628

### DIFF
--- a/concepts/switch/about.md
+++ b/concepts/switch/about.md
@@ -1,0 +1,60 @@
+# About
+
+Like other languages, C++ also provides a `switch` statement.
+Switch statements are a shorter way to write long `if ... else if` statements.
+To make a switch, we start by using the keyword `switch` followed by an integer.
+We then declare each one of the conditions with the `case` keyword.
+We can also declare a `default` case, that will run when none of the previous `case` conditions matched.
+Each case should end with a `break` (or a `return`) statement.
+
+```cpp
+int price{0};
+int adults{3};
+int kids{2};
+
+switch (int group_size{adults + kids}) {
+    case 1:
+        price = 50;
+        break;
+    case 2:
+        price = 70;
+        break;
+    default:
+        price = group_size * 30;
+}
+```
+
+## Fall-through
+
+One important thing about the switch construct is that the code will continue to be executed until it stops at a `break` statement.
+This can lead to unexpected behavior.
+
+```cpp
+int adults{1};
+int kids{0};
+switch (int group_size{adults + kids}) {
+    case 1:
+        price = 50;
+    case 2:
+        price = 70;
+    default:
+        price = group_size * 30;
+}
+// price will be 30!
+```
+
+The main use case for this feature is a statement, that has several labels:
+```cpp
+switch (group_size) {
+case 1:
+    book_room();
+    break;
+case 2:
+case 3:
+    book_appartment();
+    break;
+default:
+    book_house(group_size);
+}
+// book_appartment happens wheng roup_size is 2 or 3
+```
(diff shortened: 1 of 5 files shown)


concepts/switch/about.md
```diff
@@ -0,0 +1,60 @@
+# About
+
+Like other languages, C++ also provides a `switch` statement.
+Switch statements are a shorter way to write long `if ... else if` statements.
+To make a switch, we start by using the keyword `switch` followed by an integer.
+We then declare each one of the conditions with the `case` keyword.
+We can also declare a `default` case, that will run when none of the previous `case` conditions matched.
+Each case should end with a `break` (or a `return`) statement.
```
BethanyG, line 8: Because this is the `about` doc (_so therefore has more detail/usecases/explaination than the intro_), I'd suggest adding a little more detail around which languages implement `case/switch`, as well as _how_ they implement it.  Not a lengthy thing .. but maybe mentioning that some languages don't have fall-through (Python), and some `match` on more that numeric values (Java, Python, Ruby, PHP), and some implement pattern matching (many functional languages) -- which is .. a bit different?

I suggest this because those coming from other programming languages may or may not be surprised by the C++ behavior.

concepts/switch/about.md
```diff
@@ -0,0 +1,60 @@
...
+        break;
+    default:
+        price = group_size * 30;
+}
+```
+
+## Fall-through
+
+One important thing about the switch construct is that the code will continue to be executed until it stops at a `break` statement.
+This can lead to unexpected behavior.
```
BethanyG, line 30: Coming from Python, I wanted more explanation here as to why this behavior happens -- and why you would or would not want it to happen.

I also confuse myself often with `fall-through` vs `default`, because it is hard for me to remember that code execution continues past a match unless explicitly ended by `break;` or `return`.

concepts/switch/about.md
```diff
@@ -0,0 +1,60 @@
...
+        price = 50;
+    case 2:
+        price = 70;
+    default:
+        price = group_size * 30;
+}
+// price will be 30!
+```
+
+The main use case for this feature is a statement, that has several labels:
```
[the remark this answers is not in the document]
BethanyG: So this _feels_ to me like it needs more explanation.  Why would a statement have several labels?  What is the use case?

concepts/switch/about.md
BethanyG: Additionally, maybe some words about why you'd use this form instead of `if-else` (_or when you'd use each??_).

concepts/switch/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+
+The main use case for this feature is a statement, that has several labels:
+```cpp
+switch (group_size) {
+case 1:
+    book_room();
+    break;
+case 2:
+case 3:
+    book_appartment();
```
BethanyG, line 54:
```suggestion
    book_apartment();
```

concepts/switch/introduction.md
```diff
@@ -0,0 +1,60 @@
...
+        price = 70;
+        break;
+    default:
+        price = group_size * 30;
+}
+```
+
+## Fall-through
+
+One important thing about the switch construct is that the code will continue to be executed until it stops at a `break` statement.
```
BethanyG, line 29:
```suggestion
One important thing about the switch construct is that the code will continue to execute until it is stopped by a `break` statement.
```

BethanyG (approved):
@vaeng  - Overall, I think this is  great start.  Most of my comments are suggestions/thoughts - so they don't have to be accepted.  I did miss having links within the text to documentation or term definition -- but that's a personal style.  Anyone who is is doubt can certainly google things.  ;-)

concepts/switch/about.md
vaeng: That's a good addition, many people will get beef with the compiler and don't get why, because the switch statement in C++17 is rather limited.

PR #631, opened 2023-06-10
WIP: perfect-numbers

sbromberger:
ping @vaeng for guidance and feedback. I ran `configlet` for the docs and for `config.json` formatting, and added a UUID and other metadata to the track-level `config.json`.

### DIFF
--- a/exercises/practice/perfect-numbers/.meta/config.json
+++ b/exercises/practice/perfect-numbers/.meta/config.json
@@ -3,20 +3,7 @@
     "sbromberger"
   ],
   "contributors": [
-    "alebaffa",
-    "bitfield",
-    "ekingery",
     "ferhatelmas",
-    "hilary",
-    "ilmanzo",
-    "kgibilterra",
-    "leenipper",
-    "petertseng",
-    "robphoenix",
-    "sebito91",
-    "Skarlso",
-    "tleen",
-    "eklatzer"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/perfect-numbers/.meta/config.json
+++ b/exercises/practice/perfect-numbers/.meta/config.json
@@ -1,0 +1,37 @@
+{
+  "authors": [
+    "sbromberger"
+  ],
+  "contributors": [
+    "alebaffa",
+    "bitfield",
+    "ekingery",
+    "ferhatelmas",
+    "hilary",
+    "ilmanzo",
+    "kgibilterra",
+    "leenipper",
+    "petertseng",
+    "robphoenix",
+    "sebito91",
+    "Skarlso",
+    "tleen",
+    "eklatzer"
+  ],
+  "files": {
+    "solution": [
+      "perfect_numbers.cpp",
+      "perfect_numbers.h"
+    ],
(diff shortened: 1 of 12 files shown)


exercises/practice/perfect-numbers/.meta/config.json
```diff
@@ -0,0 +1,37 @@
+{
+  "authors": [
+    "sbromberger"
+  ],
+  "contributors": [
```
vaeng, line 5: I'm pretty sure it is only you as an author, and you mention the version you forked from to hint that others has worked on the exercise as well.
https://exercism.org/docs/building/tracks/practice-exercises:
>- contributors: The GitHub username(s) of the exercise approach introduction's contributor(s) (optional)
>-  Including reviewers if their reviews are meaningful/actionable/actioned.

exercises/practice/perfect-numbers/.meta/example.cpp
```diff
@@ -0,0 +1,40 @@
+#include <cmath>
+#include <numeric>
```
vaeng, line 2: What's that for?

sbromberger: oops. I don't need it anymore.

Is there a tool that will warn of unnecessary `#include`s?

exercises/practice/perfect-numbers/.meta/config.json
sbromberger: Where do I put the credit for the version I started with (golang)?

exercises/practice/perfect-numbers/.meta/example.cpp
vaeng: You could use on of these: 
https://stackoverflow.com/questions/614794/detecting-superfluous-includes-in-c-c

vaeng, line 2:
```suggestion
```

exercises/practice/perfect-numbers/.meta/config.json
vaeng: I mixed up concept and practice exercises. I think it should be possible in both.

So I would suggest to keep "ferhatelmas" as a contributor, but drop the rest.

exercises/practice/perfect-numbers/.meta/example.cpp
sbromberger: Thanks!

exercises/practice/perfect-numbers/.meta/config.json
```diff
@@ -3,20 +3,7 @@
     "sbromberger"
   ],
   "contributors": [
-    "alebaffa",
-    "bitfield",
-    "ekingery",
     "ferhatelmas",
```
vaeng, line 6:
```suggestion
    "ferhatelmas"
```
I think you have to remove that , to make please json.

vaeng (changes requested):

sbromberger:
@vaeng thanks for all the guidance on this! Looking forward to creating more.

PR #633, opened 2023-06-13
[Fix] Approach formatting

vaeng:
Make configlet happy


ErikSchierboom (approved):

vaeng:
> @vaeng what command did you run to fix these files? I want to make sure it becomes part of the 'process' (even if just as a documented manual step). Thanks.

`./bin/configlet fmt -u`

PR #639, opened 2023-06-26
[Exercise] High-Scores

### LINKED ISSUE #635: Add "high-scores" exercise
For #12in23 there is one featured exercise missing to get the badge in C++: 

https://github.com/exercism/problem-specifications/tree/main/exercises/high-scores

vaeng:
This is the C++ implementation of the high scores exercise for #12in23.

Closes #635.


config.json
```diff
@@ -875,6 +875,19 @@
...
           "pointers"
         ]
+      },
+      {
+        "slug": "high-scores",
+        "name": "High Scores",
+        "uuid": "10448fb8-60eb-4c67-93e8-488ecbde8676",
+        "practices": [
+          "vector-arrays"
+        ],
```
siebenschlaefer, line 885: perhaps "class", too?

exercises/practice/high-scores/.docs/instructions.append.md
```diff
@@ -0,0 +1,11 @@
+This exercise uses the [vector container][vector].
+You might find it useful to use the [algorithm][algorithm] header to solve this exercise.
+The last tests check for [immutability][immutability].
+None of the functions that you are tasked to write should change the scores of the game.
+
+If you want some extra practice, you can implement a function that adds new scores to the game.
+
+[vector] : https://en.cppreference.com/w/cpp/header/vector
```
siebenschlaefer, line 8: I'd rather link to the class not the header: https://en.cppreference.com/w/cpp/container/vector

exercises/practice/high-scores/.meta/example.cpp
```diff
@@ -0,0 +1,32 @@
...
+int HighScores::personal_best() {
+    return *max_element(scores.begin(), scores.end());
+};
+
+vector<int> HighScores::top_three() {
+    vector<int> sorted_scores{scores};
+    sort(sorted_scores.begin(), sorted_scores.end(), greater<int>());
+    size_t limit{3};
+    if (scores.size() < limit) limit = scores.size();
+    vector<int> top3{};
```
siebenschlaefer, line 24: I know this is just an example and a proof that the exercise is solvable but in my experience some students look at this as "the official solution". Do you mind replacing the default-initialization and the loop with an initialization that uses iterators. That would be shorter and IMHO easier to read.

exercises/practice/high-scores/.meta/example.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <array>
+#include <vector>
+
+using namespace std;
```
siebenschlaefer, line 6: This is the only think that I really don't like. Please, no `using` directive in the global namespace of `.h` files, it forces other files that `#include` this to "use" that namespace themselves. Personally I'd use the prefix `std::` everywhere in the `.h` file or add a `using std::vector` declaration in the `namespace arcade` but if you really want the `using namespace` you could move it into the `namespace arcade`.

exercises/practice/high-scores/high_scores.cpp
```diff
@@ -0,0 +1,27 @@
+#include "high_scores.h"
+
+#include <algorithm>
+
+namespace arcade {
+
+    vector<int> HighScores::list_scores() {
+        // TODO: Return the current scores for this session.
```
siebenschlaefer, line 8: "the current" -> "all"

exercises/practice/high-scores/high_scores.cpp
```diff
@@ -0,0 +1,27 @@
...
+        return {0, 1, 2};
+    };
+
+    int HighScores::latest_score() {
+        // TODO: Return the latest scores for this session.
```
siebenschlaefer, line 13: "scores" -> "score" (singular)

exercises/practice/high-scores/high_scores.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <array>
+#include <vector>
+
+using namespace std;
```
siebenschlaefer, line 6: like above: please, no `using` directive in the global namespace of a `.h` file

exercises/practice/high-scores/high_scores.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace arcade {
+
+class HighScores {
+   private:
+    vector<int> scores;
+
+   public:
+    HighScores(vector<int> scores) : scores(scores){};
```
siebenschlaefer, line 15: should the implementation of the constructor be given in the initial setup?

siebenschlaefer (commented):
Looks good! There's just one thing that bothers me (`using namespace` in the global namespace of the `.h` files) and I have some small suggestions.

Also, how about constness? Should the local variables `scores`, `expected` and/or `hc` in the test  be `const`?

config.json
vaeng: I have not yet finished the syllabus to a working degree yet, so I want to postpone these entries.

exercises/practice/high-scores/.meta/example.h
vaeng: Yes, I am completely with you there.
I have not used it there yet and was unsure of that. I am glad you don't like it.

exercises/practice/high-scores/high_scores.cpp
```diff
@@ -0,0 +1,27 @@
...
+        return 0;
+    };
+
+    int HighScores::personal_best() {
+        // TODO: Return the highest scores for this session.
+        return 0;
+    };
+
+    vector<int> HighScores::top_three() {
+        // TODO: Return the top 3 scores for this session in descending order.
```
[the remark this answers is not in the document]
vaeng: The canonical data is testing for it.

exercises/practice/high-scores/high_scores.h
vaeng: The original problem spec suggested putting the exercise as "easy" so I made it easier by offering the complete header. Do you think that the "burden" of handling the header file is okay for an "easy" exercise?

exercises/practice/high-scores/.meta/example.cpp
vaeng: Yes, that is probably easier to read. My tired head had a heap in the first iteration and I tried to replace it with something easier. I would be happy to implement a suggestion if you have one ready.

vaeng:
> Also, how about constness? Should the local variables `scores`, `expected` and/or `hc` in the test be `const`?

I have not implemented scores as const, so that the student can decide to implement an "add_score" function and mentioned it in the append.md.

I'm okay with the const in the test file, but I don't know if this is any "help" for the students reading the tests.

exercises/practice/high-scores/.meta/example.cpp
siebenschlaefer: How about: `return {begin(sorted_scores), begin(sorted_scores) + limit};`

exercises/practice/high-scores/high_scores.h
siebenschlaefer: I'm fine with the definition of the class given to the students. I was just wondering about the definition of the constructor because we don't do that in any other practice exercise. I'd just *declare* the constructor in the `.h` file and leave its implementation to the students. But in the end I'm OK with either way.

BethanyG (approved):
This looks good to me.  😄

exercises/practice/high-scores/high_scores.h
```diff
@@ -0,0 +1,24 @@
...
+#include <vector>
+
+namespace arcade {
+
+class HighScores {
+   private:
+    std::vector<int> scores;
+
+   public:
+    HighScores(std::vector<int> scores) : scores(scores){};
```
siebenschlaefer, line 13: just to avoid that this question is getting lost among the others:
Do we break with the tradition and provide the implementation of the constructor?  
In all other exercises, even the ones with where we provide some declarations or some empty functions in the initial code we still leave the implementation of all functions to the students.

siebenschlaefer (approved):

exercises/practice/high-scores/high_scores.h
vaeng: My intention is "the header file does not need to be changed". I would prefer to hand over an idiomatic file that has the constructor defined, to skip it on purpose and give the impression, that it has to go into the source file.

I did not know that there was some tradition to the degree we hand over the scaffolding. May exercises are super barebones and as a student, it felt more like a chore to fill in the header file from the test file content. Especially since nothing compiles if you skip it.

If we decide to leave out the constructor definition, there will be an unused variable warning and the code will not compile, unless I fix the CMakeLists.txt accordingly. I had the plan to keep those changes to the concepts.

PR #644, opened 2023-06-29
Add C++ Functions in Concepts

leap-heng:
Hello there, I have seen that the C++ Functions concept has not yet been added to the repository yet, therefore I would like to contribute this. 
Looking forward for your suggestions.

Thank you

### DIFF
--- a/concepts/functions/.meta/config.json
+++ b/concepts/functions/.meta/config.json
@@ -1,0 +1,14 @@
+{
+    "blurb": "Make a piece of code reusable and organized with functions.",
+    "authors": [
+        "leap-heng"
+    ],
+    "contributors":
+    [
+        "leap-heng",
+        "HS-pheng",
+        "ktry3",
+        "MenglyS",
+        "uerys"
+    ]
+}
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -1,0 +1,111 @@
+# Functions
+
+Functions are a way to group code that belongs together. 
+They can be called from other functions or from the main function. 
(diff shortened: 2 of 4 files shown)


vaeng:
Hello leap-heng,

sorry for letting you hang for four days. I have not seen it or a post on the forums.

This is good work, I would be interested to know which exercise you would like to fit with the concept?
Also, we try to have examples with more than just `a` or `b` variables, so it can have some more interesting things to it.
This would be an example:
https://github.com/exercism/cpp/blob/dc7f70c165501cb33799df8063c58628bc1ea239/concepts/classes/introduction.md?plain=1#L28C4-L28C4

leap-heng:
Hello vaeng,

Thank you for reviewing my pull request, I will change the variables name in my example and make a new pull request. However, can you elaborate more with this phrase "I would be interested to know which exercise you would like to fit with the concept?". 

Thank you very much

vaeng:
You don't need to have a new issue or PR for everything, you can edit and commit to the current one. That is useful, to see the history of the files and the discussion.

Do you want to keep this one, or the other one?

The general questions for a concept are:
- Where does it fit into [the existing syllabus tree](https://exercism.org/tracks/cpp/concepts)?
- What [possible exercise](https://exercism.org/docs/building/tracks/stories) can teach all the mentioned concepts without using knowledge that has not yet been taught?

leap-heng:
Hello @vaeng, now I have commit new edits to my repository, kindly check if there is anything you would like to suggest. Thanks

PR #649, opened 2023-06-30
[Concept] Docs - Classes

vaeng:
This is the first concept on classes. There will be more which handle inheritance, destructors and other topics.


concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+
+Constructors are a big topic and have many nuances.
+If you are not explicitly defining a constructor for your class, the compiler will do the job for you.
+This has happened in the first example above.
+All variables are set to the value that was stated in the definition of the class.
+If you had not given any values, the variables would be uninitialized, which might have unintended consequences.
+
+## Structs
+
+Structs are a remainder of the C past that developed into C++.
```
BethanyG, line 89:
```suggestion
Structs are a holdover from the C past that developed into C++.
```

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+They are pretty much the same thing as classes, there is only one exception.
```
BethanyG, line 90:
```suggestion
They are effectively the same thing as `classes` with one important exception.
```

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+By default everything in a class is private. Structs are public until defined otherwise.
```
BethanyG, line 91:
```suggestion
By default, everything in a `class` is `private`.
Structs, on the other hand are `public` until defined otherwise.
```

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+
+Conventionally, the struct keyword is often used for data-only structures.
```
BethanyG, line 93:
```suggestion
Conventionally, the `struct` keyword is often used for **data-only structures**.
```

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+The class keyword is preferred for objects that have both data and functions.
```
BethanyG, line 94:
```suggestion
The `class` keyword is preferred for objects that have both data **and** function members.
```

BethanyG (approved):
Nice work!  All of these are really nits -- none of them are essential, and boil down to style/opinion.  Feel free to ignore.  There were a couple places where I think more explanation would help, but leaving it as-is would be fine as well.

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+
+Classes can have **member variables** and **member functions**.
+They are accessed by the **member selection** operator `.`.
+Just as variables outside of classes, it is advisable to initialize them with a value upon declaration.
+This value will then become the default for newly created objects from this class.
+
+## Encapsulation and Information Hiding
+
+Classes offer the option to restrict the usage of their members.
+The two basic cases are `private` and `public`.
```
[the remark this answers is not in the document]
vaeng: I was hoping the example makes it clear and I don't (yet) have to go into the nitpicking that `protected` brings, when it is introduced. You are probably right it should get a bit more explanation

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+    int damage{5};
+};
+
+Wizzard el{"Eleven"};       // deals  5 damage
+Wizzard vecna{"Vecna", 50}; // deals 50 damage
+```
+
+Constructors are a big topic and have many nuances.
+If you are not explicitly defining a constructor for your class, the compiler will do the job for you.
+This has happened in the first example above.
```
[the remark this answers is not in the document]
vaeng: Do you think that will bring additional insight? Because the compiler would insert:
```cpp
Wizard(){};
```

concepts/classes/introduction.md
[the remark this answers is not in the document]
BethanyG: It might just be my penchant for over-explanation surfacing.  😉 

I do think saving `protected` for when you have a very clear use case would be good.  I feel like `public`/`private`/`protected` often get swirled around and confused when someone is learning them all at once.

concepts/classes/introduction.md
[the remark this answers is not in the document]
BethanyG: ah!  And that is the misunderstanding here.  When I first read this, I took it to mean that if you didn't provide:

```c++
    Wizzard(std::string new_name) {
      name = new_name;
    }
```
..the compiler would fill in for you.  But the compiler will only provide an _empty_ constructor, hence the mention of setting member variable values.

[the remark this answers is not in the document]
vaeng: What do you think of this version?

concepts/classes/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+silverhand.name = "Laeral";
+// damage is private -> Compilation error
+silverhand.damage = 500; 
+```
+
+## Constructors
+
+Constructors offer the possibility to assign values to member variables when the object is created.
+They have the same name as the class and no return type.
+A class can have several constructors:
```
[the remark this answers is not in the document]
vaeng: Is this enough?

concepts/classes/introduction.md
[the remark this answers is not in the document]
vaeng: Better?

concepts/classes/introduction.md
```diff
@@ -0,0 +1,95 @@
...
+class Wizzard {
+  public:
+    int cast_spell() {
+      return damage;
+    }
+    std::string name{};
+  private:
+    int damage{5};
+};
+
```
siebenschlaefer, line 37: You could consider adding comments to this code example, explaining key lines.
E.g. something like
```
  public:   // from here on all members are accessible publicly
    inc cast_spell() {  // defines the public member function cast_spell
    ...
    std::string name{};  // defines the public member variable `name`
  private:  // from here on all members are private
    int damage{5};  // defines the private member variable `damage`
```

concepts/classes/introduction.md
```diff
@@ -0,0 +1,95 @@
...
+Wizzard el{"Eleven"};       // deals  5 damage
+Wizzard vecna{"Vecna", 50}; // deals 50 damage
+```
+
+Constructors are a big topic and have many nuances.
+If you are not explicitly defining a `constructor` for your `class`, the compiler will do the job for you.
+This has happened in the first example above.
+All variables are set to the value that was stated in the definition of the class.
+If you had not given any values in that definition, the variables would be uninitialized, which might have unintended consequences.
```
siebenschlaefer, line 89: "would" -> "might" because all types with a default constructor will be default-initialized

concepts/classes/introduction.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Structs
+
+Structs are a holdover from the C past that developed into C++.
```
siebenschlaefer, line 93: that sounds like structs are somehow discouraged. but they are even recommended by the C++ Core Guidelines for compound entities that do not have an invariant (i.e. when they just hold a couple of member variables that can be modified freely)

concepts/classes/.meta/config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "blurb": "In C++, classes are objects with their own set of characteristics and abilities. They make it easier to organize and reuse code by grouping related data and functions together in a structured and organized way.",
```
siebenschlaefer, line 2: "classes are objects" -> "classes are types"

concepts/classes/introduction.md
```diff
@@ -0,0 +1,95 @@
+# Introduction
+
+It is time to get to one of the core principles of C++: object-oriented programming (OOP).
```
siebenschlaefer, line 3: "principles" -> "paradigms" (at least that's what it's most often called in my experience)

concepts/classes/introduction.md
vaeng: Good idea! This would help indeed

concepts/classes/introduction.md
vaeng:
```suggestion
Structs came from the language's original C roots and are as old as C++ itself.
```
How does that sound to you?

concepts/classes/introduction.md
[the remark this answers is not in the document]
BethanyG: Yes!  I like the change.  😄

concepts/classes/introduction.md
```diff
@@ -0,0 +1,99 @@
...
+// name is public and can be changed:
+silverhand.name = "Laeral";
+// damage is private -> Compilation error
+silverhand.damage = 500; 
+```
+
+## Constructors
+
+Constructors offer the possibility to assign values to member variables at object creation.
+They have the same name as the `class` and do not have return type.
```
BethanyG, line 58:
```suggestion
They have the same name as the `class` and do not have a return type.
```

concepts/classes/introduction.md
```diff
@@ -0,0 +1,99 @@
...
+A class can have several constructors.
+This is useful if you are not always in need to set all variables.
```
BethanyG, line 60:
```suggestion
This is useful if you do not always have a need to set all variables.
```

concepts/classes/introduction.md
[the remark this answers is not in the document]
BethanyG: Yup. That works.  And when all else fails, you can comment.  But I do think this version reads better.

concepts/classes/introduction.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Structs
+
+Structs are a holdover from the C past that developed into C++.
+They are effectively the same thing as `classes` with one important exception.
+By default, everything in a `class` is `private`.
+Structs, on the other hand are `public` until defined otherwise.
+Conventionally, the `struct` keyword is often used for **data-only structures**.
+The `class` keyword is preferred for objects that have both data **and** function members.
+The class keyword is preferred for objects that have both data and functions.
```
BethanyG, line 99: I think either line 98 or line 99 .. but not both.

siebenschlaefer: @vaeng Oh, I completely missed these lines.  
Are you sure about them? In my experience it's completely fine for a `struct` to have member variables and member functions.

From a technical standpoint a `struct` and a `class` are almost identical, the only differences are that the implicit "default" access specifier (`public` for `struct`, `private` for `class`), and an implicit "default" access-specifier when defining a derive class (`public` inheritance when deriving from a `struct`, `private` inheritance when deriving from a `class`).

So it's "just" about conventions. The advice I heard from leading experts is that classes are for types with an invariant, something that must be true over the lifetime of an object. The constructor is responsible for establishing that invariant and it must hold at the end of each public member function. That's when you need to make member variable `private`, it allows the class to control the values and preserve the invariant. structs are for types without an invariant, where member variables can be modified from the outside. But that doesn't necessarily prevent them for having member functions.

vaeng: You are right. My "convention" is just the way I practiced it in the past.
The "use structs for pod" is surely a consequence of something without an invariant. I wonder how we can write that advice without being too technical.
I went through a lot of math courses and know people who did them with me who never became confident in finding invariants for their logical reasoning. 
Maybe I should give an example like "a wizard's life or mana cannot be negative, so these values are controlled by functions, rather than offer the possibility to set those directly". That might be a bit too wordy? Should this be moved the about.md?

vaeng: What do you think of this change? @siebenschlaefer @BethanyG

siebenschlaefer: Looks good!

PR #655, opened 2023-07-02
Sort featured July #12in23 exercises into the tree

vaeng:
Users who chose to do learning mode can see the featured exercises in the tree.

I did all five exercises with the minimum of tools available and came to the sorting seen in the config.json file.

This is what I came up with:
---

### Bob
- loops
- strings

### Allergies
- classes
- std::unordered_set
- string
- const
- reference
- unsigned

### Reverse String
- string
- const
- reference
- loops

### High Scores
- Classes
- vector
- size_t
- iterators

### Armstrong Numbers
- loops
- includes

---

Some of these concepts are not yet in the tree, so I will make another PR to add to their instructions about what challenges may lay ahead of the students.

### DIFF
--- a/config.json
+++ b/config.json
@@ -183,8 +183,12 @@
         "slug": "reverse-string",
         "name": "Reverse String",
         "uuid": "c6946af0-3c2f-4f76-8b30-a1643cd5be63",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "loops"
+        ],
+        "prerequisites": [
+          "strings"
+        ],
         "difficulty": 2,
         "topics": [
           "strings"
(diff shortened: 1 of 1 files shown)


vaeng:
I also removed some of the beta status for the exercises, so the non-maintaining user can see them.

config.json
```diff
@@ -183,8 +183,12 @@
         "slug": "reverse-string",
         "name": "Reverse String",
         "uuid": "c6946af0-3c2f-4f76-8b30-a1643cd5be63",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "loops"
+        ],
+        "prerequisites": [
```
BethanyG, line 189: Is there a concept exercise on `strings`?  If not, this exercise is going to stay locked for students.

BethanyG (approved):
I think this might still be a tad hazy in my mind.  But right now, we have three exercises that practice `loops` with a prerequisite of completing `strings`.  But  I don't see a whole lot of exercises that practice strings.

That being said, I think this would be good to get out there so folx can try the exercise tree and have these anchored to concepts.

config.json
vaeng: Yes, there is log levels (https://github.com/exercism/cpp/tree/main/exercises/concept/log-levels)
https://github.com/exercism/cpp/blob/06f46cbc72c7a6b4086e9d7f2bb054a29fe812fd/config.json#L41

BethanyG: Nevermind.  I am blind as a bat, clearly.  😆

PR #659, opened 2023-07-03
[Improve] Link practice exercises to syllabus 

vaeng:
Added some of the existing exercises to the tree.
We might have to add new exercises to make it less focused on strings and loops.


SleeplessByte (approved):

SleeplessByte:
(I assumed you just wanted the tick so it can be merged)

PR #662, opened 2023-07-03
lasagna: fix two typos

siebenschlaefer:
Two small typos:
- the name of the function remainingOvenTime starts with an uppercase R in the comment
- the heading for the section "Whitespace" has a different level (`####`) than the headings of the other sections (`###`)


vaeng (approved):
Thanks for catching those

PR #664, opened 2023-07-04
[Fix] Concept - Loops: floating point testing

vaeng:
Change tests to account for some rounding errors.

The way the testing is done currently pretty much only accepts the order of the exemplar solution and that should not hinder students from passing.


siebenschlaefer (approved):
LGTM

PR #669, opened 2023-07-07
[Concept] Docs - Enumerations

vaeng:
Will get an exercise with the switch concept in the style of https://github.com/exercism/csharp/blob/main/exercises/concept/attack-of-the-trolls/.docs/instructions.md

### DIFF
--- a/concepts/enums/.meta/config.json
+++ b/concepts/enums/.meta/config.json
@@ -1,0 +1,8 @@
+{
+  "blurb": "In C++, the array and vector types provide flexible ways to store and access multiple elements of the same data type, with arrays offering fixed-size containers while vectors offer dynamic resizing capabilities, making them suitable for various data storage and manipulation tasks.",
+  "authors": [
+    "vaeng"
+  ],
+  "contributors": [
+  ]
+}
--- a/concepts/enums/.meta/config.json
+++ b/concepts/enums/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "In C++, the array and vector types provide flexible ways to store and access multiple elements of the same data type, with arrays offering fixed-size containers while vectors offer dynamic resizing capabilities, making them suitable for various data storage and manipulation tasks.",
+  "blurb": "In C++, enumerations (enums) define custom data types with named values, ensuring variables hold specific and valid options, improving code readability and reliability.",
   "authors": [
     "vaeng"
   ],
--- a/concepts/enums/introduction.md
+++ b/concepts/enums/introduction.md
@@ -54,7 +54,7 @@ This is the reason they are called `scoped enumerations`.
 ~~~~exercism/advanced
 ## Unscoped Enumerations
 You might be thinking that with the name `scoped`, there would also be be `unscoped` enumerations -- and you would be correct.
-`Unscoped enumerations` are becoming less popular, because they all share the same global namespace.
+`Unscoped enumerations` are becoming less popular because they all share the same global namespace.
 Because of the sharing, you could not have two `unscoped enumerations` with the same `enumerators` like _plastic_ in the example above.
 
 Also, `unscoped enumerations` implicitly convert to integers. 
--- a/concepts/enums/introduction.md
+++ b/concepts/enums/introduction.md
@@ -5,11 +5,11 @@
 So far we have mostly handled numbers and strings.
 To model the real world, we might want a limited number of values that a variable can take on.
 You might want a dedicated type with a few distinct values with distinct names.
-In a skateboard factory, the deck material can be selected from maple, bamboo, or plastic.
+For example, in a skateboard factory, having the deck material be a choice of only maple, bamboo, or plastic.
 
-You could use integers to encode those values, but would have to use extra code to check if there is an invalid value coming from the system for the material.
+You could use integers to encode those values, but you would have to use extra code to check if there is an invalid value coming from the system for the material.
 The meaning of those [_magic numbers_][magic numbers] is difficult to trace over the source code and they are prone to mix-ups.
-This is a case for `enumerations`.
+`enumerations` can be used to encourage expressive code and to restrict unintentional comparison mistakes.
 
 ## Example usage
 
@@ -24,7 +24,7 @@ enum class DeckMaterial {
     plastic
 };
 ```
-Now, look at a pricing function in the skate shop and take note of the scope resolution operator (`::`) to specify an `enumerator` from the `enumeration`:
+Now, look at a pricing function in the skate shop and take note of the scope resolution operator (`::`) specifying an `enumerator` from the `enumeration`:
 
 ```cpp
 double deck_price(double base_price, DeckMaterial material) {
@@ -37,7 +37,7 @@ double deck_price(double base_price, DeckMaterial material) {
 
 ## Scoping
 
-Imagine you have a second `enumeration` for the material of the wheels:
+Imagine you have a second `enumeration` for wheel material:
 
 ```cpp
 enum class WheelMaterial {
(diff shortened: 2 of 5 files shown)


concepts/enums/.meta/config.json
```diff
@@ -0,0 +1,8 @@
+{
+  "blurb": "In C++, the array and vector types provide flexible ways to store and access multiple elements of the same data type, with arrays offering fixed-size containers while vectors offer dynamic resizing capabilities, making them suitable for various data storage and manipulation tasks.",
```
siebenschlaefer, line 2: why is this talking about arrays and vectors?

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+    lemons,  // 0
+    oranges, //1
+};
+
+enum IceCream {
+    walnut, // 0
+    apples,      // 1
+};
+
+bool comparison{apples == oranges};
```
siebenschlaefer, line 75: no criticism: I like that line!

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+// => true
+```
+
+If you want to convert `scoped enumerations` into integers you can use `static_cast<int>`.
```
siebenschlaefer, line 79: into -> to

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
+# Introduction
+
+## Enumerations
+
+So far we have mostly handled numbers and strings.
+To model the real world, we might want a limited number of options that a variable can take on.
+If you work for a skateboard manufacturer the deck material can be selected from maple, bamboo, or plastic.
+
+You could use integers to encode those values, but would have to use extra code to check if there is an invalid value coming from the system for the material.
```
siebenschlaefer, line 9: I'm not sure if enums can really prevent invalid values since integers can be converted to enums without any kind of check.  
I'd rather argue that enums allow you to use expressive symbols (`maple`) instead of magic values (`0`), and they prevent mixing different types by accident.  
But that might be semantics.

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
+# Introduction
+
+## Enumerations
+
+So far we have mostly handled numbers and strings.
+To model the real world, we might want a limited number of options that a variable can take on.
```
siebenschlaefer, line 6: (maybe) "options" -> "values"

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+If you work for a skateboard manufacturer the deck material can be selected from maple, bamboo, or plastic.
```
siebenschlaefer, line 7: can be shortened to: "In a skateboard factory the deck material ..."

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+
+You could use integers to encode those values, but would have to use extra code to check if there is an invalid value coming from the system for the material.
+This is a case for `enumerations`.
+
+## Example usage
+
+The specific term for this kind of enumeration is `scoped enumeration`.
+The deck material can be written like this.
+Note the `enum class` keyword and the `;` at the end of the definition:
+
```
siebenschlaefer, line 17: "like this." is followed by a note instead of "this" example.  
Is there a better way to phrase/structure/order the two sentences and the example?

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+```cpp
+enum class DeckMaterial {
+    maple,
+    bamboo,
+    plastic
+};
+```
+Now, look at a pricing function in the skate shop and take note of the scope resolution operator (`::`) to specify an `enumerator` from the `enumeration`:
```
siebenschlaefer, line 25: to specify -> that specifies (native English speaker, please correct me)

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+
+```cpp
+enum class WheelMaterial {
+    steel,
+    clay,
+    plastic
+};
+
+bool comparison{DeckMaterial::plastic == WheelMaterial::plastic};
+// => false
```
siebenschlaefer, line 48: false => does not compile

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+```
+Although the wheels and the deck can both be made of _plastic_, the two cannot be confused.
+Each `enumeration` will have its `enumerators` in its own scope - its own namespace.
```
siebenschlaefer, line 51: maybe put namespace in quotes

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+This is the reason they are called `scoped enumerations`.
+
+~~~~exercism/advanced
+## Unscoped Enumerations
+
+You might think that there should be `unscoped` enumerations as well and you are correct.
+`Unscoped enumerations` lost favor to the version above, because they populate the global namespace.
```
siebenschlaefer, line 58: is "to lose favor to sth." a common idiom? (@BethanyG)

concepts/enums/introduction.md
```diff
@@ -0,0 +1,80 @@
...
+
+```cpp
+enum CitrusFruits {
+    lemons,  // 0
+    oranges, //1
+};
+
+enum IceCream {
+    walnut, // 0
+    apples,      // 1
```
siebenschlaefer, line 72: align the comments

concepts/enums/introduction.md
vaeng: What do you think of this change?

concepts/enums/introduction.md
BethanyG: Yeah - _technically_ I think `to lose favor to` works(ish) - as in there is nothing wrong grammar-wise.  

But as far as usage in idiom, it's:
-  `to lose favor with` (_a person or institution_), 
- `to lose favor as` (_a method, tool, strategy, or philosophy_), 
- `to lose favor among` (_a group_),   or 
- `to lose favor`  (_vague general approval or popularity_) .

You can say `lose favor compared to` - but that can sound a bit stuffy.  I think most AmEn speakers would use `as` or `with` in that circumstance instead.

concepts/enums/introduction.md
```diff
@@ -0,0 +1,87 @@
...
+```
+Although the wheels and the deck can both be made of _plastic_, the two cannot be confused.
+Each `enumeration` will have its `enumerators` in its own scope - its own `namespace`.
+This is the reason they are called `scoped enumerations`.
+
+~~~~exercism/advanced
+## Unscoped Enumerations
+
+You might think that there should be `unscoped` enumerations as well and you are correct.
+`Unscoped enumerations` lost favor to the version above, because they populate the global namespace.
```
BethanyG, line 57:
```suggestion
`Unscoped enumerations` have lost favor over time, because they can clutter up the global namespace.
```

BethanyG: Another word instead of `clutter` could be `crowd`.  And you could add a qualification like:
`Unscoped enumerations` have lost favor over time, because they can  unnecessarily crowd the global namespace.

Clutter can also be qualified as:

 "`Unscoped enumerations` have lost favor over time, because they can create clutter in the global namespace."

concepts/enums/introduction.md
```diff
@@ -0,0 +1,87 @@
...
+For this reason you could not have two `unscoped enumerations` with the same `enumerators` like _plastic_ in the example above.
```
BethanyG, line 58:
```suggestion
Because of the shared namespace, you could not have two `unscoped enumerations` with the same `enumerators` like _plastic_ in the example above.
```

concepts/enums/introduction.md
```diff
@@ -0,0 +1,87 @@
...
+
+Also, `unscoped enumerations` implicitly convert to integers. 
+Look at the example to see a surprising result:
```
BethanyG, line 61:
```suggestion
Look at the example below for a surprising result:
```

BethanyG (approved):
Nice work here.  😄 

WOW with the unscoped enumerations.  It's like we need a special admonition that has a foot and a gun as an icon.  Or maybe a "its a trap!!" admonition.  I know everything in C++ has a reason for being in the language - but some things are so so dangerous for beginners...

concepts/enums/introduction.md
vaeng: Okay, another version?

concepts/enums/introduction.md
```diff
@@ -0,0 +1,87 @@
...
+```
+Although the wheels and the deck can both be made of _plastic_, the two cannot be confused. 
+They are different _types_:  `DeckMaterial` plastic and `WheelMaterial` plastic.
+Each `enumeration` will have its `enumerators` in its own scope - its own `namespace`.
+This is the reason they are called `scoped enumerations`.
+
+~~~~exercism/advanced
+## Unscoped Enumerations
+You might be thinking that with the name `scoped`, there would also be be `unscoped` enumerations -- and you would be correct.
+`Unscoped enumerations` are becoming less popular, because they all share the same global namespace.
```
BethanyG, line 57:
```suggestion
`Unscoped enumerations` are becoming less popular because they all share the same global namespace.
```

BethanyG (approved):
I really like this re-write!  I found one errant comma -- but beyond that, this looks really good.

PR #684, opened 2023-07-13
Improve texts

vrifox:
I just stumbled on these and wanted to fix them. Notice that English is not my native language. :)


siebenschlaefer:
The two space-related changes look find, but I'm not sure about the additional comma.  
@BethanyG Do you mind taking a look?

PR #689, opened 2023-07-17
Update instructions.md

cordair:
As it can be seen [in some posts](https://forum.exercism.org/t/my-code-not-running-on-test-cases-what-am-i-doing-wrong/4004), sometimes it's not clear what the details of our implementations should be in terms of names for the functions or where some pieces of code should be placed.
I consider that Exercism should be a placed especially suitable for beginners and non-experts, that's why I think we should try to make the guidelines to an exercise more clear as to what we're expected to do.
That's why I modified the guidelines for one of the exercises in the C++ syllabus (*raindrops*).

### DIFF
--- a/exercises/practice/raindrops/.docs/instructions.md
+++ b/exercises/practice/raindrops/.docs/instructions.md
@@ -17,4 +17,14 @@ The rules of `raindrops` are that if a given number:
 - 30 has both 3 and 5 as factors, but not 7, so the result would be "PlingPlang".
 - 34 is not factored by 3, 5, or 7, so the result would be "34".
 
+## Things to consider
+- You're expected to define a function named *convert* inside the *raindrops* namespace.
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
Hi!

We like to discuss these things on the [Forum](https://forum.exercism.org/c/programming/cpp/78) (in a new post pleas), that's why this PR has been closed automatically.

This PR modifies the the file `instructions.md`. This file gets copied verbatim from the language-independent [problem-specification](https://github.com/exercism/problem-specifications/blob/main/exercises/raindrops/description.md?plain=1).  
Any tracks-specific instructions should rather be made in `instructions.append.md`.

But my main critique is this: These are the things that students should either already know, or that research for themselves, or that they should have learned in the syllabus with its concepts and concept exercises.  
`.h` and `.cpp` files are explained in the [Includes](https://exercism.org/tracks/cpp/concepts/includes) concept, functions are explained in the [Basics](https://exercism.org/tracks/cpp/concepts/basics) concept.  
This syllabus is still evolving and growing (under the lead of @vaeng), if concepts like functions or includes are not explained well enough we should rather improve the concepts and concept exercises instead of adding small paragraphs to the instructions of the *practice exercises*.

Cheers!

PR #690, opened 2023-07-17
[Improve] String find example

### LINKED ISSUE #673: strings: simplify `find` example
The documentation for the `strings` concept contains a section on the `find` method.
It contains the following example:

```cpp
std::string new_release{"apple released a new app!"};
new_release.find("app");
// => 0
new_release.find("!");
// => 24
```

I wonder if perhaps the example string could be shortened for students to more easily see that the index in the second example is correct. I had no issue with `"app"` having index `0`, but `!` having index 24 was a bit harder to spot :)

vaeng:
It should now be easier to get the same result as the code by counting the letters in the example.

Fixes #673.

### DIFF
--- a/concepts/strings/introduction.md
+++ b/concepts/strings/introduction.md
@@ -44,12 +44,12 @@ std::string ghost = material.substr(8, 3);
 
 The `find` function is also very useful.
 It is called as a member function on the string and takes a string as the argument.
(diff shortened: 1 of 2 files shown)


siebenschlaefer (approved):
LGTM

PR #691, opened 2023-07-17
[Improve] Make fully qualified string type name explicit

### LINKED ISSUE #674: strings: make it explicit that the full type is `std::string`
The `strings` concept has the following section:

````markdown
In C++ the string type and the associated functions have to be included from the strings library before usage. You can do so by adding #include <string> to the top of your file. They will then populate the std namespace. The string literal uses the double quote character: ".

```cpp
#include <string>
std::string w_berry_quote{"A well-made sentence, I think, is a thing of beauty."};
```
````

The example makes it clear that one has to use `std::string` as the type, but maybe this could also be explicitly mentioned in the text?

vaeng:
Closes #674.

### DIFF
--- a/concepts/strings/introduction.md
+++ b/concepts/strings/introduction.md
@@ -8,6 +8,7 @@ Strings are manipulated by calling the string's methods.
 In C++ the string type and the associated functions have to be included from the strings library before usage.
 You can do so by adding  `#include <string>` to the top of your file.
 They will then populate the `std` namespace.
+The fully qualified name of the string type is `std::string`.
 The string literal uses the double quote character: `"`.
 
 ```cpp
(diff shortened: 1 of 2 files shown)


siebenschlaefer (approved):
This is definitely an improvement, I'll approve it.

Technically a name is "fully qualified" if it starts with `::` (for example `::std::string`), see \[contents\] in the [C++17](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf) and [C++20](https://isocpp.org/files/papers/N4860.pdf) standards, apparently removed in the [C++23](https://open-std.org/jtc1/sc22/wg21/docs/papers/2023/n4950.pdf) standard.

But that might be too much of a technicality for this syllabus and I don't know any better name.

vaeng:
> This is definitely an improvement, I'll approve it.
> 
> Technically a name is "fully qualified" if it starts with `::` (for example `::std::string`)
> But that might be too much of a technicality for this syllabus and I don't know any better name.

I was hesitant to write that, but I think it is the closest I can get to a correct term. If we find something better, I'm happy to change it again.

PR #692, opened 2023-07-17
[Improve] VehiclePurchase - Add paragraph about format

### LINKED ISSUE #676: vehicle-purchase: be explicit about what is to be returned in task 2
The second task ("Choose between two potential vehicles to buy") doesn't completely explain to the student what they need to do, part of it is hidden in the example.
I'm referring to the structure of the string to be returned, which is: "<CHOSEN_VEHICLE> is clearly the better choice."
It might be worth making this explicit in the task.

vaeng:
Closes #676.


siebenschlaefer (approved):
LGTM

PR #697, opened 2023-07-26
Update introduction.md

pdmoore:
Fix typos in the Assignment Operators section.

### DIFF
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -72,11 +72,11 @@ people += 2;
 eggs += 3 * people;
 // eggs is now 6
 ```
-Vairables `people` & `eggs` are initialized to `0`.
-Then, we add integer value `2` over the existing value `0` of the variable `people` and assigne it back to `people`.
+Variables `people` & `eggs` are initialized to `0`.
+Then, we add integer value `2` over the existing value `0` of the variable `people` and assign it back to `people`.
 `people` becomes `2` now.
 Later, we add `3` eggs for each person, which turns out to be `6` eggs in total.
-Now add this `6` to existing value `0` of the variable `eggs` and assigne it back to `eggs`.
+Now add this `6` to existing value `0` of the variable `eggs` and assign it back to `eggs`.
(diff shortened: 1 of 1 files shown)


siebenschlaefer (approved):
LGTM.

I think this whole text sounds and looks a little bit odd ("add a value *over* the existing value", "assigns a variable *with* a literal", "operators modifies", inconsistent commas) and could use some care from a native English speaker.

pdmoore:
Thanks for the quick turnaround on this!
Agreed on the revamp. I wanted to separate the simple fix from a larger change. I think there should be mention of operations in parentheses taking precedence over left-to-right order. The solution also seems to require std:ceil but doesn't cover floor vs ceiling rounding.
This exercise also explains using an apostrophe to separate digits which would have been better placed in an earlier exercise in the syllabus, Last Will, which uses that notation without any mention of it.

vaeng (approved):
Thanks for the corrections! 🥳

vaeng:
> Thanks for the quick turnaround on this!

Thanks for opening a PR!


> Agreed on the revamp. I wanted to separate the simple fix from a larger change. I think there should be mention of operations in parentheses taking precedence over left-to-right order. The solution also seems to require std:ceil but doesn't cover floor vs ceiling rounding.

I merged the fixes for the typo, there might be other places that have the same texts, I will try to correct those as well. Generally there is the introduction from the concept, which is reused in the about and again in each exercise that uses the concept. Three times the work. I would be happy to automate that bit with templates.

> This exercise also explains using an apostrophe to separate digits which would have been better placed in an earlier exercise in the syllabus, Last Will, which uses that notation without any mention of it.

Good point. I was sure I had those in the right order, I have to check the config.json if there was a mix-up somewhere.

pdmoore:
> I merged the fixes for the typo, there might be other places that have the same texts, I will try to correct those as well.

I wasn't aware of the duplication - will keep it in mind for the future.

For this particular set of typos they also need fixing in:
exercises/concept/freelancer-rates/.docs/introduction.md
./concepts/numbers/about.md

PR #699, opened 2023-08-03
[Concept] Docs -  Functions

### LINKED ISSUE #660: Add Function in C++ Concepts
Hello there, I have not yet seen the function in C++ concepts been added to the repository yet. However these are my initial points to be added.

- Function description
- Function declaration
- Function definition
- Function call
- void Functions
- Function overloading
- Function template
- and Default arguments

Thank you and more suggestions are welcome :)

vaeng:
This concept is called functions, although the students are using functions from the first concept exercise.
I hope it is a good place, for them, as they introduce quite a lot of noise, that is needed for later concepts, but would not help in the early stages of the tree.

I would place these in the syllabus after references to make it clear, that a reference is also a change in the function type signature.

As an exercise, I am thinking to implement an advanced version of the lasagna cooking class from the first concept. This time it will be some bigger operation with recipes that cannot be changed and stuff that plays with overloads and defaults.

Closes #660.

### DIFF
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -1,3 +1,124 @@
 # About
 
-[comment]: (Please fill with introduction.md content. For real.)
+## Terminology
+
+You have already used and written C++ functions in a couple of concepts.
+It is time to get technical.
+The code snippet below shows the most common terms for easy reference.
+As C++ ignores whitespace, the formatting has been changed to put each element on a single line.
+
+```cpp
+// Function declaration:
+bool                                              // Return type
+admin_detected(string user, string password)      // Type signature
+;                                                 // Don't forget the ';' for the declaration
+
+// Function definition:
+bool                                              // Return type
+admin_detected                                    // Function name
+(string user, string password)                    // Parameter list
+{ return user == "admin" && password == "1234"; } // Function body
+```
+~~~~exercism/advanced
+The declaration works like a note to the compiler, that there is a function of that name, return type and parameter list.
+The code will not work, if the definition is missing.
+Declarations are optional, they are needed if you use the function before its definiton.
+Declarations can solve problems like cyclic references and they can be used to separate the interface from the implementation.
+~~~~
+
+## The const qualifier
+
+Sometimes you want to make sure that values cannot be changed after they have been initialized.
+C++ uses the `const` keyword as a qualifier for constants.
+
+```cpp
+const int number_of_dragon_balls{7};
+number_of_dragon_balls--; // compilation error
+```
+
+~~~~exercism/note
+You will often see constants written in _UPPER_SNAKE_CASE_.
+It is recommened to reserve this casing for macros, if there is no other convention.
+~~~~
+
+If you try to change a constant variable after it has been set, your code will not compile.
+This helps to avoid unintended changes but also opens up optimization possibilities for the compiler.
+As a human, it is also easier to reason about the code, if you know that certain parts will not be affected.
+
+You can also have `const` as a qualifier for function parameters.
+
+```cpp
+string guess_number(const int& secret, const int& guess) {
+    if (secret < guess) return "lower.";
+    if (secret > guess) return "higher.";
+    return "exact!";
+}
+```
+
+When you pass a `const` reference to the function, you can be sure that it will be left unchanged.
+You will often see `const` references for objects that might be costly to copy, like longer strings.
+A third use-case for the `const` qualifier are member functions that do not change the instance of a class.
+
+```cpp
+class Stubborn {
+    public:
+    Stubborn(string reply) {
+        response = reply;
+    }
+    string answer(const string& question) const {
+        if (question.lenghth() == 0) { return "";}
+        return response;
+    }
+    private:
+    string response{};
+};
+```
+
+The `Stubborn` member function `answer` uses a `const string&` reference as a parameter.
+This avoids a copy operation from the original object that was passed to the function.
+
+## Function overloading
+
+Multiple functions can have the same name if the parameter list is different.
+That is called function overloading and it is usually done if these functions perform very similar tasks.
+
+The function header without the return type is the function's __type signature__.
+A change in the type signature results in a new function.
+
+The `play_sound` example has six different overloads to accommodate different scenarios:
+
+```cpp
+// different argument types:
+void play_sound(char note);         // C, D, E, ..., B
+void play_sound(string solfege);    // do, re, mi, ..., ti
+void play_sound(int jianpu);        // 1, 2, 3, ..., 7
+
+// different number of arguments:
+void play_sound(string solfege, double duration);
+
+// different qualifiers:
+void play_sound(vector<string>& solfege);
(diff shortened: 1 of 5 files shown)


concepts/functions/about.md
vaeng, line 1: This needs to be fixed before the merge.

siebenschlaefer (commented):
Looks good overall.

But do we really need `const` in this concept?  
For me `const` is part of the type, leaving it out would make the section about function overloading simpler.

The section about function overloading sounds a little bit technical.  
If that's what you're aiming for that's fine.  
I'd prefer starting with the motivation, something like
"Multiple functions can have the same name if the number or types of the parameters are different. That is usually done if these functions perform very similar tasks just with different parameters or types."  
But the decision is yours, I believe the syllabus is better if there's one consistent voice.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,115 @@
...
+
+// Function definition:
+bool                                              // Return type
+admin_detected                                    // Function name
+(string user, string password)                    // Parameter list
+{ return user == "admin" && password == "1234"; } // Function body
+```
+~~~~exercism/advanced
+The declaration works like a note to the compiler, that there is a function of that name, return type and paramter list.
+The code will not compile, if the definition is missing.
```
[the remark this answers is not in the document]
vaeng: Is "work" too general?

concepts/functions/introduction.md
```diff
@@ -0,0 +1,115 @@
...
+```cpp
+string guess_number(const int& secret, const int& guess) {
+    if (secret < guess) return "lower.";
+    if (secret > guess) return "higher.";
+    return "exact!";
+}
+```
+
+When you pass a `const` reference to the function, you can be sure that it will be left unchanged.
+You will often see `const` references for objects that might be costly to copy, like very long strings.
```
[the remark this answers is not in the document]
vaeng: Huh, I had no idea a good threshold is so low!

concepts/functions/introduction.md
```diff
@@ -0,0 +1,115 @@
...
+    private:
+    string response{};
+};
+```
+
+## Function Overloading
+
+If you take the function header without the return type, you get a function's __type signature__.
+The type signature is defined by the function's name, the number of parameters, their types, and their qualifiers.
+If you change anything about the signature, the compiler will search for a fitting function.
```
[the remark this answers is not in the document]
vaeng: What do you think of the wording: "A change in the signature results in a new function."?

concepts/functions/introduction.md
```diff
@@ -0,0 +1,115 @@
...
+## Default arguments
+
+Some functions can get very lengthy and many of its calls might use the same values for most of the parameters.
+The repetition in those calls can be avoided with default arguments.
+
+```cpp
+void record_new_horse_birth(string name, double weight, string color="brown-ish", string dam="Alruccaba", string sire="Poseidon");
+
+record_new_horse_birth("Urban Sea", 130); // color will be brown, dam Alruccabam, sire Poseidon
+record_new_horse_birth("Highclere", 175, "off-white", "Fall Aspen");   // sire will be Poseidon
```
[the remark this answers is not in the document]
vaeng: You are right about the int to double conversion, that was an oversight in a refactor.
I am not sure about a good way to hint at the string conversion without making it too complicated or introducing `""s`. I think there should be some concept about string_view and `""sv` at some point.

vaeng:
> But do we really need `const` in this concept? For me `const` is part of the type, leaving it out would make the section about function overloading simpler.

I have seen some examples where the const conversion was not clear in the code and I thought it might as well come in here. It does not feel enough for its own concept, but too much/late to go into a constexpr concept?

> The section about function overloading sounds a little bit technical. If that's what you're aiming for that's fine. I'd prefer starting with the motivation, something like "Multiple functions can have the same name if the number or types of the parameters are different. That is usually done if these functions perform very similar tasks just with different parameters or types." But the decision is yours, I believe the syllabus is better if there's one consistent voice.

I should look at that, I want the whole thing to feel more fun and "aha, interesting" than a manual text.

vaeng:
@siebenschlaefer 
Thank you for the feedback. I think it is much better now.
Would you be so kind and review the changes?

concepts/functions/introduction.md
```diff
@@ -0,0 +1,119 @@
...
+## Terminology
+
+You have already used and written C++ functions in a couple of concepts.
+It is time to get technical.
+The code snippet below shows the most common terms for easy reference.
+As C++ ignores whitespace, the formatting has been changed to put each element on a single line.
+
+```cpp
+// Function declaration:
+bool admin_detected(string user, string password) // Function header
```
siebenschlaefer, line 12: I missed this in my previous review: Is "Function header" an established term?

concepts/functions/introduction.md
```diff
@@ -0,0 +1,119 @@
...
+    Stubborn(string reply) {
+        response = reply;
+    }
+    string answer() const {
+        return response;
+    }
+    private:
+    string response{};
+};
+```
```
siebenschlaefer, line 73: another thing I missed: Before the snippet you explain that `const` references are often used for objects that are expensive to copy. You could consider showing that in the snippet.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,119 @@
...
+The function header without the return type is the function's __type signature__.
+A change in the type signature results in a new function.
+
+The `play_sound` example has six different overloads to accommodate different scenarios:
+
+```cpp
+// different argument types:
+void play_sound(char note);         // C, D, E, ..., B
+void play_sound(string solfege);    // do, re, mi, ..., ti
+void play_sound(int jianpu);        // 1, 2, 3, ..., 7
```
siebenschlaefer, line 89: not a critique: I just learned two new words from you: solfege and jianpu. Thanks!

siebenschlaefer (commented):
Looks good!

siebenschlaefer (approved):

concepts/functions/introduction.md
vaeng: 🤯 ?!
I was so sure it is! And yet extensive search did not result in any findings. Maybe that came from some other language and I always assumed it is the same in Cpp. It seems so consequential to have a body and a header.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,124 @@
...
+A third use-case for the `const` qualifier are member functions that do not change the instance of a class.
+
+```cpp
+class Stubborn {
+    public:
+    Stubborn(string reply) {
+        response = reply;
+    }
+    string answer(const string& question) const {
+        if (question.lenghth() == 0) { return "";}
```
siebenschlaefer, line 69: found a typo! :-)

siebenschlaefer (commented):
last minute review of last minute changes

concepts/functions/introduction.md
vaeng: Oh, no! I will put that into the exercise for that concept

PR #700, opened 2023-08-06
Strengthen tests in Vehicle Purchase exercise

pdmoore:
Current tests for choose_vehicle allow a simplistic implementation of
return option1 + " is clearly the better choice.";
to pass all tests without comparing option1 and option2.

This is due to the current tests all passing in option1 and option2 sorted in lexicographic order already.

This change swaps option1 and option2 for a couple of the tests which forces the implementation to compare the parameters.

This change will not impact existing solutions that are correctly implemented.


siebenschlaefer (approved):
Looks good to me. Thanks!

vaeng (approved):

PR #709, opened 2023-08-29
Update making_the_grade.cpp - fixed wrong TODO comment

apprentiLAB-Suzanne:
In function letter_grades, comment was about previous function, above_threshold


siebenschlaefer (approved):
Good catch.

PR #710, opened 2023-09-03
Update hints.md

Prin-S:
.pop_back() does NOT return the last item in a vector


vaeng:
Thank you for the PR.

vaeng (approved):

PR #712, opened 2023-09-07
Update tests.md with example

apprentiLAB-Suzanne:
Added example about where to find the line to move past the next test.

### DIFF
--- a/exercises/shared/.docs/tests.md
+++ b/exercises/shared/.docs/tests.md
@@ -9,13 +9,18 @@ When you start a new exercise locally, the files are configured so that only the
 Get that first test compiling, linking and passing by following the [three rules of test-driven development][three-laws-of-tdd].
 Create just enough structure by declaring namespaces, functions, classes, etc., to satisfy any compiler errors and get the test to fail.
 Then write just enough code to get the test to pass.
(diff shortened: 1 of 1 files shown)


vaeng:
Hello @apprentiLAB-Suzanne,

thanks for your PR. It is a good thing to show an actual file. The bob exercise might be unknown, so I would show it and not make the person search for the source code of the bob test. Probably something like:
> To include the next test you have to move the line with `if defined(EXERCISM_RUN_ALL_TESTS)` past the desired test.
> See the example below from the Bob exercise:
```diff
-#if defined(EXERCISM_RUN_ALL_TESTS)
TEST_CASE("shouting")
{
    REQUIRE("Whoa, chill out!" == bob::hey("WATCH OUT!"));
}
+#if defined(EXERCISM_RUN_ALL_TESTS)
```

apprentiLAB-Suzanne:
Hi @vaeng, 
thanks for the suggestion, I updated the PR !
I kept "next test", instead of "desired test". I find that it might be simpler for beginners, if you prefer otherwise I can change it 🙂 
Cheers

vaeng (approved):
LGTM!

PR #717, opened 2023-09-27
Clarify instructions for anagram exercise

shrvtv:
One of the examples in current instructions for anagram exercise (StoP is not an anagram of sTOp) is counter-intuitive and confusing without explanation. This commit clarifies said example by explaining why exactly it is not an anagram.


apprentiLAB-Suzanne:
Perhaps you could be even more precise : you added "because it's the same word", so something like "because the order of the letters do not change" ?

siebenschlaefer:
The description is taken from the language-agnostic [problem-specification](https://github.com/exercism/problem-specifications/blob/main/exercises/anagram/description.md?plain=1).  
If you want to suggest any modifications please discuss them in the [forum](https://forum.exercism.org/c/exercism/bugs-and-features/126) first and get a consensus before creating a PR on GitHub.  
Cheers!

PR #718, opened 2023-09-28
build: activate representer

vaeng:
**Do not merge before** preparation scripts have finished. This has to be merged after the okay from Eric.

(Rep is high, because I forgot to set the label in the representer)


ErikSchierboom:
Merged!

PR #720, opened 2023-10-07
`int idx` --> `size_t idx` to prevent integer overflow

tuan-karma:
```
for(int idx = number * number; idx <= limit; idx +=number)
	composite[idx] = true;
```
This code can be integer overflow if the limit > sqrt(maxInt). 
--> Then it leads to wrongly admit that idx <= limit. 
--> Then it leads to an out of bounds access at `composite[idx] = true` then it leads to **Segmentation Fault** (for example, when you set limit = 47000).

### DIFF
--- a/exercises/practice/sieve/.approaches/introduction.md
+++ b/exercises/practice/sieve/.approaches/introduction.md
@@ -31,10 +31,10 @@ namespace sieve {
 		std::vector<int> primes;
 		primes.reserve(limit / 2);
     
-		for (int number = 2; number <= limit; number++) {
+		for (long long number = 2; number <= limit; number++) {
 			if (!composite[number]) {
 				primes.emplace_back(number);
-				for(std::size_t idx = number * number; idx <= limit; idx +=number)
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
Yes, there's a potential integer overflow if the `limit` is too big. Good catch!

But the suggested change doesn't fix that, the underlying type of the operation `number * number` is still `int`, you will still get an integer overflow if the `limit` is too big, *before* the conversion to `std::size_t`.  
Also, I would recommend not to mix signed and unsigned types in the same expression (see [C++CG](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-mix)).

How about defining `number` and `idx` as `long long` or `std::int64_t`? That should solve the problem, at least on platforms where an `int` has fewer than than 64 bits. What do you think?

vaeng:
4 months have passed without a reaction from the original poster. I am thus, merging it with SIebenschläfers suggestions.  I hope that is okay. Tuan should still get reputation points on Exercism.

vaeng (approved):

PR #724, opened 2023-10-14
Fix(making the grade): correct "witch example"

### LINKED ISSUE #721: making-the-grades: A typo in example
In introduction, there is an example of vector modifiers.
The container here should be vector instead of array.

`std::array<std::string, 3> witches {"Holly", "Alyssa", "Shannen"};
witches.pop_back();
// Shannen is no longer with the witches
witches.emplace_back("Rose");
// Rose has joined the team`

vaeng:
fixes #721


siebenschlaefer (approved):
LGTM.

PR #727, opened 2023-10-17
feat: add class info for headers concept

vaeng:
I have a good use case and idea to bring the header exercise into the track. I want to place the concept after classes because they might add a lot of questions if they stay unaddressed. The concept would need some updates to help the students along.

### DIFF
--- a/concepts/headers/introduction.md
+++ b/concepts/headers/introduction.md
@@ -51,7 +51,7 @@ The split between those two might seem arbitrary and the following example can g
 
 ```cpp
 // A file named robot_flower.h
-#if !defined(ROBOT_FLOWER_H)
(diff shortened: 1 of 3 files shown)


concepts/headers/introduction.md
```diff
@@ -13,33 +13,79 @@ Some projects use `.hpp` or skip the extension completely.
...
+// A file named quick_math.h
 #pragma once
-namespace magic {
-    int my_func(int a, int b);
+namespace quick_math {
+    double super_root(double x, int n);
 }
 ```
 
 ```cpp
```
vaeng, line 26: These updates need to be propagated to the `about.md`  files when finished.

ErikSchierboom (approved):

PR #731, opened 2023-10-19
docs: improve wording for header concept

vaeng:
Reduces the "advice" character of the class section and gives two examples instead.

### DIFF
--- a/concepts/headers/about.md
+++ b/concepts/headers/about.md
@@ -104,7 +104,7 @@ namespace robots {
 }
 ```
 
-Projects might use layouts that are combinations of those layouts and there is a lot of discussion what might be the best fit for each use case.
+Projects might use combinations of those layouts and there is a lot of discussion what might be the best fit for each use case.
 
 ## Include Guards via pragma once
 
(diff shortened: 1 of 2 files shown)


concepts/headers/about.md
```diff
@@ -76,12 +74,37 @@ namespace robots {
...
+            void start_next_day() {if (!needs_water) ++size; needs_water = true;}
+    };
+}
 ```
 
-Member variables are usually kept in the header together with the implementation of trivial member functions.
-The simple `give_water`, `get_name` and `get_size` functions are so small, that one would not move them into the `.cpp` file.
-The constructor and the `start_next_day` functions shall be considered more _complex_ in this example and have thus been moved.
-The definitions are prefixed with the namespace `robots` and the class type `Flower`.
+Projects might use layouts that are combinations of those layouts and there is a lot of discussion what might be the best fit for each use case.
```
siebenschlaefer, line 107: might use layouts that are combinations -> might use combinations

siebenschlaefer (commented):
I like it

siebenschlaefer (approved):

siebenschlaefer (approved):

PR #733, opened 2023-10-31
high_scores.h: #include <array>

EQt:
No need to `#include <array>`

[no important files changed]


vaeng:
Thanks for the catch. Would you please include https://github.com/exercism/cpp/blob/main/exercises/practice/high-scores/.meta/example.h as well?

vaeng (approved):
LGTM

vaeng:
We should be careful to add `[no important files changed]` in the PR, so that the tests are not re-run

PR #747, opened 2023-12-01
Fix string literal comparison errors in bob_test.cpp

mgrubb:
When comparing against string literals with `==` operator compiler throws `/Users/mgrubb/Exercism/cpp/bob/bob_test.cpp:15:25: error: result of comparison against a string literal is unspecified (use an explicit string comparison function instead) [-Werror,-Wstring-compare]
    REQUIRE("Whatever." == bob::hey("Tom-ay-to, tom-aaaah-to."));`

This prevents the tests from passing.
This change creates std::string constants to compare against.

### DIFF
--- a/exercises/practice/bob/bob_test.cpp
+++ b/exercises/practice/bob/bob_test.cpp
@@ -4,132 +4,139 @@
 #else
 #include "test/catch.hpp"
 #endif
+#include <string>
 
 // Bob exercise test case data version 1.4.0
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
Please discuss contributions like this in the [forum](https://forum.exercism.org/c/programming/cpp/78) first.

May I ask: What is the return type of `hey` in your function? It's probably not `std::string` nor `std::string_view`.

PR #749, opened 2023-12-02
docs: small fixes for doctor-data instructions

vaeng:
Suggestions by @clechasseur

### DIFF
--- a/exercises/concept/doctor-data/.docs/instructions.md
+++ b/exercises/concept/doctor-data/.docs/instructions.md
@@ -31,7 +31,7 @@ The constructor can apparently be called with two or three arguments.
 The first argument seems to be of type `string`, the second one is a number.
 It is possible to initialize the `Vessel` class with a third argument.
 The third argument comes from a `star_map` namespace.
-It is an enumerator of type `System`.
+It is an `enum` of type `System`.
 You even got one of the enumerations: `BetaHydri`.
 
 Prepare the source and header files with your discovered information.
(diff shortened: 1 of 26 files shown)


clechasseur (approved):

siebenschlaefer (approved):
If you guys really want this I'll approve. 
But I like the previous version more. The "elements" of an `enum` are called enumerators, `enum` is the type itself.

vaeng:
@siebenschlaefer
Here is @clechasseur original reasoning:
> In the first task, the text mentions an "enumerator of type System". I think the use of the term "enum" (or perhaps "enumeration") would be better, because "enumerator" is used in many languages to mean something quite different from a C++ enum.

I might go the middle way by inserting a short "reminder" about the naming convention of enums in C++ and reverting it to the original state. What do you think?

siebenschlaefer:
I don't object to this change. I understand that these concepts don't target language lawyers and that it might sense to use terms that are easier to understand than what is used in the standard.
I just wanted to give the C++ perspective where "enumerator" is the correct term (see [[dcl.enum]](https://eel.is/c++draft/enum) in the current draft of the standard.

clechasseur (changes requested):
@siebenschlaefer After a careful reading of your comments and the provided link, I believe that the following terms should be used

- `enumeration`: refers to the type (in this case, `System`)
- `enumerator`: refers to an enum variant (in this case, `BetaHydri`)

I would therefore propose the following rephrasing, which more closely matches the standard (emphasis mine):

> The third argument comes from a `star_map` namespace.
> It is an **enumeration type** named `System`.
> You even got one of the **enumerators**: `BetaHydri`.

As an aside, I didn't know that the term _enumerator_ was used in the C++ standard to refer to an enum variant. Because of that, I thought using _enumerator_ was misleading to people coming from other languages. I do in fact like to use proper C++ terminology when possible. 🙂

vaeng:
Also add clarification to the garbled data: https://forum.exercism.org/t/corruption-in-the-c-track-repo/9088/6

PR #750, opened 2023-12-06
Fixed unnecessary withespace

adamarous:


vaeng (approved):
Thank you for your sharp eye!

PR #753, opened 2023-12-26
Update introduction.md

ShuhaoZQGG:
Improve readability of robot_flower.cpp by enhancing formatting.


vaeng:
Hello @ShuhaoZQGG,
thanks for the contribution. I think you are correct that the readability is better in normal code, but I would rather have the examples short.

PR #759, opened 2024-01-07
feat(leap): add approaches

### LINKED ISSUE #758: feat: add approaches for leap exercise
See https://forum.exercism.org/t/48in24-exercise-01-16-leap/8964

vaeng:
closes #758

@BethanyG 
I would appreciate some language nitpicks :)

### DIFF
--- a/exercises/practice/leap/.approaches/boolean-chain/content.md
+++ b/exercises/practice/leap/.approaches/boolean-chain/content.md
@@ -1,0 +1,94 @@
+# Chain of boolean expressions
+
+```cpp
+bool is_leap_year(int year) {
+    return (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0);
+}
+```
+
+The first boolean expression uses the [remainder operator][remainder-operator] to check if the year is evenly divided by `4`.
+
+- If the year is not evenly divisible by `4`, then the chain will "short circuit" due to the next operator being a [logical AND][logical-and] (`&&`),
+  and will return `false`.
+- If the year _is_ evenly divisible by `4`, then the [logical NOT operator][logical-not] is used to check if the year is _not_ evenly divisible by `100`.
+- If the year is not evenly divisible by `100`, then the expression is `true` and the chain will "short-circuit" to return `true`,
+  since the next operator is a [logical OR][logical-or] (`||`).
+- If the year _is_ evenly divisible by `100`, then the expression is `false`, and the returned value from the chain will be if the year is evenly divisible by `400`.
+
+| year | year % 4 == 0 | year % 100 != 0 | year % 400 == 0 | is leap year |
+| ---- | ------------- | --------------- | --------------- | ------------ |
+| 2020 | true          | true            | not evaluated   | true         |
+| 2019 | false         | not evaluated   | not evaluated   | false        |
+| 2000 | true          | false           | true            | true         |
+| 1900 | true          | false           | false           | false        |
+
+The chain of boolean expressions is efficient, as it proceeds from testing the most likely to least likely conditions.
+
+## Shortening
+
+By using the [implicit conversion][implicit-conversion] from `integer` to `bool` the code can be shortened using the [logical NOT operator][logical-not].
+Only `0` is converted to `false` and _every non-zero_ value is converted to `true`.
+
+```cpp
+bool is_leap_year(int year) {
+    return !(year % 4) && ((year % 100) || !(year % 400 == 0));
+}
+```
+
+It can be thought of as the expression _not_ having a remainder.
+
+## Precedence
+
+You may have noticed the braces around the boolean expressions above.
+Consider the following code:
+
+```cpp
+bool many_braces    = ((year % 100) != 0) || ((year % 400) == 0);
+bool some_braces    = (year % 100 != 0) || (year % 400 == 0);
+bool no_braces      =  year % 100 != 0 || year % 400 == 0;
+```
+
+All lines produce the same result, but differ a lot in readability.
+The reason for their equal results, is the order of operator evaluation (`%`, `!=`, `||`, etc) or [precedence][operator-precedence].
+In this example the first operator to get the attention of the compiler is the remainder, followed by the equality operators.
+Only then will the compiler go for `&&` and then check `||`.
+Some people might overlook that `and` is evaluated before `or` and thus introduce bugs.
+This is a potential reason to keep some braces, although they are not necessarily needed in the code.
+
+```cpp
+true || false && true;
+\\ => false
+(true || false) && true;
+\\ => true
+```
+
+## Short-circuit evaluation
+
+If a boolean expression's value is not going to change with further evaluations, C++ code will [short-circuit][hort-circuit-evaluation] and skip these "unesssary" evaluations.
+They might not change the final result of the boolean, but side-effects will not take place.
+
+```cpp
+bool print_and_always_false() {
+    std::cout << "Hello!";
+    return false;
+}
+
+bool print_and_always_true() {
+    std::cout << "World!";
+    return true;
+}
+
+print_and_always_false() && print_and_always_true();
+// will only print "Hello!". `print_and_always_true`` is never called.
+```
+
+Short-circuiting is a way to speed up your programs by sorting your chain of booleans in a sensible order.
+One strategy might be to evaluate expensive operations late or very likely and unambiguous results first.
+
+[remainder-operator]: https://en.cppreference.com/w/cpp/language/operator_arithmetic
+[logical-not]: https://en.cppreference.com/w/cpp/keyword/not
+[logical-and]: https://en.cppreference.com/w/cpp/keyword/and
+[logical-or]: https://en.cppreference.com/w/cpp/keyword/or
+[implicit-conversion]: https://en.cppreference.com/w/cpp/language/implicit_conversion
+[operator-precedence]: https://en.cppreference.com/w/cpp/language/operator_precedence
+[hort-circuit-evaluation]: https://en.wikipedia.org/wiki/Short-circuit_evaluation
--- a/exercises/practice/leap/.approaches/introduction.md
+++ b/exercises/practice/leap/.approaches/introduction.md
@@ -1,0 +1,72 @@
+# Introduction
+
+There are several idiomatic approaches to solve Leap.
+You can use a chain of boolean expressions to test the conditions.
+Or you can use a [ternary operator][ternary-operator].
+You might also see functions from the `std::chrono` or `boost::gregorian` libraries.
+
+## General guidance
+
+The key to solving Leap is to know if the year is evenly divisible by `4`, `100` and `400`.
+For determining that, you will use the [remainder operator][remainder-operator].
+
+## Approach: Chain of Boolean expressions
+
+```cpp
+bool is_leap_year(int year) {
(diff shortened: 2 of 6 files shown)


exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
+# Chain of boolean expressions
+
+```cpp
+bool is_leap_year(int year) {
+    return (year % 4 == 0) && (year % 100 != 0 || year % 400 == 0);
```
siebenschlaefer, line 5: very tiny nitpick: You could either omit the first pair of parentheses:
`return year % 4 == 0 && (year % 100 != 0 || year % 400 == 0);`
or add parentheses to the second and third sub-expression, like in the "shortened" example below:
`(year % 4 == 0) && ((year % 100 != 0) || (year % 400 == 0));`

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+- If the year _is_ evenly divisible by `100`, then the expression is `false`, and the returned value from the chain will be if the year is evenly divisible by `400`.
+
+| year | year % 4 == 0 | year % 100 != 0 | year % 400 == 0 | is leap year |
+| ---- | ------------- | --------------- | --------------- | ------------ |
+| 2020 | true          | true            | not evaluated   | true         |
+| 2019 | false         | not evaluated   | not evaluated   | false        |
+| 2000 | true          | false           | true            | true         |
+| 1900 | true          | false           | false           | false        |
+
+The chain of boolean expressions is efficient, as it proceeds from testing the most likely to least likely conditions.
```
siebenschlaefer, line 25: Howard Hinnant wrote a [short paragraph](https://howardhinnant.github.io/date_algorithms.html#is_leap) about the order of operations on his web page about date-related algorithms.

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Shortening
+
+By using the [implicit conversion][implicit-conversion] from `integer` to `bool` the code can be shortened using the [logical NOT operator][logical-not].
```
siebenschlaefer, line 29: `integer` -> `int`?

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+Only `0` is converted to `false` and _every non-zero_ value is converted to `true`.
+
+```cpp
+bool is_leap_year(int year) {
+    return !(year % 4) && ((year % 100) || !(year % 400 == 0));
+}
+```
+
+It can be thought of as the expression _not_ having a remainder.
```
siebenschlaefer, line 38: I don't understand this sentence.
I've always thought about `a % b` (in a boolean context) as "having" a remainder, not "not having" a remainder.

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Precedence
+
+You may have noticed the braces around the boolean expressions above.
```
siebenschlaefer, line 42: braces -> parentheses

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+Consider the following code:
+
+```cpp
+bool many_braces    = ((year % 100) != 0) || ((year % 400) == 0);
+bool some_braces    = (year % 100 != 0) || (year % 400 == 0);
+bool no_braces      =  year % 100 != 0 || year % 400 == 0;
+```
```
siebenschlaefer, line 49: in all three lines: replace "braces" with "parentheses" or "parens"

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+
+All lines produce the same result, but differ a lot in readability.
+The reason for their equal results, is the order of operator evaluation (`%`, `!=`, `||`, etc) or [precedence][operator-precedence].
```
siebenschlaefer, line 52: in both lines: remove comma

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+In this example the first operator to get the attention of the compiler is the remainder, followed by the equality operators.
```
siebenschlaefer, line 53: That sounds like the three "a % b == 0" sub-expressions would be evaluated first.  
Sadly I don't have a better alternative yet, perhaps Bethany ...?

exercises/practice/leap/.approaches/boolean-chain/content.md
```diff
@@ -0,0 +1,94 @@
...
+Only then will the compiler go for `&&` and then check `||`.
+Some people might overlook that `and` is evaluated before `or` and thus introduce bugs.
+This is a potential reason to keep some braces, although they are not necessarily needed in the code.
```
siebenschlaefer, line 56: Personally I would phrase this whole paragraph as a bad example: mixing `&&` and `||` at the same level of an expression is error-prone. Most linters and compilers (when warnings are enabled) will complain about that, I've heard/read advice from several experts against that.
For example: https://softwareengineering.stackexchange.com/q/201175/126640

exercises/practice/leap/.approaches/introduction.md
```diff
@@ -0,0 +1,72 @@
...
+
+There are several idiomatic approaches to solve Leap.
+You can use a chain of boolean expressions to test the conditions.
+Or you can use a [ternary operator][ternary-operator].
+You might also see functions from the `std::chrono` or `boost::gregorian` libraries.
+
+## General guidance
+
+The key to solving Leap is to know if the year is evenly divisible by `4`, `100` and `400`.
+For determining that, you will use the [remainder operator][remainder-operator].
```
siebenschlaefer, line 11: no comma

exercises/practice/leap/.approaches/introduction.md
```diff
@@ -0,0 +1,72 @@
...
+bool is_leap_year(int year) {
+    return year % 100 == 0 ? year % 400 == 0 : year % 4 == 0;
+}
+```
+
+For more information, check the [Ternary operator approach][approach-ternary-operator].
+
+## Other approaches
+
+Besides the aforementioned, idiomatic approaches, you could also approach the exercise as follows:
```
siebenschlaefer, line 35: i think the first comma can be omitted

siebenschlaefer (approved):
You put a lot of effort in. I like it.

I've mentored this exercise a lot. There are some more notable approaches:

1. Some students like to create three variables for the three atomic sub-expressions:
    ```cpp
    bool div_by_4 = year % 4 == 0;
    // ...
    return div_by_4 && (!div_by_100 || div_by_400);
    ```
    IMHO that's not bad, it makes the final `return` easy to read and the overhead is probably acceptable in most cases.
2. A few students did use a lambda:
    ```cpp
    auto div_by = [year](int divisor) { return year % divisior == 0; };
    return div_by(4) && (!div_by(100) || div_by(400));
    ```
    That's similar to (1) but avoids the computational overhead.
3. A lot of students want to solve this exercise step by step. Either starting with the most narrow case:
    ```cpp
    if (year % 400 == 0)
        return true;
    if (year % 100 == 0)
        return false;
    if (year % 4 == 0)
        return true;
    return false;
    ```
    Or starting with the most general case:
    ```cpp
    if (year % 4 != 0)
        return false;
    if (year % 100 != 0)
        return true;
    if (year % 400 == 0)
        return false;
    return true;
    ```
    That's definitely not wrong and it's easy to read. I usually challenge them to find a one-liner to get a shorter and more concise solution ... and then choose what they like more.
4. The two approaches from (3) can also be written as a series of nested `if` statements.
    ```cpp
    if (year % 4 == 0)
    {
        if (year % 100 == 0)
        {
            if (year % 400 == 0)
                return true;
            return false
        }
        return true;
    }
    return false;
    ```

Whether you want to add or mention some of them in your approaches is of course up to you.

siebenschlaefer:
Oh, I forgot: Three things that come up often in my mentoring sessions:

1. Omitting the name of the parameter in the declaration. I tell them:
    > But a user of your function does not only need to know the type of the parameter (`int`) but also its meaning (it's a year).
    > I'd suggest giving that parameter a name in the declaration. That makes no difference to the compiler, but it helps humans who read that declaration. It's a cheap way of documenting that aspect of a function.

2. Using an unsigned type for the parameter. I tell them:
    > I guess you did that because you think the year will always be positive, right?
    > [ES.100-102 of the C++ Core Guidelines](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-mix) talk about signed and unsigned types. They basically state that you should use signed types for arithmetic, unsigned types for bit manipulation, and that you should never mix them.
    > Even if you use an unsigned type for the argument the user could still pass a negative integer to the function and that's a perfectly valid thing to do because the conversion from signed to unsigned is well defined. But the result of the function would probably be unexpected.
    > Some might argue that negative years or the year `0` do not exist, or that the Gregorian Calendar wasn't invented back then.  The Roman Catholic Church made the [Gregorian Calendar](https://en.wikipedia.org/wiki/Gregorian_calendar) official in 1582, starting with October, 15th. Is 1580 a leap year? If 1580 is a leap year, why not the year `0` or `-444`?
The adoption of the calendar took many years, Saudi Arabia only started using it in 2016. Was 2012 a leap year or does it depend on the location?
    > If you want to have some "start year" and consider all years before that not to be leap years you should at least document that.
    > But IMHO a better alternative is to implement `is_leap_year()` without a start year. Extending the Gregorian calendar backwards to dates before Oct 15th, 1582, even with the year `0` and negative years, gives you the [Proleptic Gregorian calendar](https://en.wikipedia.org/wiki/Proleptic_Gregorian_calendar). That calendar is widely used, e.g. by ISO 8601.

3. Using a "result variable" like this:
    ```cpp
    bool result = false;
    if (year % 4 == 0) {
        result = true;
        if (year % 100 == 0) {
            result = false;
            if (year % 400 == 0) {
                result = true;
            }
        }
    }
    return result;
    ```
    I always try motivating them to find a solution that returns directly, with multiple `return` statements or (even better) to find a single expression ... and then choose what they like more.

I know that not all of this information will be usable in the "approaches", but maybe at least some other mentor might find it useful.

exercises/practice/leap/.approaches/boolean-chain/content.md
vaeng: Do you want me to include the link, improve the wording, or just provide an interesting link?

exercises/practice/leap/.approaches/boolean-chain/content.md
vaeng: Ic ould be more explicit an write down every line as they are evaluated. But that might be a bit too much.

vaeng:
> I know that not all of this information will be usable in the "approaches", but maybe at least some other mentor might find it useful.

This is very good information. As many of those are versions that would evolve into one of the provided approaches, I would not include them as their own approaches.

Do we have mentoring notes on the track were this could be used? I have not done anything with the mentoring system yet.

vaeng:
I will merch this to use it for the article, we could add more in the future.

PR #762, opened 2024-01-09
feat: add rotational cipher exercise

vaeng:
According to https://gist.github.com/ErikSchierboom/2f121db2bc55003129408edcd19183dc, this exercise is the second most popular (among other tracks) exercise, that has not yet been made available on the cpp track.

Let's change that.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1041,8 +1041,12 @@
         "slug": "rotational-cipher",
         "name": "Rotational Cipher",
         "uuid": "21d02535-4ab6-455d-88e1-08ec3b578894",
-        "practices": [],
-        "prerequisites": [],
+        "practices": [
+          "strings"
+        ],
+        "prerequisites": [
+          "loops"
+        ],
         "difficulty": 3,
         "topics": []
       }
--- a/config.json
+++ b/config.json
@@ -1036,6 +1036,15 @@
         "difficulty": 3,
         "topics": [
         ]
+      },
+      {
+        "slug": "rotational-cipher",
+        "name": "Rotational Cipher",
+        "uuid": "21d02535-4ab6-455d-88e1-08ec3b578894",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 3,
+        "topics": []
       }
     ],
     "foregone": [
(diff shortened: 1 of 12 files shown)


exercises/practice/rotational-cipher/.meta/example.cpp
```diff
@@ -0,0 +1,22 @@
+#include "rotational_cipher.h"
+
+#include <sstream>
+
+namespace rotational_cipher {
+    std::string  rotate(const std::string& text, int shiftKey) {
+
+        std::ostringstream encoded_text;
```
siebenschlaefer, line 8: style nitpick: "lowerCamelCase" for the parameter but "snake_case" for the variable. Is that intentional?

exercises/practice/rotational-cipher/.meta/example.cpp
```diff
@@ -0,0 +1,22 @@
+#include "rotational_cipher.h"
+
+#include <sstream>
+
+namespace rotational_cipher {
+    std::string  rotate(const std::string& text, int shiftKey) {
```
siebenschlaefer, line 6: style nitpick: are the two spaces between return type and function name intentional? (in the .h file, too)

config.json
```diff
@@ -1036,6 +1036,15 @@
...
         ]
+      },
+      {
+        "slug": "rotational-cipher",
+        "name": "Rotational Cipher",
+        "uuid": "21d02535-4ab6-455d-88e1-08ec3b578894",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 3,
+        "topics": []
```
siebenschlaefer, line 1047: I think "strings" and "loops" would be good topics.
Maybe "math", too, because of the modulo arithmetic.

siebenschlaefer (approved):
Great! No issues or complaints, just two small nitpicks and maybe some topics.

config.json
vaeng: Every concept that is added in the "practices" field, makes the syllabus tree a bit bigger. Loops is quite big already, but we could add loops as a prereq and add it to the strings concept, so it is only available if people finished loops?

siebenschlaefer: Oh, I didn't realize that. I'll try to remember that.

PR #763, opened 2024-01-10
feat: add diamond exercise

vaeng:
Another exercise from https://gist.github.com/ErikSchierboom/2f121db2bc55003129408edcd19183dc

### DIFF
--- a/exercises/practice/diamond/.meta/config.json
+++ b/exercises/practice/diamond/.meta/config.json
@@ -11,8 +11,8 @@
       "diamond_test.cpp"
     ],
     "example": [
-      "example.cpp",
-      "example.h"
+      "/.meta/example.cpp",
+      "/.meta/example.h"
     ]
   },
   "blurb": "Given a letter, print a diamond starting with 'A' with the supplied letter at the widest point.",
--- a/exercises/practice/diamond/.meta/config.json
+++ b/exercises/practice/diamond/.meta/config.json
@@ -1,0 +1,21 @@
+{
+  "authors": [
+    "vaeng"
+  ],
+  "files": {
+    "solution": [
+      "diamond.cpp",
+      "diamond.h"
+    ],
+    "test": [
+      "diamond_test.cpp"
+    ],
+    "example": [
+      "example.cpp",
+      "example.h"
+    ]
+  },
+  "blurb": "Given a letter, print a diamond starting with 'A' with the supplied letter at the widest point.",
+  "source": "Seb Rose",
+  "source_url": "https://web.archive.org/web/20220807163751/http://claysnow.co.uk/recycling-tests-in-tdd/"
+}
--- a/exercises/practice/diamond/.meta/config.json
+++ b/exercises/practice/diamond/.meta/config.json
@@ -11,8 +11,8 @@
       "diamond_test.cpp"
     ],
     "example": [
-      "/.meta/example.cpp",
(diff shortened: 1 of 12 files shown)


exercises/practice/diamond/.meta/config.json
```diff
@@ -0,0 +1,21 @@
...
+    "solution": [
+      "diamond.cpp",
+      "diamond.h"
+    ],
+    "test": [
+      "diamond_test.cpp"
+    ],
+    "example": [
+      "/.meta/example.cpp",
+      "/.meta/example.h"
```
siebenschlaefer, line 15: are you sure about the leading slashes?

exercises/practice/diamond/diamond_test.cpp
```diff
@@ -0,0 +1,53 @@
...
...                                     V    ", "     U                                       U     ", "      T                                     T      ", "       S                                   S       ", "        R                                 R        ", "         Q                               Q         ", "          P                             P          ", "           O                           O           ", "            N                         N            ", "             M                       M             ", "              L                     L              ", "               K                   K               ", "                J                 J                ", "                 I               I                 ", "                  H             H                  ", "                   G           G                   ", "                    F         F                    ", "                     E       E                     ", "                      D     D                      ", "                       C   C                       ", "                        B B                        ", "                         A                         "};
```
siebenschlaefer, line 49: Would it be possible to format the expected result in a way that the "diamond" can be seen in the source code?

exercises/practice/diamond/.meta/example.cpp
```diff
@@ -0,0 +1,25 @@
+#include "diamond.h"
+
+namespace diamond {
+    std::vector<std::string> rows(char middle_letter) {
+        if (middle_letter == 'A') return {"A"};
+        std::vector<std::string> diamond;
+        for(char c{'A'}; c <= middle_letter; ++c) {
+            int outer_spacing = middle_letter - c;
+            std::string row{};
+            row.append(std::string(outer_spacing, ' '));
```
siebenschlaefer, line 10: Did you know that `std::string::append()` has an overload that takes a count and a `char`? That would shorten these `append` lines a little bit.

siebenschlaefer (approved):
Looks good!

I didn't check the example implementation thoroughly, I have to trust you on this.

exercises/practice/diamond/.meta/config.json
vaeng: Now that you say it, I would remove them. Although configlet is not complaining.

exercises/practice/diamond/diamond_test.cpp
vaeng: Great idea. That would be nice, I will try that and see if it reaches the user in the web editor.

exercises/practice/diamond/.meta/example.cpp
vaeng: Good catch, coding felt clunky, but I did not have that in my toolbelt to make it better.

exercises/practice/diamond/diamond_test.cpp
siebenschlaefer: Just to avoid a misunderstanding: I meant reformatting the source code, each element (line) of `expected` on its own line so that the pattern is visible in the code.
Changing how a failed test gets displayed would be nice but I suspect that would be much harder.

PR #768, opened 2024-01-13
feat: add sublist exercise

vaeng:


vaeng:
Fails to compile under windows. I don't see if `std::each` has any problems on windows system. I might have to change the example.cpp

ErikSchierboom (approved):

siebenschlaefer (commented):
Looks good to me.  
Just one tiny thing:

`List_comparison` looks like a mix of "PascalCase" and "snake_case". AFAIK this would be a first here on the C++ track. *Personally* I would prefer either PascalCase *or* snake_case, but that's not a hill I want to die on. I just want to make sure you did that on purpose.

vaeng:
@siebenschlaefer :
 > PascalCase" and "snake_case".

Isn't that what the [core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-enum) suggest? I am open to changing it,

siebenschlaefer:
I just grep-ed the practice exercises.  
The only example solutions with "PascalCase" and "Titlecase" are:

- acronym enum: `enum DelimiterState`
- bank-account: `class Bankaccount`
- complex-numbers: `class Complex`
- high-scores: `class HighScores`
- robot-simulator: `enum class Bearing` and `class Robot`
- simple-linked-list: `class List` and `struct Element`

Everywhere else we use "flatcase" or "snake_case" for `class`, `struct` or `enum`.

Choose what you like more, I'm fine with all of them.

PR #770, opened 2024-01-14
[ellens_alien_game] change "hit" from task 1 to task 2

IsaacG:
See [forum](http://forum.exercism.org/t/ellens-alien-game-c-why-task-1-fails/9150/5)


IsaacG:
I jumped the gun and should have asked the reporter if they're like to file a PR.

vaeng:
I did not read until the end of the thread and also did a PR before I saw yours. We are bad people :D

PR #785, opened 2024-01-24
Add exercise two-bucket.

siebenschlaefer:
This is a translation of the "two-bucket" exercise.

I've decided to provide the complete `.h` file with a definition of the `enum`, the `struct` that holds the result, and a declaration of the function.
Although I like the idea of letting the students write their own `.h` file by extracting the interface from the tests I think in this case it would be just chore.
But feel free to object, that's not a strong preference of mine.

The thing that bothered me the most is how to handle cases that are not possible to solve.
Requiring exceptions for "expected" error cases that might occur frequently and are part of the normal control flow is IMHO bad style.
If I could I would use C++23's `std::expected` but that's not an option at the moment.
Here are some alternatives:

- I could add a member `possible` to the return type (that's what the C version does). But it feels C-ish.
- The function could return some error code or `bool` and use output parameters for the actual result. Again, C-ish and IMHO bad style.
- The function could return a `std::optional` that is empty if the task is impossible. But I'm not sure if the semantics were clear and it would be harder for students who do not know `std::optional`.

If any of you has an idea, I'm all ears.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1106,6 +1106,15 @@
         "prerequisites": [],
         "difficulty": 5,
         "topics": []
+      },
+      {
+        "slug": "two-bucket",
+        "name": "Two Bucket",
+        "uuid": "83c92b7f-fb0e-43d8-ab7d-d514408b8575",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 7,
+        "topics": []
       }
     ],
     "foregone": [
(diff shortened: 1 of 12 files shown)


vaeng (approved):
LGTM.

Do you want to add mentoring notes and/or an approach article? Your solution looks very compact and nice.

siebenschlaefer:
The example solution was just the simplest solution I could think of. There is a more general approach that treats this problem as a breadth-first search without inspecting its properties and a mathematical approach that solves a Diophantine equation (I was obsessed with this problem for a week in November and am kind of proud that I found this one.)

I definitely want to write either approaches or a blog post but not in the next two weeks.  
As for the mentoring notes, I want to wait a little bit until we have a reasonable sample of solutions from which we can extract talking points.

vaeng:
> I definitely want to write either approaches or a blog post but not in the next two weeks.

The exercise is not due to be featured in #48in24 until October, so there is a lot of time :)

Issue #779, opened 2024-01-16
Add #48in24 featured exercises


vaeng:
As of Jan 16 2024, the following exercises will show up as featured exercises in #48in24, but are not yet available on the C++ track.

In the order of appearance in the challenge:

- [x] list-ops (Feb-20) (@vaeng, #801)
- [x] linked-list (March-05) (@vaeng, https://github.com/exercism/cpp/pull/567)
- [x] parallel-letter-frequency (March-12) (@ahans, #800)
- [x] zebra-puzzle (May-07) (@rajatnai49, https://github.com/exercism/cpp/pull/788)
- [x] dnd-character (May-21) (@vaeng, #802)
- [x] pig-latin (May-28) (@vaeng, #812)
- [x] yacht (June-11) (@vaeng, #803)
- [x] spiral-matrix (July-09) (@ahans, #807)
- [x] kindergarten-garden (July-30) (@vaeng, #804)
- [x] knapsack (August-27) (@siebenschlaefer, https://github.com/exercism/cpp/pull/795)
- [x] two-bucket (October-08)

Take a look at https://docs.google.com/spreadsheets/d/1X00y60Z6BqYbTMnmmuMYzMoS0eO4KKquSWq2ijXb4Qo/edit#gid=0 to see current status.  
There is now a dedicated matrix with the status of all exercises in #48in24, which tracks have implemented them, and which exercises are featured on a specific track:
https://exercism.org/challenges/48in24/implementation_status

If you want some help to build all the files, you might like the [exercise generation](https://github.com/exercism/cpp/tree/feat--practice-exercise-generator/bin) script I wrote.

vaeng:
I will add list-ops

vaeng:
I already started linked-list a while ago, so please don't start another. I promise to finish it soon 😃

rajatnai49:
Hey, @vaeng I like to start contributing with the zebra-puzzle !

vaeng:
> Hey, @vaeng I like to start contributing with the zebra-puzzle !

Sure thing. Have fun and keep us in the loop. It's not an easy task, so there will be a lot of discussions and feedback.

rajatnai49:
Okay @vaeng I try my best.

rajatnai49:
Hey @vaeng I opened pr but it closed automatically, can you guide me for what steps I have to follow for successfully open the pr

siebenschlaefer:
> I can have a look at `parallel-letter-frequency`.

[...]

PR #788, opened 2024-01-29
Add practice exercise Zebra Puzzle

rajatnai49:
issue number: #779


exercises/practice/zebra-puzzle/.meta/config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "authors": ["rajatnai49"],
+  "files": {
+    "solution": [
+      "zebra_puzzle.cpp",
+      "zebra_puzzle.h"
```
vaeng, line 6: I have to redo the config, until then please manually correct this part to:

```suggestion
      ".meta/zebra_puzzle.cpp",
      ".meta/zebra_puzzle.h"
```

exercises/practice/zebra-puzzle/.meta/example.cpp
vaeng, line 1: This is the most difficult part of the puzzle. 
I would prefer this solution to solve the riddle programmatically, rather than having the answer hardcoded. Do you think you can do that?

vaeng (changes requested):
Please add a programmatic way of solving this puzzle.

You can define enums, functions and other more complicated testing scenarios to achieve this goal.

exercises/practice/zebra-puzzle/.meta/config.json
siebenschlaefer: > I have to redo the config, until then please manually correct this part to:

Are you sure? I think these two entries for the initial stub are correct, the example solution (`example.h`, `example.cpp`) should be moved to the `.meta/` directory.

vaeng: I think our config.json is not correct, so the files and the part above are not generated according to the rest of the track:

https://github.com/exercism/cpp/blob/dad723ff78e3054f917469895b52c7fb66217f71/config.json#L30

vaeng:
Thanks for the update!

Do you think we could have the details of the story in the test file? So the result is not always the same and students cannot just answer "Norwegian"?

So that we have a custom test case, where we give another story? I will ask the other maintainers if they did that. What do you think @siebenschlaefer ?

It is possible a bigger rewrite and I think you should get more than the standard 12 reputation for that, if you want to continue.

siebenschlaefer:
Hej @rajatnai49,

I'm a little bit torn. The example solution still contains a lot of "human reasoning", e.g. why the first house is yellow is deduced from facts 6, 15, and 2 and then hard-coded (`houses[0].color = YELLOW;`)  
I would prefer an example solution that does that as little as possible, let your program find the answer.

For example, you could use five nested loops (for the nationality, the color, ...), try all permutations, and skip a particular permutation as soon as possible when it doesn't match the facts. Or you could try to implement a completely different approach, that's up to you.  
If you need inspiration you could look at the solutions of this exercise in other tracks or at [rosettacode](https://rosettacode.org/wiki/Zebra_puzzle).
And if you need help, let me know, either here or on our Discord server.

And two small things:
- Why are `drinks_water()` and `owns_zebra()` member functions of a `class`? Unless that `class` has some state I'd suggest defining them as free functions instead.
- For constants use `const` or `constexpr` variables instead of macros. That's a broadly accepted best practice.

rajatnai49:
Sorry @siebenschlaefer for disappointment😞, and I know that all permutation solution but I think it will be good if we just first remove that and after that try every possible cases for perticular conditions, and it is much efficient than trying all permutation
But, I understood your point that you want code to do every work, so fine I will work
And also I will take note of those practices this time around!!

siebenschlaefer:
> Sorry @siebenschlaefer for disappointment😞

No worries, please! As vaeng [wrote earlier](https://github.com/exercism/cpp/issues/779#issuecomment-1913264259) this is not an easy exercise. And maybe we maintainer should have communicated more clearly what kind of example solutions we're looking for.

And if you need any help, let us know.

siebenschlaefer:
@rajatnai49 Are you still working on this? Is there anything we can do to help you?

rajatnai49:
Hey, @siebenschlaefer sorry for delay but we are organizing the Hackathon in the university so I didn't have much time to work on the issue so I will give you update in next week for sure!!

rajatnai49:
@siebenschlaefer I added the solution as you explained earlier, review that and if any other requirements are there then let me know !

siebenschlaefer:
Much better, this is now an algorithmic solution.

---

But it takes a really long time to run because it tries all possible potential and finds the solution after more than 171 million permutations. On my computer it runs in a a little bit more than one minute. Students wouldn't be able to submit this example solution because it would not complete the tests within the allowed time (I think the test runner has a limit of 20 or 30 seconds.)  
We need to bring that time down because we want an example solution that *could* be successfully submitted.

There are five nested loops (for the `cigarettes`, `drinks`, `pets`, `nationalities`, and `colors`). Currently the innermost loop assigns values to the members of the `houses` and checks whether that's a valid solution.

Instead you could assign assign the values to the houses as soon as possible, i.e. the `.cigarette` in the outer loop, the `.drink` in the first inner loop, the `.pet` in the second inner loop, the `.nationality` in the third inner loop, and the `.color` in the innermost loop. That alone cuts the runtime in half (at least on my computer).  

Instead of checking the complete solution in the innermost loop you could check each of the 14 clauses as soon as possible and `continue` if they don't hold. For example the outermost loop creates permutations of the cigarette brands, the first inner loop creates permutations of the drinks. You could check clauses 8 ("Milk is drunk in the middle house") and 12 ("The Lucky Strike smoker drinks orange juice.") right there, in the first inner loop. That would allow you to skip generating all the permutations of pets, nationalities, and colors when you know that the will never create a valid solution because `houses[2].drink` is not `Milk` or because the Lucky Strike smoker does not drink OJ. If you do that for all clauses (check each clause in the outermost possible loop) you can reduce the runtime drastically, on my computer it runs in under one second.

---

And some smaller nitpicks: 

Four of the five `std::map` instances are not used at all. I'd suggest deleting them. That makes the code easier to read and to maintain.

As far as I can see the `return` statement in line 159 is the only one that is not indented by a multiple of four spaces.

---

And finally a more general question where @vaeng might also want to weigh in:

Currently the tests require two separate and free functions that answer the two questions. But is that really the best structure for this problem. It basically forces the students to write some helper function that solves the problem and gets called by the two "public" functions, and it will make the test runner to solve the puzzle twice (unless the students implement some caching) without a clear benefit.

How about a single function that returns both answers at the same time? That's how the Go track does it.

exercises/practice/zebra-puzzle/.meta/config.json
```diff
@@ -0,0 +1,19 @@
...
+    ],
+    "test": [
+      "zebra_puzzle_test.cpp"
+    ],
+    "example": [
+      "example.cpp",
+      "example.h"
```
vaeng, line 13:
```suggestion
      ".meta/example.cpp",
      ".meta/example.h"
```

vaeng (dismissed):

vaeng:
> How about a single function that returns both answers at the same time? That's how the Go track does it.

We could also go the class way and generate the puzzle solution on construction and have the answers in variables ... Might be ugly :D

siebenschlaefer:
> We could also go the class way and generate the puzzle solution on construction and have the answers in variables ... Might be ugly :D

Right, that's another option. (although *personally* I'm not a fan)

rajatnai49:
@siebenschlaefer totally understood ☺️, working on that. and both single function or class construction approaches are great because there is a only one particular answer for the zebra puzzle, so solving puzzle two time, like in the python track will not be add any value, So let me know on which path we are going.

vaeng:
> So let me know on which path we are going.

a single function that returns both answers at the same time.

rajatnai49:
@vaeng @siebenschlaefer 
can you please give me the status/review of this PR, anything required?

exercises/practice/zebra-puzzle/.meta/example.cpp
```diff
@@ -0,0 +1,244 @@
...
+using namespace std;
+
+namespace zebra_puzzle 
+{
+
+typedef struct 
+{
+    string drinksWater;
+    string ownsZebra;
+} Solution;
```
siebenschlaefer, line 16: If you '#include` the `.h` file you won't need to repeat this definition (see above).

exercises/practice/zebra-puzzle/.meta/example.h
```diff
@@ -0,0 +1,19 @@
...
+#include <string>
+
+namespace zebra_puzzle 
+{
+
+typedef struct 
+{
+    std::string drinksWater;
+    std::string ownsZebra;
+} Solution;
```
siebenschlaefer, line 13: `typedef struct { ... } type_name;` is a C idiom.  
In C++ you can write `struct type_name { ... };`, just like you did with `House` in the `.cpp` file.

exercises/practice/zebra-puzzle/.meta/tests.toml
```diff
@@ -0,0 +1,14 @@
...
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
+
+[16efb4e4-8ad7-4d5e-ba96-e5537b66fd42]
+description = "solve the zebra puzzle and answer who drinks water and who owns zebra"
```
siebenschlaefer, line 13: This file should describe which tests from the language-independent [problem specification](https://github.com/exercism/problem-specifications/blob/main/exercises/zebra-puzzle/canonical-data.json) are implemented in this track, the UUID and description should match those from the JSON file.  
The `zebra_puzze_tests.cpp` implements both tests ("resident who drinks water" and "resident who owns zebra") so this TOML file should reflect that.

exercises/practice/zebra-puzzle/zebra_puzzle.cpp
```diff
@@ -0,0 +1,21 @@
+#include "zebra_puzzle.h"
+
+#include <string>
+
+using namespace std;
```
siebenschlaefer, line 5: While even some C++ programmers like to have this `using namespace std;`  in their `.cpp` files most C++ experts recommend not to do that. I'd suggest removing that line, we don't want to promote bad practices.

exercises/practice/zebra-puzzle/zebra_puzzle.cpp
```diff
@@ -0,0 +1,21 @@
+#include "zebra_puzzle.h"
+
+#include <string>
```
siebenschlaefer, line 3: This is not necessary, `<string>` is already included in the `.h` file.

exercises/practice/zebra-puzzle/zebra_puzzle.cpp
```diff
@@ -0,0 +1,21 @@
...
+using namespace std;
+
+namespace zebra_puzzle 
+{
+
+typedef struct 
+{
+    std::string drinksWater;
+    std::string ownsZebra;
+} Solution;
```
siebenschlaefer, line 14: *\*If\** you want to define that type for our students, do that in the `.h` file where it is required.  
(and without the `typedef` idiom, see above.)

exercises/practice/zebra-puzzle/zebra_puzzle.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+Solution solve() 
+{
+    // IMPLEMENT YOUR LOGIC HERE
+}
```
siebenschlaefer, line 19: *If* you want to provide a function stub for the students as a hint which function they must implement, and what it returns, declare it in the `.h` file, like in the other exercises that do this.

siebenschlaefer (commented):
This looks good!

There are a few smaller things that I'd like to improve before merging.

The only potential problem that I can see is that the tests use `SECTION` which does not map directly to the two tests in the [problem specification](https://github.com/exercism/problem-specifications/blob/main/exercises/zebra-puzzle/canonical-data.json). The test name "drinks water" and UUID does not reflect that combined nature of the tests. I'd rather have two separate and independent tests.  

@vaeng what do you think?

vaeng:
@rajatnai49 
The exercise is going to be featured in two weeks. Can you cover siebenschläfer's suggestions in the next days, so we can have it online soon?

siebenschlaefer (approved):
Looks good. Thank you very much!

vaeng (approved):

vaeng:
This is it. Thaks a lot for all the work in this exercise @siebenschlaefer and especially @rajatnai49!

Issue #775, opened 2024-01-14
Update test cases


vaeng:
Grab what you want, mark it and make a PR. These exercises need updates:

- [ ] allergies
- [ ] anagram
- [ ] armstrong-numbers
- [ ] bank-account
- [ ] complex-numbers
- [ ] crypto-square
- [ ] gigasecond
- [ ] grade-school
- [ ] hamming
- [ ] high-scores
- [ ] isogram
- [ ] luhn
- [ ] matching-brackets
- [ ] pangram
- [ ] perfect-numbers
- [ ] phone-number
- [ ] prime-factors
- [ ] protein-translation
- [ ] say
- [ ] triangle
- [ ] word-count

vaeng:
I will start with: allergies

CadeMichael:
I'd love to help but am not sure what needs updating?

IsaacG:
> I'd love to help but am not sure what needs updating?

The above exercises need their tests synced from the problem specs.

See docs for 
* [Practice Exercises](https://exercism.org/docs/building/tracks/practice-exercises)
* [configlet tool: syncing](https://exercism.org/docs/building/configlet#h-syncing-exercise-data-with-the-problem-specifications-repo)

Reading through prior commits on this repo (look for PRs that mention syncing tests) might prove helpful.

vaeng:
@CadeMichael 
Give me some time, I'll write a how-to for Test Cases and put it in the tracks wiki.

vaeng:
Here is a [wiki entry on test case updates](https://github.com/exercism/cpp/wiki/Test-Cases-for-Practice-Exercises) to explain what I have been doing so far, maybe it is helpful for you  @CadeMichael 
Do you have additions or suggestions @siebenschlaefer ?

CadeMichael:
Thank You I'll check it out and try to update for anagram!

CadeMichael:
Sorry for the slow progress long weekend. I got configlet working and found the testcases that need improvement

```
[warn] anagram: missing 6 test cases
       - detects two anagrams (03eb9bbe-8906-4ea0-84fa-ffe711b52c8b)
       - does not detect an anagram if the original word is repeated (630abb71-a94e-4715-8395-179ec1df9f91)
       - words are not anagrams of themselves (68934ed0-010b-4ef9-857a-20c9012d1ebf)
       - words are not anagrams of themselves even if letter case is partially different (589384f3-4c8a-4e7d-9edc-51c3e5f0c90e)
[...]

PR #789, opened 2024-01-29
Sync hamming tests to problem-description data (#775)

ahans:
As discussed in #784, I was also confused about the `include = false` lines. The way I understand it, in `canonical-data.json` test cases are updated by adding a new version and a `reimplements` field that references the old version in the file. The old version is left untouched. I can imagine this is done to better be able to follow the edit history. But this also means that obsolete tests should _not_ be part of the implementation. So that is what I have done here.


vaeng:
We can surely discuss this on the forums. Please note, that it would be better to have uuids as well in the test.cpp file, so they are easier to compare.

ahans:
This has been updated.

vaeng (approved):

PR #801, opened 2024-02-12
feat: add list-ops exercise

vaeng:


exercises/practice/list-ops/.docs/instructions.md
```diff
@@ -0,0 +1,19 @@
+# Instructions
+
+Implement basic list operations.
+
+In functional languages list operations like `length`, `map`, and `reduce` are very common.
+Implement a series of basic list operations, without using existing functions.
```
vaeng, line 6: As it is stated in the instructions, we should not reiterate that in an additional amendment, right?

For the example I even skipped .size() for the length. But as a mentor, I think I would let it pass.

ErikSchierboom: > As it is stated in the instructions, we should not reiterate that in an additional amendment, right?

Whatever you think makes the most sense 🤷 It's fine to let it pass as a mentor, but I would argue that just shelling out to existing functions does not really teach the student much.

ErikSchierboom (approved):

PR #804, opened 2024-02-12
Feat--add-kindergarten-garden-exercise

vaeng:


exercises/practice/kindergarten-garden/.meta/example.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <array>
+#include <string>
+
+namespace kindergarten_garden {
+    std::array<std::string, 4> plants(std::string diagram, std::string student);
```
ahans, line 7: In case you don't like the `std::string_view`, maybe pass the arguments via const reference at least?

vaeng: I do like string views. A lot in fact. But catch2 is not a fan and sometimes fails. I'll look into it again. Const ref would be better

exercises/practice/kindergarten-garden/.meta/example.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+#include <map>
+
+namespace kindergarten_garden {
+const std::map<std::string, std::string> flower_lookup{
+    {"G", "grass"}, {"C", "clover"}, {"R", "radishes"}, {"V", "violets"}};
+
+std::array<std::string, 4> plants(std::string diagram, std::string student) {
+    size_t row_length{diagram.find('\n')};
+    int index = 2 * (student.at(0) - 'A');
```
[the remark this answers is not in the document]
vaeng: I'm pretty sure that the trick is exactly that. Changing the tests is a delicate thing on exercism, I don't think I am the only one seeing and using this connection. So there would be some big discussions. If incoming changes would break this "trick" we could re-implement the example solution with the new tests.
/

exercises/practice/kindergarten-garden/.meta/example.h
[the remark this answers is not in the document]
vaeng: Enums are a great idea to have more material for the syllabus!

exercises/practice/kindergarten-garden/.meta/example.cpp
```diff
@@ -0,0 +1,18 @@
+#include "kindergarten_garden.h"
+
+#include <map>
+
+namespace kindergarten_garden {
+const std::map<std::string, std::string> flower_lookup{
```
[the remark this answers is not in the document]
vaeng: Thanks for the suggestions. The example solution does not aim to be the best, fastest, or prettiest solution for an exercise. In the most basic sense, it is just a "proof of concept", that the code and the tests can be run.

If you want to take the time and write about different strategies you might be interested in creating an approach for the exercise, so the students can enjoy them.

Currently, the example is only shown to mentors (and people who crawl through the github repository). So the audience is very narrow. Approaches are shown to everyone.

exercises/practice/kindergarten-garden/.meta/example.cpp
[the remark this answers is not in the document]
ahans: I didn't mean it in the sense that the tests would _actually_ be changed, but more in the sense of writing code that is flexible and extendable. But using this trick is also fine. Just wanted to give you my perspective.

exercises/practice/kindergarten-garden/.meta/example.cpp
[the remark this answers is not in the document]
ahans: Thanks, that makes sense. Again, just wanted to give my perspective. I don't have any authority around here anyways, so you might just as well ignore whatever I write. 😉

exercises/practice/kindergarten-garden/.meta/example.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include <array>
+#include <string_view>
+
+namespace kindergarten_garden {
+
+enum class Plants { grass = 'G', clover = 'C', radishes = 'R', violets = 'V' };
```
ahans, line 8:
```suggestion
enum class Plants : char { grass = 'G', clover = 'C', radishes = 'R', violets = 'V' };
```

ahans (commented):
Thanks for considering my previous comments. Here's one more very minor one.

exercises/practice/kindergarten-garden/.meta/example.h
vaeng: Too bad I still don't get an implicit cast with this. I could skip to loc.

ahans: True. But there's no such thing as implicit conversion from underlying to enum. You could cast the input array as enums, but I wouldn't consider that very clean. There's one other thing you could do to make the code.a bit more compact, though. Let me add another comment there.

exercises/practice/kindergarten-garden/.meta/example.cpp
```diff
@@ -0,0 +1,21 @@
...
+    const int first_plant_index = 2 * (student.at(0) - 'A');
+    const size_t second_plant_index{diagram.find('\n') + first_plant_index + 1};
+    const std::array<char, 4> positions{
+        diagram[first_plant_index], diagram[first_plant_index + 1],
+        diagram[second_plant_index], diagram[second_plant_index + 1]};
+    std::array<Plants, 4> result;
+    std::transform(positions.cbegin(), positions.cend(), result.begin(),
+                   [](char c) { return static_cast<Plants>(c); });
+
+    return result;
```
ahans, line 18: Since we always only need four elements, I think it'd be just as fine to duplicate the `static_cast`s and directly construct and return the result:
```suggestion
    return {
        static_cast<Plants>(diagram[first_plant_index]), static_cast<Plants>(diagram[first_plant_index + 1]),
        static_cast<Plants>(diagram[second_plant_index]), static_cast<Plants>(diagram[second_plant_index + 1]
    };
```

But all of that is pretty much a matter of taste now ...

vaeng: I had that in one iteration but decided on the transform. As you say, a matter of taste. :)

ErikSchierboom (approved):

PR #807, opened 2024-02-12
Add practice exercise spiral-matrix

ahans:


config.json
```diff
@@ -1124,6 +1124,14 @@
...
         "prerequisites": [],
         "difficulty": 7
+      },
+      {
+        "slug": "spiral-matrix",
+        "name": "Spiral Matrix",
+        "uuid": "d5569902-5815-4b05-ab13-13826fef7826",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 5
```
vaeng, line 1134: Do you think this difficulty is at a proper level? I have not done this exercise, but your code is not a 5 compared to the rest. Is there a simpler way, that just looks less idiomatic, or is it one of the more complicated exercises in your opinion?

exercises/practice/spiral-matrix/spiral_matrix.cpp
vaeng, line 1: In the other exercises, we have the stub with a namespace and the header include here. Is it intentional to leave it empty?

ahans: No, sorry! I think it was elsewhere that the implementation was completely empty and up to the student to fill out. But we should be consistent here. Will add the proper template.

ahans:
I also rebased the branch onto latest `main` now.

PR #810, opened 2024-02-16
Feature/sync gigasecond

ahans:


ahans:
@exercism/cpp here is the new PR. I had forced-pushed to the previous to include the `no important files changed` tag. I wasn't aware that this would prevent re-opening. Sorry about that!

SleeplessByte:
The failed check says:

````
This PR touches files which potentially affect the outcome of 
the tests of an exercise. This will cause all students' solutions
to affected exercises to be re-tested.

If this PR does **not** affect the result of the test (or, for
example, adds an edge case that is not worth rerunning all tests
for), **please add the following to the merge-commit message**
which will stops student's tests from re-running. Please
copy-paste to avoid typos.

```
[no important files changed]
```

For more information, refer to the [documentation](https://exercism.org/docs/building/tracks#h-avoiding-triggering-unnecessary-test-runs). 
If you are unsure whether to add the message or not, please ping
`@exercism/maintainers-admin` in a comment. Thank you!
````

**Comment by ahans** [comment]
> In this case ya probably want to re-run all the student solutions.

I only added that one excluded test to `tests.toml` and renamed the tests. I didn't change any actual test code. The changes there are only formatting (looks like `.clang-format` has changed in the meantime or hadn't been applied to that file before). So I'm not sure we should really re-run all student solutions. However, I can of course drop the tag. Let me know what to do!

**Comment by SleeplessByte** [comment]
I didn't see the _include false_ on the last line. If no tests were added (or changed) then indeed, it should NOT re-run ;)

```
[fcec307c-7529-49ab-b0fe-20309197618a]
description = "does not mutate the input"
include = false
```

vaeng (approved):
LGTM, thanks!

PR #811, opened 2024-02-20
feat: add `pig-latin` practice exercise

ahans:
I'm not entirely happy with the example implementation yet.

### DIFF
--- a/exercises/practice/pig-latin/pig_latin.cpp
+++ b/exercises/practice/pig-latin/pig_latin.cpp
@@ -1,0 +1,68 @@
+#include "pig_latin.h"
+
+#include <algorithm>
+#include <array>
+
+namespace pig_latin {
+
+namespace {
+
+[[nodiscard]] bool startswith(std::string_view word, std::string_view begin) {
+    if (word.length() < begin.length()) {
+        return false;
+    }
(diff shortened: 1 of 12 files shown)


exercises/practice/pig-latin/pig_latin.cpp
```diff
@@ -0,0 +1,68 @@
+#include "pig_latin.h"
```
ahans, line 1: This file and the `.h` obviously need to go to `.meta/example.{h,cpp}`.

vaeng:
Oh, no. I did not read that! Sorry. I also made a PR for that: #812

ahans:
> Oh, no. I did not read that! Sorry. I also made a PR for that: #812

Well, my bad for not mentioning that properly. Your call which one to keep. I think my implementation is more generic. I would have few comments how to make yours slightly more efficient as well.

vaeng:
> > Oh, no. I did not read that! Sorry. I also made a PR for that: #812
> 
> Well, my bad for not mentioning that properly. Your call which one to keep. I think my implementation is more generic. I would have few comments how to make yours slightly more efficient as well.

I'd be happy to see your suggestions. Let's make it a double :)

PR #812, opened 2024-02-20
feat: add pig-latin exercise

vaeng:


exercises/practice/pig-latin/.meta/example.cpp
```diff
@@ -0,0 +1,31 @@
+#include "pig_latin.h"
+
+namespace pig_latin {
+std::string translate_word(std::string word) {
```
ahans, line 4: It should be possible to pass `word` as `std::string_view`.

exercises/practice/pig-latin/.meta/example.cpp
```diff
@@ -0,0 +1,31 @@
...
+    for (std::string start : {"a", "e", "i", "o", "u", "yt", "xr"}) {
```
ahans, line 5: No need to copy the character (though for one or two characters it probably doesn't really make a difference):
```suggestion
    for (std::string_view start : {"a", "e", "i", "o", "u", "yt", "xr"}) {
```

exercises/practice/pig-latin/.meta/example.cpp
```diff
@@ -0,0 +1,31 @@
...
+        if (word.substr(0, start.size()) == start) return word + "ay";
+    }
+    for (std::string start : {"thr", "sch", "squ", "ch", "qu", "th", "rh"}) {
```
ahans, line 8: While this list of consonant clusters is sufficient to pass the tests, those cover only a fraction of the possibilities. Words for which the current implementation would not give the correct pig latin version include "friend", "splash", "smile", "string".

That's why I said in some other comment that my implementation is more generic. Instead of explicitly listing consonant clusters, it moves characters to the back until it finds a vowel. Only `qu` needs special handling in that case.

exercises/practice/pig-latin/.meta/example.cpp
```diff
@@ -0,0 +1,31 @@
...
+    return word.substr(1) + word.front() + "ay";
+}
+
+std::string translate(std::string_view phrase) {
+    size_t start = 0;
+    std::string translation{};
+    for (size_t found = phrase.find(' '); found != std::string::npos;
+         found = phrase.find(' ', start)) {
+        translation += translate_word(std::string(phrase.begin() + start,
+                                                  phrase.begin() + found)) +
```
ahans, line 21: You probably don't have explicitly create a new `std::string` here, but can use a subview via `phrase.substr(start, found - start)`

exercises/practice/pig-latin/.meta/example.cpp
```diff
@@ -0,0 +1,31 @@
...
+                       " ";
+        start = found + 1;
+    }
+    if (start != phrase.size())
+        translation +=
+            translate_word(std::string(phrase.begin() + start, phrase.end()));
```
ahans, line 27: Dito.

ahans (commented):
Here are my thoughts on your implementation. It certainly passes as a "proof of concept" either way.

siebenschlaefer (approved):
Looks good!

Solid example solution. There might be a few small details that could be improved (e.g. passing `std::string` by reference to `const` or calling `phrase.substr(start)` or some of the things from @ahans' comments) but nothing that would disqualify this as an example solution or block this PR from being merged.

exercises/practice/pig-latin/.meta/example.cpp
vaeng: I would vote to move that discussion and possible solutions to an article for the exercise.

PR #813, opened 2024-02-23
fix: improve tests and example for dnd character exercise

vaeng:
Thanks for the hints, it is now easier to understand for students.
You are correct about the last test, I removed it.


siebenschlaefer (approved):
Looks great!

PR #814, opened 2024-02-23
Fix the tests for Catch2-v2 and Catch2-v3

siebenschlaefer:
Catch2-v3 did change the namespace for the `MatcherBase`.  
This patch aims to support both Catch2 v2 and v3.
But in the long run we should move to v3 locally and on the test runner.


vaeng:
I have not seen your PR, sorry. 
Why is yours failing and mine is not?

PR #822, opened 2024-02-29
docs: add approach for all your base

vaeng:


exercises/practice/all-your-base/.approaches/from-base-ten/content.md
```diff
@@ -0,0 +1,118 @@
+# Converting from base 10
```
siebenschlaefer, line 1: This is a pet peeve of mine: I've mentored this solution so many times and many students seem to think that an `int` is inherently "decimal". When they convert the input to a intermediate numeric value they refer to it as "decimal" and sometimes they even add unnecessary conversion steps because of that.

I would strongly prefer calling it a "numeric value" instead of "base 10", here an in the rest of the approaches.

exercises/practice/all-your-base/.approaches/from-base-ten/content.md
```diff
@@ -0,0 +1,118 @@
...
+
+There are different strategies to convert a decimal number to another base.
+Here is a selection of a few approaches and thoughts on their performance.
+Keep in mind, that the number of digits you need to process is the most important factor for your algorithms performance.
+A simple for loop can be a lot faster for a small number than a [parallel algorithm][parallel-computing] that has more overhead to be set up.
```
siebenschlaefer, line 6:
```suggestion
A simple `for` loop can be a lot faster for a small number than a [parallel algorithm][parallel-computing] that has more overhead to be set up.
```

exercises/practice/all-your-base/.approaches/from-base-ten/content.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## 1: While-loop with `%` and `/`
+
+This approach uses a [while-loop][while-loop] to repeatedly divide the `intermediate` by the `output_base`, extracting the [remainders][remainder-operator]c as digits which are then stored in the `output_digits` vector.
```
siebenschlaefer, line 11:
```suggestion
This approach uses a [while-loop][while-loop] to repeatedly divide the `intermediate` by the `output_base`, extracting the [remainders][remainder-operator] as digits which are then stored in the `output_digits` vector.
```

ErikSchierboom (approved):
I love the content. They're a bit different from other approaches in that they don't constitute a full solution to the exercise, but parts of it. The only other way to handle this would be to have one approach use what you feel is the most relevant option of these two approaches, and then convert these two approaches to articles which the approach can then refer to. That said, I don't know if there is any other approach to solving this exercise without converting from base ten first, and if not, this is probably fine.

exercises/practice/all-your-base/.approaches/from-base-ten/content.md
```diff
@@ -0,0 +1,118 @@
...
+std::vector <unsigned int> output_digits{};
+std::ldiv_t v{intermediate, 0};        
+while (v.quot > 0) {
+    v = std::div(v.quot, output_base);
+    output_digits.emplace_back(v.rem);
+}
+```
+
+Depending on the processor this might save one operation per loop iteration.
+You do need to allocate some memory for the `v` variable of type `std::ldiv_t`.
```
siebenschlaefer, line 52: I wouldn't call that "allocate" because that sounds too much like "dynamically allocated memory".
How about something like "define a variable".

vaeng:
> The only other way to handle this would be to have one approach use what you feel is the most relevant option of these two approaches, and then convert these two approaches to articles which the approach can then refer to.

I had that as a first version. Approaches for `simple loops`, `loops that can work in parallel`, and `parallel algorithms`. But the second one was superfluous without the last, so I mashed them together and it got really long, so I cut a different way.

I waded through ~100 solutions from the community solutions and did some research on solving base transfers in general. However, I could not find anything that would work without an intermediate step over base 10.

exercises/practice/all-your-base/.approaches/introduction.md
```diff
@@ -0,0 +1,68 @@
+# Introduction
+
+The process of the _all your base_ exercise can be divided into  up to four parts:
+- conversion from input digit vector into a base 10 number
+- conversion from intermediate number into target base vector
+- possible reversing of the resulting vector
+- argument error checking
```
siebenschlaefer, line 7: There's also the interesting approach of doing the conversion directly, without an intermediate numeric value.
You need two functions that are usually found in "arbitrary-precision integer" a.k.a. "BigInteger" implementations: 
- a function `multiply()` that takes a `std::vector` of digits (in `out_base`) and the `in_base`, and multiplies the digits with the `in_base`.
- a function `add()` that takes a `std::vector` of digits (in `out_base`) and a single digit in `in_base` and adds the value of this single digit to the digits.

exercises/practice/all-your-base/.approaches/introduction.md
```diff
@@ -0,0 +1,68 @@
+# Introduction
+
+The process of the _all your base_ exercise can be divided into  up to four parts:
```
siebenschlaefer, line 3: Here and in the rest of the PR: "All Your Base" is the title of the exercise, I think it should be capitalized and either put in italics (like here) or in quotes.

exercises/practice/all-your-base/.approaches/introduction.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+For more information, check the [converting to base 10 approach][to-base-ten].
+
+## Approach: Converting from base 10
+
+```cpp
+while (intermediate != 0) {
+    unsigned int new_digit = intermediate % output_base;
+    intermediate = (unsigned int)(intermediate / output_base);
```
siebenschlaefer, line 43: Is this cast really necessary, aren't `intermediate` and `output_base` both `unsigned int`?
If it is necessary, please use a named cast instead of a C-style cast.

exercises/practice/all-your-base/.approaches/to-base-ten/content.md
```diff
@@ -0,0 +1,97 @@
...
+It is strictly serial.
+
+## Numeric's `accumulate` 
+
+The for-loop with `*` and `+` can be written with the [`accumulate`][acuumulate-function] function from the [`numeric`][numeric-header] header.
+
+```cpp
+#include <numeric>
+unsigned int intermediate = std::accumulate(input_digits.begin(),
+                                            input_digits.end(), 0,
```
siebenschlaefer, line 44:
```suggestion
                                            input_digits.end(), 0U,
```

exercises/practice/all-your-base/.approaches/to-base-ten/content.md
```diff
@@ -0,0 +1,97 @@
...
+
+```cpp
+std::vector <unsigned int> positions(input_digits.size());
+std::iota(positions.begin(), positions.end(), 0);
+unsigned int intermediate = std::transform_reduce(
+    std::execution::par_unseq,  // execution policy
+    input_digits.cbegin(),      // first input container start
+    input_digits.cend(),        // first input container end
+    positions.crbegin(),        // start second container from the rear
+    0,                          // initial value
```
siebenschlaefer, line 65:
```suggestion
    0U,                          // initial value
```

siebenschlaefer:
> I waded through ~100 solutions from the community solutions and did some research on solving base transfers in general. However, I could not find anything that would work without an intermediate step over base 10.

I don't want to toot my own horn, but ...
https://exercism.org/tracks/cpp/exercises/all-your-base/solutions/siebenschlaefer

ErikSchierboom:
> I don't want to toot my own horn, but ...
https://exercism.org/tracks/cpp/exercises/all-your-base/solutions/siebenschlaefer

Then we might have at least two approaches after all :)

siebenschlaefer:
> Then we might have at least two approaches after all :)

I think they can be written as three approaches:
- The most obvious that transforms the input to a numeric value and that numeric value to the output, using sequential operations. (with some variations how to write the loops)
- A parallel and concurrent approach that uses the standard library for the parallelism/concurrency. (I like that a lot, even if it might not be faster than the first one, I haven't seen that one before.)
- A BigInteger approach.

ErikSchierboom:
That sounds like a great plan!

exercises/practice/all-your-base/.approaches/from-base-ten/content.md
kotp: A numeral that is of base 10, the meaning of decimal.  While "numeric" is anything that is numeric, sometimes decimal, sometimes of a different base.

Since this is an exercise of bases, we should be triply clear on our wording concerning numbers.

vaeng:
I made the changes.
Maybe you can add an extra approach for your method to my short outline @siebenschlaefer

vaeng:
> > Maybe you can add an extra approach for your method to my short outline @siebenschlaefer
> 
> I'll wait for that for my final review, okay?

I don't want to force other people to do extra work on my approaches 😄
So currently I have mentioned and outlined the big integer approach, and we can add the link and the approach with better and more detailed explanations later.

As far as I am concerned, this version can go live. Although I am sure there are still typos and things. That can be fixed over times, when someone sees them. If no new issues arise with these docs in the next few days, I will push them and adjust things on the road.

ErikSchierboom (approved):
Sure!

kotp (approved):
Looking good!

PR #825, opened 2024-03-01
feat: add exercise for pointer concept

vaeng:


config.json
```diff
@@ -182,6 +182,13 @@
           "classes",
           "enums"
         ]
+      },
+      {
+        "slug": "windowing-system",
+        "name": "windowing-system",
+        "uuid": "ed2148a5-674c-4660-a050-b11ab240b876",
+        "concepts": [],
```
vaeng, line 190: needs to be updated

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+It should accept a parameter as a __pointer__ of type `position` as input.
+The `move` function is similar to `resize` however, this function adjusts the _position_ of the window to the requested value, rather than the size.
+
+As with `resize` the new position cannot exceed certain limits.
+
+- The smallest position is 0 for both `x` and `y`.
+- The maximum position in either direction depends on the current size of the window.
+  The edges cannot move past the edges of the screen.
+  Values larger than these bounds will be clipped to the largest size they can take.
+  E.g. if the window's size is at `x` = 250, `y` = 100 and a move to `x` = 600, `y` = 200 is requested, then the window would be moved to `x` = 550, `y` = 200 as the screen is not large enough in the `x` direction to fully accommodate the request.
```
siebenschlaefer, line 133: "the window's size is at `x` = 250, `y` = 100"
looks like a mix of a size (width/height) and a position (x/y). I guess you meant
"the window's size is `width` = 250, `height` = 100"

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+aProgramWindow.move(&newPosition);
+aProgramWindow.position->x;
+// => 50
+aProgramWindow.position->y;
+// => 100
+```
+
+## 6. Change a program window
+
+Implement a `changeWindow` function that accepts a __pointer__ to a `programWindow` object as input and changes the window to the specified size and position.
```
siebenschlaefer, line 148: "to the specified size and position" threw me off at first.  
I had to read this task 6 three times to understand that "specified" refers to the next sentence and that these new values are fixed. How about "a fixed" instead of "the specified"?

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+The window should get a width of 400, a height of 300 and be positioned at x = 100, y = 150.
```
siebenschlaefer, line 150: Everywhere else x and y are surrounded with backticks.

exercises/concept/windowing-system/.meta/exemplar.cpp
```diff
@@ -0,0 +1,54 @@
...
+    int targetY{newPosition->y > 0 ? newPosition->y : 0};
+    int maxY{this->screenSize->height - this->windowSize->height};
+    int newY{maxY > targetY ? targetY : maxY};
+
+    this->windowPosition->move(newX, newY);
+}
+
+void changeWindow(programWindow* window) {
+    window->windowSize->resize(400, 300);
+    window->windowPosition->move(100, 150);
```
siebenschlaefer, line 51: Not that it really matters but this implementation only works if the `resize()` does not clip the new size.
E.g. if the old position was at 799/599 (with a size of 1/1) the `resize()` would have no effect and the result of `changeWindow` would be incorrect.
Reversing the order of the two operations does not help either because a move would could also be clipped.

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+// => 0
+
+point.move(100, 200);
+point.x;
+// => 100
+point.y;
+// => 200
+```
+
+## 3. Define a `programWindow` class
```
siebenschlaefer, line 78: I'm not entirely sure but I think this is the only required type in the C++ track with a name in "lowerCamelCase". All other types are in "flatcase", "snake_case", or "PascalCase".
Soft vote for changing the type name to "ProgramWindow" or "program_window".

exercises/concept/windowing-system/.meta/exemplar.cpp
```diff
@@ -0,0 +1,54 @@
+#include "windowing_system.h"
+
+namespace windowing_system {
+
+void size::resize(int newWidth, int newHeight) {
+    this->width = newWidth;
```
siebenschlaefer, line 6: just out of curiosity: why all the `this->` prefixes in this solution?

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+       |          ║                 |      ├──────────────────────┤            ║
+screenSize.height ║                 |      │                      │            ║
+       |          ║            size.height │                      │            ║
+       |          ║                 |      │       contents       │            ║
+       |          ║                 |      │                      │            ║
+       |          ║                 |      │                      │            ║
+       |          ║                 v      └──────────────────────┘            ║
+       |          ║                                                            ║
+       |          ║                                                            ║
+       v          ╚════════════════════════════════════════════════════════════╝
```
siebenschlaefer, line 26: Did you use the operator `.` instead of `->` on purpose?

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+## 6. Change a program window
+
+Implement a `changeWindow` function that accepts a __pointer__ to a `programWindow` object as input and changes the window to the specified size and position.
+
+The window should get a width of 400, a height of 300 and be positioned at x = 100, y = 150.
+
+```cpp
+programWindow aProgramWindow{};
+changeWindow(&programWindow);
```
siebenschlaefer, line 154: programWindow -> aProgramWindow

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
+# Instructions
+
+In this exercise, you will be simulating a windowing based computer system.
```
siebenschlaefer, line 3: "windowing based" sounds odd. Is that how a native speaker would phrase it?

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+       |          ║                                                            ║
+       |          ║                                                            ║
+       v          ╚════════════════════════════════════════════════════════════╝
+```
+
+## 1. Define a `size` struct for storing the dimensions of the window
+
+Define a `struct` named `size`.
+It should have two member variables `width` and `height` that store the window's current dimensions.
+The constructor function should accept initial values for these fields.
```
siebenschlaefer, line 33: here and in task 2: "constructor function" sounds rather uncommon. How about just "constructor"?

These instructions use the terms "member variables" and "fields" as synonyms. Maybe we should stick to one term.

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+// => 0
+point.y;
+// => 0
+
+point.move(100, 200);
+point.x;
+// => 100
+point.y;
+// => 200
+```
```
siebenschlaefer, line 76: maybe `pos` instead of `point` to avoid introducing another term

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+## 3. Define a `programWindow` class
+
+Define a `programWindow` class or struct with __pointers__ to the following member variables:
```
siebenschlaefer, line 80: The member variables *are* pointers, they do not point to member variables (at least not necessarily).

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+- `screenSize`: holds a pointer of type `size` with `width` 800 and `height` 600
+- `windowSize` : holds a pointer of type `size`, the initial value is the default value of the `size` instance
+- `windowPosition` : holds a pointer of type `position`, the initial value is the default value of the `position` instance
```
siebenschlaefer, line 84: I'd suggest omitting "holds" in these three lines

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+
+When the window is opened (initialized), it always has the default `windowSize` and `windowPosition` in the beginning.
+Its constructor should not take any input parameters.
+
+```cpp
+programWindow aProgramWindow{};
+programWindow.screenSize->width;
```
siebenschlaefer, line 91: programWindow.screenSize -> aProgramWindow.screenSize

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+// => 800
+
+// Similar for the other fields.
+```
+
+## 4. Add a method to resize the window
+
+The `programWindow` class should include a function `resize`.
+It should accept a __pointer__ of type `size` as input and attempts to resize the window to the specified size.
```
siebenschlaefer, line 100: attempts -> attempt

exercises/concept/windowing-system/.meta/exemplar.h
```diff
@@ -0,0 +1,30 @@
...
+
+struct size {
+    int width{80};
+    int height{60};
+    void resize(int newWidth, int newHeight);
+};
+
+struct position {
+    int x;
+    int y;
```
siebenschlaefer, line 13: The instructions want a "constructor" with default values of `0`.

exercises/concept/windowing-system/.meta/exemplar.h
```diff
@@ -0,0 +1,30 @@
...
+    void move(int newX, int newY);
+};
+
+struct programWindow {
+    programWindow();
+    ~programWindow();
+    size* screenSize{nullptr};
+    size* windowSize{nullptr};
+    position* windowPosition;
```
siebenschlaefer, line 22: not that it matters, but why default values for `screenSize` and `windowSize` but not for `windowPosition`?

exercises/concept/windowing-system/.docs/instructions.md
```diff
@@ -0,0 +1,159 @@
...
+aProgramWindow.size->width;
+// => 600
+aProgramWindow.size->height;
+// => 400
+```
+
+## 5. Add a method to move the window
+
+Besides the resize functionality, the `programWindow` class should also include a function `move`.
+It should accept a parameter as a __pointer__ of type `position` as input.
```
siebenschlaefer, line 124: instead of "a parameter ... as input" I'd suggest either "a parameter ..." or "... as input".

siebenschlaefer (commented):
I like the exercise itself. In the JavaScript track it's used to show how objects can be constructed and composed, IMHO a really good exercise for that because it's simple enough for beginners but not trivial.

But here the exercise is used to showcase pointers and dynamic memory allocation. That looks rather forced. Why should the position and size be passed by pointer, why should `programWindow` allocate memory dynamically? That just makes it more difficult, without (at least to me) clear benefits. I'm missing the motivation for that.  
The dynamic memory allocation also adds the problem of ownership. Now the students have to write a destructor (and if they follow the rule of five also at least delete the copy constructor and copy assignment operator.)

I would really like a different exercise for this pointer concept, one where pointers (and maybe dynamic memory allocation) are a better fit.

vaeng:
@siebenschlaefer 
Thanks for all the feedback, you were completely right. This one was hard and it took me quite a while to get back into the exercise game. The windowing java thing was not the right fit for cpp pointers. 
I scraped everything and created a completely new exercise, it is more limited in range and probably more direct to help students understand the concepts.

@ahans 
You are very welcome to comment :raised_hands:

vaeng:
I added a task, so students can use the & opertor.

vaeng:
Hey @ErikSchierboom, do you have time to review this new exercise? I would like to move the syllabus along, but the other two maintainers seem busy.

iHiD (approved):
Waved through blindly.

PR #827, opened 2024-03-02
tests: update crypto-square exercise

ahans:
The original tests expected that the implementation also exposes intermediate steps, such as the normalized plaintext. I think this unnecessarily restricts the implementation and is also not reflected in the canonical data. So now I changed all tests to exactly correspond to the canonical data. Let me know if that's not what I should have done. I can also just add the one extra test case. But adding the test UUIDs would be a bit hard then ...

### DIFF
--- a/exercises/practice/crypto-square/crypto_square_test.cpp
+++ b/exercises/practice/crypto-square/crypto_square_test.cpp
@@ -5,82 +5,51 @@
 #include "test/catch.hpp"
 #endif
 
-TEST_CASE("normalize_capitals")
-{
-    REQUIRE("hello" == crypto_square::cipher("Hello").normalize_plain_text());
-}
-
-#if defined(EXERCISM_RUN_ALL_TESTS)
-TEST_CASE("normalize_spaces")
-{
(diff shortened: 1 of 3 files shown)


exercises/practice/crypto-square/crypto_square_test.cpp
```diff
@@ -5,82 +5,51 @@
...
+          "[8574a1d3-4a08-4cec-a7c7-de93a164f41a]") {
+    REQUIRE("tsf hiu isn" ==
+            crypto_square::cipher("This is fun!").normalized_cipher_text());
 }
 
-TEST_CASE("normalized_cipher_text_fun")
-{
-    REQUIRE("tsf hiu isn" == crypto_square::cipher("This is fun!").normalized_cipher_text());
+TEST_CASE(
+    "8 character plaintext results in 3 chunks, the last one with a trailing "
```
vaeng, line 39: Currently, line breaks in the "name" of the exercise will break the test runner, can you keep them on one line until I fix this behavior?

vaeng (approved):

PR #829, opened 2024-03-03
Update Tests based on the Problem Specification

KevDi:
Updated Tests according to the Specification


vaeng (approved):
I like the new additions, thank you.

PR #833, opened 2024-03-04
feat: add functions exercise

vaeng:
That has slipped, sorry.
I finished the concept but never made the respective exercise.

### DIFF
--- a/exercises/concept/lasagna-master/.docs/instructions.md
+++ b/exercises/concept/lasagna-master/.docs/instructions.md
@@ -3,7 +3,7 @@
 In this exercise you are going to write some more code related to preparing and cooking your brilliant lasagna from your favorite cookbook.
 
 You have five tasks.
-The first one is related to the cooking itself, the other three are about the perfect preparation.
+The first one is related to the cooking itself, the other four are about the perfect preparation.
 
 ## 1. Estimate the preparation time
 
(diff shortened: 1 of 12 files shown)


exercises/concept/lasagna-master/.docs/instructions.md
```diff
@@ -0,0 +1,107 @@
+# Instructions
+
+In this exercise you are going to write some more code related to preparing and cooking your brilliant lasagna from your favorite cookbook.
+
+You have five tasks.
+The first one is related to the cooking itself, the other three are about the perfect preparation.
```
ErikSchierboom, line 6:
```suggestion
The first one is related to the cooking itself, the other four are about the perfect preparation.
```

exercises/concept/lasagna-master/.docs/introduction.md
ErikSchierboom, line 1: My main comment would be that this exercise is quite "meaty," in introducing a lot. Function/overloading and default parameters go well together I think. Const could potentially be its own exercise, but it's also fine to just keep it as is for now.

ErikSchierboom (approved):

exercises/concept/lasagna-master/.docs/introduction.md
vaeng: I once had an extra exercise for const planned and I have evaluated C#'s "Authentication System" exercise. I am not sure it is a good fit.
Siebenschläfer also suggested, that the const could be moved, but we need to visit constexpr and const in classes again, so there will be some repetition and I planned to sprinkle it here and there to make it easier to digest.

If we keep it this way, we should watch the numbers and split some of it away, if users don't find it as well fitting as I did while writing it.

ErikSchierboom: Sure

PR #835, opened 2024-03-04
fix(check-formatting.sh): use relative directory

ahans:
The `bin/check-formatting.sh` script changed its directory to `bin` as the first step, consequently expecting paths of files to check to be relative to `bin`, when typically they would be given relative to the working directory. When not finding a file, the script would just print a warning, but still exit with error code 0, which is probably the reason why this went unnoticed for quite a while.

This commit fixes the relative directory issue. Further, when a given path cannot be checked, we return a non-zero code now. Lastly, the script would count the number of files with formatting issues and return that number as error code. This has the issue that if the number of violating files is a multiple of 256, error code would falsly be returned.

### DIFF
--- a/bin/check-formatting.sh
+++ b/bin/check-formatting.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
 print_help() {
     echo "This script verifies if one or more files are formatted using the provided"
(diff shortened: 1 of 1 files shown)


vaeng (approved):

ahans:
@vaeng Does merging this mean that you don't want to go forward with the pre-commit approach? Existing flies still need reformatting and the CI setup that is supposed to check all files for each commit on `main` is still broken.

vaeng:
Sorry, we might have a misunderstanding here.
I thought you needed this fix to be in place to have the hooks working correctly.

(If you are on the exercism discord, you can also ping me directly to discuss things.)

PR #843, opened 2024-03-15
fix(parallel-letter-frequency): add another missing assertion

ahans:


vaeng (approved):
Thanks for fixing this (and for using conventional commit messages).

PR #847, opened 2024-03-21
feat: add literals concept

vaeng:


vaeng (commented):
about.md will be added when introduction is done.

The plan is to combine the literals and the auto concept to provide a small concept exercise.
The exercise will center around compiler warnings, that can be fixed by adding the correct literals.

ErikSchierboom (approved):

PR #850, opened 2024-03-24
build: introduce pre-commit framework

ahans:
To avoid confusion, I updated this PR now to only use [pre-commit](https://pre-commit.com) to replace (and fix) the 
existing clang-format checks. Other checks can be added in follow-up PRs.

Here is the commit message from the commit that does the actual switch. The other commit reformats all the files that went through without proper formatting due to the bug in the setup:

----

This replaces the custom clang-format script with the pre-commit framework,
making it easier to add checks in the future and simplifying GitHub
integration.

The GH workflow is adapted to run checks now as well.

To use it locally, all you need to do is install pre-commit locally, typically
via:
```
pip3 install --user pre-commit
```

You can then run `pre-commit run --all` to run it locally. Note that this does _not_ install
a Git hook. If you want that as well, run `pre-commit install`.

For more details on pre-commit and other installation options, see
https://pre-commit.com/ and https://pre-commit.com/#install


vaeng (commented):
I did not check every single file, but it seems to work on the ones I tried.

.vscode/settings.json
vaeng, line 1: Is this file needed for the PR?

ahans: Probably not. I opened this accidentally against this repo instead of my fork. It's not cleaned up in any way yet. Just wanted to showcase how it could look like. I can clean up this PR and let you know when I'm done. Then you can do a proper review.

.github/CODEOWNERS
ErikSchierboom, line 1: Why was this file changed? Could you revert?

.vscode/settings.json
ahans: That file had some whitespace violation. I removed those now from this PR. This now is strictly only clang-format.

.github/CODEOWNERS
ahans: It was changed because I had the default pre-commit checks activated that check for trailing whitespace and proper EOF. While I still believe those are good checks to have, I removed them from the current PR and will open a follow-up to reduce the chance for confusion.

exercises/concept/lasagna-master/lasagna_master_test.cpp
```diff
@@ -11,26 +11,20 @@ using namespace lasagna_master;
...
-TEST_CASE("preparationTime: Preparation time for many layers with custom average time", "[task_1]") {
+#define REQUIRE_VECTOR_APROX_EQUAL(vec1, vec2, margin)       \
+    REQUIRE(vec1.size() == vec2.size());                     \
+    for (size_t i = 0; i < vec1.size(); i++) {               \
+        REQUIRE(std::abs(vec1.at(i) - vec2.at(i)) < margin); \
+    }
+
+TEST_CASE(
+    "preparationTime: Preparation time for many layers with custom average "
+    "time",
```
ahans, line 22: Until https://github.com/exercism/cpp-test-runner/issues/73 is fixed, we will need some more `// clang-format off` for cases like this.

exercises/concept/pacman-rules/pacman_rules_test.cpp
```diff
@@ -5,59 +5,64 @@
...
 }
 
-TEST_CASE( "ghost does not get eaten because no power pellet is active, even if not touching ghost", "[task_1]") {
-  REQUIRE_FALSE( can_eat_ghost(false, false));
-}
-    
-TEST_CASE("score when eating dot", "[task_2]") {
-  REQUIRE( scored(false, true));
+TEST_CASE(
+    "ghost does not get eaten because no power pellet is active, even if not "
```
ahans, line 24: Another one.

exercises/concept/lasagna-master/lasagna_master_test.cpp
ahans: Updated this now.

exercises/concept/pacman-rules/pacman_rules_test.cpp
ahans: Also updated.

ahans:
I ran all practice exercises through the test runner to find the ones that needed manual formatting, because clang-format would split long test names over multiple lines. All occurrences are fixed now.

vaeng (approved):

vaeng:
> I ran all practice exercises through the test runner to find the ones that needed manual formatting, because clang-format would split long test names over multiple lines. All occurrences are fixed now.

That would have been my final question. Thanks for all the work on this one.

vaeng:
I merged your other PR. Can you please fix the conflicts, so we can merge this and have beautiful formatting?

ahans:
> I merged your other PR. Can you please fix the conflicts, so we can merge this and have beautiful formatting?

Ok, just saw your approval and comment here. We could have just closed the other PR without merging if we're going with this version. Using both works as well, of course.

ahans:
Alright, it's rebased now. However, there was one other potential "issue" I wanted to point out. The solution templates that only have the namespace, e.g.,
```c++
namespace diamond {

}  // namespace diamond
```
with the current clang-format config get reformatted to look like this:
```c++
namespace diamond {}  // namespace diamond
```

I find this non-ideal, since a student would have to navigate inside `{}` and hit return there.

We could fix this with a comment:
```c++
namespace diamond {
    // TODO: implement solution
}  // namespace diamond
```
Then the student would still have to remove the comment, but still better than having `{}` IMHO.

Alternatively, we could use the clang-format option `KeepEmptyLinesAtTheStartOfBlocks`. I haven't tested this yet, but I think it would leave the template alone, since a namespace counts as a block in that context and the empty line would be at the beginning of the block. On the other hand, it would also leave empty lines in other places that we may not want. So I'm a bit torn on what to do.

Let me know what you think please!

vaeng:
> We could fix this with a comment:

I like this solution. 
What do you think @siebenschlaefer ?

ahans:
> Are you back with some more time to discuss this @ahans @siebenschlaefer?

Except for the formatting issue mentioned above I think this is done. Let me rebase and implement my proposal that includes the `TODO` comment and then we can merge. I will try to find some time in the next couple of days.

vaeng:
> Alternatively, we could use the clang-format option KeepEmptyLinesAtTheStartOfBlocks. I haven't tested this yet, but I think it would leave the template alone, since a namespace counts as a block in that context and the empty line would be at the beginning of the block. On the other hand, it would also leave empty lines in other places that we may not want. So I'm a bit torn on what to do.

This is a good addition for the students. I would rather have the comment in the block and have the other places neat and tidy.

vaeng:
This looks really clean now. Although I have to admit that I did not look through all the changes.

vaeng:
Yes, I am with you. We should merge with `[no important files changed]`.

ahans:
I find it unfortunate that I have to squash this to a single commit. Having it in two separate commits I find much cleaner. What's up with this rule?

vaeng:
I have no access to the settings for this repository. Maybe @ErikSchierboom or @SleeplessByte know more about this?

ahans:
> I have no access to the settings for this repository. Maybe @ErikSchierboom or @SleeplessByte know more about this?

Ok, let's discuss this elsewhere and live with the single-commit rule for now. I squashed it down to one commit already anyway. But it's cases like this where having separate commits I would find much nicer: They need to be merged together so that the repository remains in a consistent state at all times, yet there are clearly separate steps that should remain visible.

SleeplessByte:
We have no **rule** squash it, just a common practice. You can absolutely request that it is not squashed on merge.

ahans:
> We have no **rule** squash it exercism-wide, just a common practice. You can absolutely request that it is not squashed on merge.
> 
> For cpp idk if there is something else set-up.

There probably is, because I'm fairly certain only the option "Squash and Merge" was available. There is also "Rebase and Merge", but that was grayed out with a "not allowed in this repo" comment.

vaeng:
@ahans 

this commit broke the test runner in some way. People are having problems over at the get-help discord.

ahans:
I can see the problem with pacman rules. I only looked manually over the concept exercises and apparently wasn't careful enough. luhn also has a line break now, although I did run the test runner locally over all practice exercises. My test script must have been not good enough in that it did not report all errors. :( Very sorry!

Are you taking care of any of this? Just revert for now to get it quickly back into order?

SleeplessByte:
> There probably is, because I'm fairly certain only the option "Squash and Merge" was available. There is also "Rebase and Merge", but that was grayed out with a "not allowed in this repo" comment.

Again, this is not an exercism-wide rule. This repository having it doesn't make it a rule exercism wide. That said, I think it should be dropped from this repo! @ErikSchierboom / @iHiD can you check if the merge options are on for this repo? Can it be removed?

PR #856, opened 2024-04-09
fix(literals): move reference link into admonition

ahans:
Special blocks (admonissions) are renedered by themselves without including the rest of the page. A reference link from the bottom of the file is therefore not available and cannot be resolved properly. This is fixed by moving it to the bottom of the block instead of the bottom of the file.

### DIFF
--- a/concepts/literals/introduction.md
+++ b/concepts/literals/introduction.md
@@ -68,8 +68,9 @@ auto il_nam{"보는 것이 하는 것보다 더 재미있을 수가 없지"sv};
 
 A _string_view_ can be seen as a reference to a const string.
 
+[ascii-code]: https://www.ascii-code.com/
+
 ~~~~
 
 [numbers-concept]: https://exercism.org/tracks/cpp/concepts/numbers
 [in-depth-integers]: https://www.learncpp.com/cpp-tutorial/fixed-width-integers-and-size-t/
-[ascii-code]: https://www.ascii-code.com/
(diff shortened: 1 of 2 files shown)


vaeng (commented):
Thanks for the catch

concepts/literals/introduction.md
```diff
@@ -36,7 +36,7 @@ This is a good default case and use-cases with the narrower `float` type are les
 ```cpp
 auto light_year_in_m{9.46073e+15f}; // well in the range of float
 auto earth_to_edge_comoving_distance_in_nm{4.32355e+32}; // needs double type for magnitude
-auto eulers_number{2.718281828459045d}; // needs double type for precision
+auto eulers_number{2.718281828459045}; // needs double type for precision
```
vaeng, line 39: Did you remove it, because floating-points are resolved to doubles by default?

ahans: Yeah, I should have probably opened a separate PR for this. The `d` is suffix is not part of the [C++ standard](https://en.cppreference.com/w/cpp/language/floating_literal) it seems:

> suffix, if present, is one of f, l, F, L, f16, f32, f64, f128, bf16, F16, F32, F64, F128, BF16(since C++23).

ChatGPT claims it was a common thing, but I have never seen it before and apparently it is only supported by GCC. Clang and MSVC [report an error](https://godbolt.org/z/fGK55nsM8).

vaeng: Ah, I checked it with GCC. Good catch! I should run it through other compilers. I will add some words on the standard that will be deducted when nothing is added.

ahans: @vaeng  Can we merge this and create a ticket for adding some sentences that you mentioned, so it doesn't fall through the cracks?

vaeng: Yes, let's do this.

vaeng (approved):

PR #889, opened 2024-06-23
docs: N is not existing variable

rachirib:
Fixing this also on the documentation of the exercise, This should be pointing to `i`.
Previous PR was fixing it on the learning section but it seems it is also on the exercise itself


siebenschlaefer (approved):
LGTM. Thank you

PR #901, opened 2024-09-10
Hit requirement untested

nash8114:
There is a requirement for exercise ellens-alien-game's hit() method which states: "Make sure that the health points do not drop below zero." Adding a "get_health() == 0" at the end of the hit test now tests whether this is the case.

### DIFF
--- a/exercises/concept/ellens-alien-game/ellens_alien_game_test.cpp
+++ b/exercises/concept/ellens-alien-game/ellens_alien_game_test.cpp
@@ -32,7 +32,9 @@ TEST_CASE("Alien is always hit", "[task_2]") {
     REQUIRE(alien.hit());
 }
 
-TEST_CASE("Alien is alive while health is greater than 0 and stays dead afterwards", "[task_3]") {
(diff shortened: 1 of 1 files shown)


siebenschlaefer:
I like this additional test, I've seen a few solutions that did let the `health` become negative.
@vaeng, what do you think?

vaeng (approved):
Yes, that is a good addition.

I don't see what the code formatting check is complaining about though.

nash8114:
To be fair I did not think to run a formater. I will look at it tomorrow
and push a new commit if the formatting is off somehow.
.

ahans:
> Yes, that is a good addition.
> 
> I don't see what the code formatting check is complaining about though.

The issue is that the line is too long. I ran `clang-format` and pushed the fix. I also ran the test runner locally to make sure it parses this properly. Things seem fine. I think the only case where it has issues is when a single string is split over multiple lines. Different arguments of `TEST_CASE` can be on different lines with no issue.

ahans (approved):

PR #903, opened 2024-09-22
tests: update say exercise

Elahi-cs:
This PR is the result of running this command:
image

using the [test creation helper tool](https://github.com/exercism/cpp/wiki/Test-Cases-for-Practice-Exercises#test-creation-helper-tool) linked in the wiki, with some modifications to the code necessary for the tool to work.

### DIFF
--- a/exercises/practice/say/say_test.cpp
+++ b/exercises/practice/say/say_test.cpp
@@ -5,93 +5,88 @@
 #include "test/catch.hpp"
 #endif
 
-TEST_CASE("zero")
-{
-    REQUIRE("zero" == say::in_english(0ULL));
+
+/*
+An 'error' object is used as expected value to indicate that the
+input value is out of the range described in the exercise.
+*/
+
+TEST_CASE("zero", "[5d22a120-ba0c-428c-bd25-8682235d83e8]") {
+	REQUIRE("zero" == say::say(0));
 }
 
 #if defined(EXERCISM_RUN_ALL_TESTS)
-TEST_CASE("one")
-{
-    REQUIRE("one" == say::in_english(1ULL));
+
+TEST_CASE("one", "[9b5eed77-dbf6-439d-b920-3f7eb58928f6]") {
+	REQUIRE("one" == say::say(1));
 }
 
-TEST_CASE("fourteen")
-{
-    REQUIRE("fourteen" == say::in_english(14ULL));
+TEST_CASE("fourteen", "[7c499be1-612e-4096-a5e1-43b2f719406d]") {
+	REQUIRE("fourteen" == say::say(14));
 }
 
-TEST_CASE("twenty")
-{
-    REQUIRE("twenty" == say::in_english(20ULL));
+TEST_CASE("twenty", "[f541dd8e-f070-4329-92b4-b7ce2fcf06b4]") {
+	REQUIRE("twenty" == say::say(20));
 }
 
-TEST_CASE("twenty_two")
-{
-    REQUIRE("twenty-two" == say::in_english(22ULL));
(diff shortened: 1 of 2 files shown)


Elahi-cs:
@vaeng everything should be fixed now. A couple of questions:
* I had to modify the exercise creation tool that you linked on the wiki, because it expects configlet to be on the same folder, but configlet expects to be run on the root folder of the repo. Do you see it necessary/useful to add these changes to the tool branch?
* Would it be useful to document the process to update these test cases? I'm talking setup tool -> run tool -> format with `clang-format` -> check for errors. I don't know if it's something that could be of help in the future or if it the process will be discarded once the 6 remaining exercises have their tests updated.

vaeng:
Hello @Elahi-cs,

Thanks for your patience. You did the right thing to ask in the forum and also a nudge to look at this would have been fine. 

I think your change is good and welcome, I wanted to ask the other maintainers for feedback before going further, but I haven't had the time yet.

As a result of my late reply, this September or will become tagged for hacktoberfest.

I will not be able to check your code until Tuesday, but maybe @siebenschlaefer has some capacity this weekend.

exercises/practice/say/say_test.cpp
```diff
@@ -5,93 +5,98 @@
...
-        say::in_english(987654321123ULL));
+        say::in_english(987654321123));
 }
 
-TEST_CASE("raises_an_error_below_zero")
-{
-    REQUIRE_THROWS_AS(say::in_english(-1), std::domain_error);
+TEST_CASE("numbers below zero are out of range",
+          "[49a6a17b-084e-423e-994d-a87c0ecc05ef]") {
+    REQUIRE("{'error': 'input out of range'}" == say::in_english(-1));
```
ahans, line 93: It looks like the tool cannot handle cases with errors that well yet. For invalid input, we want the student to throw an exception.
```suggestion
    REQUIRE_THROWS_AS(say::in_english(-1), std::domain_error);
```

exercises/practice/say/say_test.cpp
```diff
@@ -5,93 +5,98 @@
...
 }
 
-TEST_CASE("raises_an_error_for_one_trillion")
-{
-    REQUIRE_THROWS_AS(say::in_english(1000ULL * 1000ULL * 1000ULL * 1000ULL),
-                      std::domain_error);
+TEST_CASE("numbers above 999,999,999,999 are out of range",
+          "[4d6492eb-5853-4d16-9d34-b0f61b261fd9]") {
+    REQUIRE("{'error': 'input out of range'}" ==
+            say::in_english(1000000000000));
```
ahans, line 99: As above. Fine to reformat the number, though:
```suggestion
    REQUIRE_THROWS_AS(say::in_english(1000000000000), std::domain_error);
```

ahans (changes requested):
The changes look mostly fine, but the two cases giving an error in CI need to be fixed.

ahans:
> Related question: is there a way to run the CI tests locally? I don't like making the pipeline run the tests every time, but I also don't know if there's a good way other than using `act`.

To run exactly what happens in CI, you probably won't get around using something such as `act`. But what happens is not exactly black magic and you can check the files in `.github` to see what's going on. The actual code tests are all managed via `CMake` and run [here](https://github.com/exercism/cpp/blob/main/.github/workflows/ci.yml#L24). So if you have all the prerequisites installed, you can this locally:

```sh
cmake .
# run all tests
cmake --build .
# alternatively, run just a single test:
cmake --build . -- test_say
```

vaeng (approved):
That looks good to me, I don't see any reason not to merge this now.
@ahans remarks seem to be addressed.

PR #904, opened 2024-09-27
docs: minor typo fix in reverse-string approach

apetresc:
Fixing a minor typo and punctuation error in the "Using the standard library" approach to `reverse-string`.


siebenschlaefer (approved):
Thank you!

PR #907, opened 2024-10-08
docs: update literals introduction with C++23 standard information

### LINKED ISSUE #902: docs(litertals): add comment on compiler differences
Yeah, I should have probably opened a separate PR for this. The `d` suffix is not part of the [C++ standard](https://en.cppreference.com/w/cpp/language/floating_literal) it seems:

> suffix, if present, is one of f, l, F, L, f16, f32, f64, f128, bf16, F16, F32, F64, F128, BF16(since C++23).

ChatGPT claims it was a common thing, but I have never seen it before and apparently it is only supported by GCC. Clang and MSVC [report an error](https://godbolt.org/z/fGK55nsM8).

_Originally posted by @ahans in https://github.com/exercism/cpp/pull/856#discussion_r1563183971_

vaeng:
closes #902

### DIFF
--- a/concepts/literals/introduction.md
+++ b/concepts/literals/introduction.md
@@ -33,6 +33,15 @@ auto net_worth_carlisle_cullen{46'000'000'000LL}; // int and uint are not enough
 Floating-point numbers usually resolve to `double` during compilation.
 This is a good default case and use-cases with the narrower `float` type are less frequent than the unsigned example above.
 
+~~~~exercism/advanced
+The [C++23 standard][cpp23-literals] introduces more fine grained control with literals like `f128` or `F16`.
+Some compilers allow their own literals.
+GCC permits the usage of `d` as a literal, but this is not forseen in the standard and can hinder cross-platform compatibility.
+
+[cpp23-literals]: https://en.cppreference.com/w/cpp/language/floating_literal
+
+~~~~
+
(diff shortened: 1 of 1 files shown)


concepts/literals/introduction.md
```diff
@@ -33,6 +33,15 @@ auto net_worth_carlisle_cullen{46'000'000'000LL}; // int and uint are not enough
 Floating-point numbers usually resolve to `double` during compilation.
 This is a good default case and use-cases with the narrower `float` type are less frequent than the unsigned example above.
 
+~~~~exercism/advanced
+The [C++23 standard][cpp23-literals] introduces more fine grained control with literals like `f128` or `F16`.
+Some compilers allow their own literals.
+GCC permits the usage of `d` as a literal, but this is not forseen in the standard and can hinder cross-platform compatibility.
```
BNAndras, line 39:
```suggestion
GCC permits the usage of `d` as a literal, but this is not foreseen in the standard and can hinder cross-platform compatibility.
```

Using foreseen feels awkward. I think we're just saying `d` isn't in the standard so let's say that. Foreseen makes me think the standard was supposed to include `d` but it wasn't for some reason. That might be misrepresenting the truth.

iHiD (approved):

PR #910, opened 2024-10-08
fix(say): remove misleading comment

ahans:
I think a student could be very confused by this comment when we expect them to throw in case of an error. This comment is verbatim copy from the [problem specification](https://github.com/exercism/problem-specifications/blob/main/exercises/say/canonical-data.json#L4). At the very least, it should be reworded to make clear that we expect an exception, but I think it's best removed.

[no important files changed]

### DIFF
--- a/exercises/practice/say/say_test.cpp
+++ b/exercises/practice/say/say_test.cpp
@@ -5,11 +5,6 @@
 #include "test/catch.hpp"
 #endif
 
-/*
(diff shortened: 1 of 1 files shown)


ahans:
[no important files changed]

ahans:
@vaeng What do you think?

vaeng (approved):
Yes, that is not necessary with our implementation

Issue #878, opened 2024-05-17
doctor-data task 4 "get_older_bob" is confusing


Peaj:
I think in this case either the test case is to simple or the function name is misleading.
The test only checks if the function returns the string "Bob".
Out of two Vessels it seems to return the name of the one called "Bob".

The function name suggests that in the case of two Bob Vessels it would return the older one.
As it only returns the name there is no way to check though.

Also what happens when none of the inputs is a Vessel called "Bob"?

With all accessible information the most sensible solution to the task here might be to ignore the inputs and always return "Bob".

I think this function needs more test cases and/or should be ranamed.

heijp06:
I could pick this up.

The only solution I see is that the `heaven::get_older_bob()` function is renamed to `heaven::get_older_vessel()`. Both in the test file and in the instructions of the exercise. The function compares the age of 2 objects of type `Vessel`. Apart from in the name of this function, "bob" only appears in the test file. Doing this change would invalidate previous solutions to this exercise. I see no way around that.

With this exercise there is also [this issue](http://forum.exercism.org/t/c-doctor-data-inconsistency-in-the-task-1-description/12014). I could do that as well or should that be done in a separate PR?

vaeng:
@heijp06 

Thanks for the initiative. You are very welcome to combine the issues into a single PR. Make sure to correct the errors in all the places, where the information might be doubled.

For code it can be examples in the instructions.md, help.md, the example files and the tests.

I like the change to older_vessel. We should add another test case, so it is clearer for the users. I might have been too close to the source material, where the vessels are all bob and the original vessel, and thus oldest is also called bob.

PR #915, opened 2024-10-10
Build concept exercises + fix concept exercise that does not build.

heijp06:
I wanted to fix issue #878, but noticed that the `doctor-data` exercise was not built or tested when the track was built. It turned out that this was the case for all concept exercises. I am assuming this is NOT by design.

I did the following:
- Change the build such that it builds both "concept" and "practice" exercises.
- Fix a concept exercise `freelancer-rates` that did not build.
- ~~Fix a typo in the text that comes with the `freelancer-rates` exercise.~~

`freelancer-rates` failed to build because of the following line of code:
```
int rounded_up{std::ceil(after_discount)};
```
Narrowing conversions (`ceil()` returns a `double`) are [not allowed with list-initialization](https://en.cppreference.com/w/cpp/language/list_initialization).

I tested the build of the track with `cmake . && make` on Ubuntu 22.04.


heijp06:
@vaeng can you review this?

vaeng:
Thank you for your PR.

My CMake-foo is not the best, I would prefer @ahans or @siebenschlaefer to look at your changes.

Please do not mix different issues into one PR. It is easier to handle pull requests if they only concern one problem at a time.
E.g. one for the docs, one for the build pipeline, and one for the non-compiling exercise (that has a fix in pr #916 with an added comment, because I have not seen your PR earlier).

vaeng (approved):

PR #916, opened 2024-10-11
Fix(freelancer-rates): avoid conversion warning in exemplar

vaeng:


vaeng:
Thanks to @heijp06 for catching this.

PR #917, opened 2024-10-11
Update `reverse-string` missing test cases

Elahi-cs:
As mentioned by @BNAndras in #775 , some of `reverse-string`'s test cases were created after opening the issue. This PR adds the missing test cases.

### DIFF
--- a/exercises/practice/reverse-string/.meta/tests.toml
+++ b/exercises/practice/reverse-string/.meta/tests.toml
@@ -1,6 +1,13 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [c3b7d806-dced-49ee-8543-933fd1719b1c]
 description = "an empty string"
@@ -19,3 +26,12 @@ description = "a palindrome"
 
 [b9e7dec1-c6df-40bd-9fa3-cd7ded010c4c]
 description = "an even-sized word"
+
+[1bed0f8a-13b0-4bd3-9d59-3d0593326fa2]
+description = "wide characters"
+
+[93d7e1b8-f60f-4f3c-9559-4056e10d2ead]
+description = "grapheme cluster with pre-combined form"
+
+[1028b2c1-6763-4459-8540-2da47ca512d9]
+description = "grapheme clusters"
--- a/exercises/practice/reverse-string/reverse_string_test.cpp
+++ b/exercises/practice/reverse-string/reverse_string_test.cpp
@@ -5,29 +5,47 @@
 #include "test/catch.hpp"
 #endif
 
-TEST_CASE("an_empty_string")
-{
-    REQUIRE("" == reverse_string::reverse_string(""));
+
+/*
+If property based testing tools are available, a good property to test is reversing a string twice: reverse(reverse(string)) == string
+*/
+
+TEST_CASE("an empty string", "[c3b7d806-dced-49ee-8543-933fd1719b1c]") {
+	REQUIRE("" == reverse_string::reverse_string(""));
 }
 
 #if defined(EXERCISM_RUN_ALL_TESTS)
-TEST_CASE("a_word")
-{
-    REQUIRE("tobor" == reverse_string::reverse_string("robot"));
+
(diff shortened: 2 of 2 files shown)


vaeng:
I'm pretty sure there is a discussion about this exercise with @siebenschlaefer somewhere.
I think the final consent was to not include the grapheme-based tests, as they are painful to implement in c++14 (and probably still are for c++23).

Maybe it was on the forums?

BNAndras:
Julia hides the grapheme tests behind [a flag the student needs to set](https://github.com/exercism/julia/blob/eec611adcb374b15e7ba0bef9d9b3c04f90f5205/exercises/practice/reverse-string/runtests.jl#L38) so that might be an option here. That'd require the example solution to be reworked.

vaeng:
> Julia hides the grapheme tests behind [a flag the student needs to set](https://github.com/exercism/julia/blob/eec611adcb374b15e7ba0bef9d9b3c04f90f5205/exercises/practice/reverse-string/runtests.jl#L38) so that might be an option here. That'd require the example solution to be reworked.

Great idea. I like that. We could hide it with compiler flag option and describe the how-to in `.docs/introduction.append.md`. Marked as an additional challenge, it would not even interfere with the test runner.

Elahi-cs:
@vaeng should I try to make these changes to the test cases to hide them behind a compiler flag in this same pull request? Could you elaborate as to how that wouldn't interfere with the test runner?

vaeng:
Do you understand how the `EXERCISM_RUN_ALL_TESTS` flag is used?

We could have a similar flag, like `ADVANCED_STRING_TESTS` and describe the usage and reasoning in an .append.md for expert students.

You do have to use this PR for both, as the CI checks will not run otherwise and they make a logical group for a PR.

(Also you need some code formatting to pass the other check).

Do you need help for these?

Elahi-cs:
> Do you understand how the `EXERCISM_RUN_ALL_TESTS` flag is used?
> 
> We could have a similar flag, like `ADVANCED_STRING_TESTS` and describe the usage and reasoning in an .append.md for expert students.

I do! I just have a question: Would `EXERCISM_RUN_ALL_TESTS` include all tests in this case? Or would both `EXERCISM_RUN_ALL_TESTS` and `ADVANCED_STRING_TESTS` need to be used to run both the tests and the grapheme cluster tests?

> You do have to use this PR for both, as the CI checks will not run otherwise and they make a logical group for a PR.
> 
> (Also you need some code formatting to pass the other check).
> 
> Do you need help for these?

That makes sense! I'll get on that then. I'll let you know if I have any further questions, thank you for your guidance.

vaeng:
Are you still working on this, or do you want me to finish your work @Elahi-cs

vaeng:
@ahans @siebenschlaefer

Do you think the append will suffice?

exercises/practice/reverse-string/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
+# Additional Instructions
+
+## Grapheme Clusters
+
+Unicode supports a lot of characters and languages.
+The displayed 'graphemes' are constructed from code points.
```
BNAndras, line 6: Neither graphemes nor code points are introduced before this sentence. An external link here might be useful as background info.

exercises/practice/reverse-string/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
+# Additional Instructions
+
+## Grapheme Clusters
+
+Unicode supports a lot of characters and languages.
```
BNAndras, line 5: Unicode is not introduced before this sentence. I'd similarly recommend an external link and perhaps a framing sentence about how Unicode is relevant before diving into the technical details.

exercises/practice/reverse-string/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
...
+The displayed 'graphemes' are constructed from code points.
+Their combination and _order_ is the basis of the resulting grapheme.
+Imagine, a simple algorithm that might flip the 'dots' from one place to another during reversal: `Noël` becomes `leöN`.
+Now consider more [complex constructions][grapheme-construction] like the Bengali word `ঝিল্লি`, which consists of six code points.
+
+As graphemes make the string reversal process a lot more complicated, we made the respective test cases optional.
+You can try them out by setting the `TEST_GRAPHEMES` compiler flag or editing the test file accordingly.
```
ahans, line 12:
```suggestion
You can try them out by setting the `TEST_GRAPHEMES` preprocessor define or editing the test file accordingly.
```

exercises/practice/reverse-string/.docs/instructions.append.md
```diff
@@ -0,0 +1,14 @@
...
+
+[grapheme-construction]: https://r12a.github.io/uniview/?charlist=%E0%A6%9D%E0%A6%BF%E0%A6%B2%E0%A7%8D%E0%A6%B2%E0%A6%BF
```
ahans, line 14: To me this sounds reasonable, but @BNAndras's points are valid. Having a few more external references probably wouldn't hurt.

However, what I'm missing most here is something about how to approach this in C++. Simply using something like `std::u16string`  won't work, since in the test cases we clearly use UTF-8. Are students expected to implement some UTF-8 handling themselves? Can they use some library?

PR #919, opened 2024-10-12
Install boost on windows

heijp06:
@vaeng I took a first stab at writing the installation instructions for Boost on Windows. Could you take a look?

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -66,32 +66,17 @@ You should be able to use other integrated IDE's such as JetBrains' CLion in a s
 
 ### Installing Boost on Windows
 
-The `meetup` and `gigasecond` exercises depend on the [Boost](https://www.boost.org/) libraries. The following steps
-describe how to set it up.
+The `meetup` and `gigasecond` exercises depend on the [Boost][boost-website] libraries.
+The following steps describe how to set that up.
 
-The steps assume you install Boost under `C:\Program Files\Boost`, writing here requires Administrator privileges. You
-can install in a different location if you prefer. The steps are for Boost 1.86, but any recent version should work. The
-minimum version required by the exercises is 1.58. If you use a different installation location or version than the ones
-used here, make sure to update the paths accordingly. The steps here are based on the
-[installation guide](https://www.boost.org/doc/libs/1_85_0/more/getting_started/windows.html) at
-[Boost.org](https://www.boost.org/). The steps further assume you have C++ development using Visual Studio configured.
-
-1. [Download Boost](https://www.boost.org/users/download/), downloading the [7-Zip](https://www.7-zip.org/) (.7z)
-archive is [recommended](https://www.boost.org/doc/libs/1_86_0/more/getting_started/windows.html#zip).
-1. Start 7-Zip as Administrator and extract the archive to `C:\Program Files\Boost`.
-1. You should now have a folder `C:\Program Files\Boost\boost_1_86_0` that contains several files and folders.
-1. Add `C:\Program Files\Boost\boost_1_86_0` to the `INCLUDE` environment variable.
-1. Compile Boost:
-   1. Open a `Developer Command Prompt` as Administrator.
-   1. Navigate to the boost installation: `cd "C:\Program Files\Boost\boost_1_86_0"`.
-   1. Configure the build process: `.\bootstrap.bat`.
-   1. Get a coffee (or two) while executing the build: `.\b2`.
-   1. The output might indicate it `failed` or `skipped # targets`. Ignore that.
-1. If you had Visual Studio open while doing all of this, close it to make sure the
-change to `INCLUDE` gets picked up.
+1. Download the [prebuilt windows libraries][sourceforge-boost-binaries] for Boost.
+   - The exercises require a minimum Boost version of 1.58.
+   - You can find the version of MSVC by clicking "Modify" in the Visual Studio Installer.
+1. Extract the libraries to a local folder.
+1. Set the `BOOST_ROOT` environment variable to the folder where you installed the Boost libraries.
+1. If you had Visual Studio open while doing this, close it to make sure the value of `BOOST_ROOT` gets picked up.
 1. Start Visual Studio and open the folder containing the `gigasecond` or `meetup` exercise.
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -64,6 +64,35 @@ of the Microsoft Visual C++ compiler.
 
 You should be able to use other integrated IDE's such as JetBrains' CLion in a similar manner.
 
+### Installing Boost on Windows
+
+The `meetup` and `gigasecond` exercises depend on the [Boost](https://www.boost.org/) libraries. The following steps
+describe how to set it up.
+
+The steps assume you install Boost under `C:\Program Files\Boost`, writing here requires Administrator privileges. You
```
ahans, line 72: I think it'd be more appropriate to suggest a folder in the user's personal files. Alternatively, putting it into something such as `c:\libs` is also common.

docs/INSTALLATION.md
```diff
@@ -64,6 +64,35 @@ of the Microsoft Visual C++ compiler.
...
+can install in a different location if you prefer. The steps are for Boost 1.86, but any recent version should work. The
+minimum version required by the exercises is 1.58. If you use a different installation location or version than the ones
+used here, make sure to update the paths accordingly. The steps here are based on the
+[installation guide](https://www.boost.org/doc/libs/1_85_0/more/getting_started/windows.html) at
```
ahans, line 76: While this is the official guide, most people would probably be better off with [prebuilt binaries](https://sourceforge.net/projects/boost/files/boost-binaries/). They are also linked from the [downloads](https://www.boost.org/users/download/) page.

docs/INSTALLATION.md
```diff
@@ -64,6 +64,35 @@ of the Microsoft Visual C++ compiler.
...
+[Boost.org](https://www.boost.org/). The steps further assume you have C++ development using Visual Studio configured.
+
+1. [Download Boost](https://www.boost.org/users/download/), downloading the [7-Zip](https://www.7-zip.org/) (.7z)
+archive is [recommended](https://www.boost.org/doc/libs/1_86_0/more/getting_started/windows.html#zip).
+1. Start 7-Zip as Administrator and extract the archive to `C:\Program Files\Boost`.
+1. You should now have a folder `C:\Program Files\Boost\boost_1_86_0` that contains several files and folders.
+1. Add `C:\Program Files\Boost\boost_1_86_0` to the `INCLUDE` environment variable.
```
ahans, line 83: I'm surprised that adding the Boost path to `INCLUDE` makes CMake find it. I'm fairly certain that's not the standard way of doing it. Instead, passing it to CMake via `BOOST_ROOT` I would find more natural.

It looks like with recent CMake releases, there's another complication in that they want to remove (or at least deprecate) the `FindBoost` . But it's probably fine to ignore that for now.

ahans (commented):
Thanks a lot for your effort improving documentation. Let's see if we can make it even easier for students!

vaeng (commented):
Thank you for the guide.

Apart from the comments from Shane, please also see the exercism [formatting guidelines for markdown documents](https://exercism.org/docs/building/markdown/markdown). 

Especially the part about links and keeping each sentence to its own line. It helps a lot with maintaining and tracking in git.

heijp06:
Thank you both, that is really great feedback!

I will try to see if I can make it work the way @ahans suggests. I prefer using something under `%USERPROFILE%` to install Boost, maybe just `%USERPROFILE%\Libs`.

Alternatively there is still the option to use a package manager like [vcpkg](https://vcpkg.io/) which would allow one to just put [boost-date-time](https://vcpkg.io/en/package/boost-date-time.html) in the project folder. I have not tried that route yet, but could if you think it is worth the try.

@vaeng The existing INSTALLATION.md does not follow the rules you link to, especially with regards to reference links. If I am going to adhere to the guidelines, I think it is best the entire document does. Would you want me to go over the entire document and make it rules compliant? If so would you want me to do that in this PR or in a separate PR. The latter option makes the most sense to me since we then separate the formatting changes to the existing document from the information I intend to add.

vaeng:
I think a separate PR to format the rest of the document is the cleanest solution, if your additions are already compliant.

docs/INSTALLATION.md
heijp06: I removed any mention of an installation folder.

docs/INSTALLATION.md
heijp06: Using the prebuilt binaries is a great suggestion, I completely missed that. This really cleaned things up.

docs/INSTALLATION.md
heijp06: I no longer suggest to add anything to `INCLUDE` and use `BOOST_ROOT` instead.

heijp06:
@vaeng, @ahans I did as you suggested. Things should be (much) better now.

docs/INSTALLATION.md
```diff
@@ -64,6 +64,20 @@ of the Microsoft Visual C++ compiler.
 
 You should be able to use other integrated IDE's such as JetBrains' CLion in a similar manner.
 
+### Installing Boost on Windows
+
+The `meetup` and `gigasecond` exercises depend on the [Boost][boost-website] libraries.
+The following steps describe how to set that up.
+
+1. Download the [prebuilt windows libraries][sourceforge-boost-binaries] for Boost.
```
vaeng, line 72: The enumeration is picked up correctly in the rich-view in github, but here every item is labeled as number 1. I am surprised that this works anyhow. :confused:

heijp06: @vaeng Please not that the same document [already uses this syntax](https://github.com/exercism/cpp/blob/main/docs/INSTALLATION.md?plain=1#L63).

It seems there is no formal markdown spec, but there is e.g. [this](https://commonmark.org/), [which states](https://spec.commonmark.org/0.31.2/#lists):

> The [start number](https://spec.commonmark.org/0.31.2/#start-number) of an [ordered list](https://spec.commonmark.org/0.31.2/#ordered-list) is determined by the list number of its initial list item. The numbers of subsequent list items are disregarded.

Using `1.` for all items eases maintenance when adding, removing or reordering items and helps with diffs.

vaeng (approved):

PR #920, opened 2024-10-30
Update `grade-school` tests

Elahi-cs:
This PR aims to update all tests in the `grade-school` exercise. The tests were created manually because the fact that some object-oriented adjustments were necessary made it impossible for the `exercise_creation_helper` to actually help. I'm hesitant the changes were done correctly because of my lack of experience with the platform, but I've tried to stick to the original test cases where possible and to use the same style with the missing test cases.

@vaeng let me know what you think.

### DIFF
--- a/exercises/practice/grade-school/.meta/tests.toml
+++ b/exercises/practice/grade-school/.meta/tests.toml
@@ -1,24 +1,86 @@
-# This is an auto-generated file. Regular comments will be removed when this
-# file is regenerated. Regenerating will not touch any manually added keys,
-# so comments can be added in a "comment" key.
-
-[6d0a30e4-1b4e-472e-8e20-c41702125667]
-description = "Adding a student adds them to the sorted roster"
-
-[233be705-dd58-4968-889d-fb3c7954c9cc]
-description = "Adding more student adds them to the sorted roster"
-
-[75a51579-d1d7-407c-a2f8-2166e984e8ab]
-description = "Adding students to different grades adds them to the same sorted roster"
+# This is an auto-generated file.
+#
+# Regenerating this file via `configlet sync` will:
+# - Recreate every `description` key/value pair
+# - Recreate every `reimplements` key/value pair, where they exist in problem-specifications
+# - Remove any `include = true` key/value pair (an omitted `include` key implies inclusion)
+# - Preserve any other key/value pair
+#
+# As user-added comments (using the # character) will be removed when this file
+# is regenerated, comments can be added via a `comment` key.
 
 [a3f0fb58-f240-4723-8ddc-e644666b85cc]
-description = "Roster returns an empty list if there are no students enrolled"
+description = "Roster is empty when no student is added"
+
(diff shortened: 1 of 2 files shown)


vaeng:
Thanks for the update. Something broke, so the tests are not passing

Elahi-cs:
> Thanks for the update. Something broke, so the tests are not passing

@vaeng I might need a little bit of guidance on this one. After running configlet, several new tests were added that were not there before. I've done my best to implement them, but some of them are not passing the pipeline tests. An example of this is `Cannot add student to same grade in the roster more than once`. This was my implementation:

```c++
TEST_CASE("Cannot add student to same grade in the roster more than once",
          "[87c871c1-6bde-4413-9c44-73d59a259d83]") {
    grade_school::school school_;
    school_.add("Blair", 2);
    school_.add("James", 2);
    school_.add("James", 2);
    school_.add("Paul", 2);
    REQUIRE(school_.grade(2) == vector<string>{"Blair", "James", "Paul"});
}
```

However it doesn't seem to be required by the solutions:
```
/home/runner/work/cpp/cpp/build_exercises/practice/grade-school/grade_school_test.cpp:59: FAILED:
  REQUIRE( school_.grade(2) == vector<string>{"Blair", "James", "Paul"} )
with expansion:
  { "Blair", "James", "James", "Paul" }
  ==
  { "Blair", "James", "Paul" }
  ```

Should I disregard these tests and mark them as not included in the `tests.toml`? Or is there something wrong with either my implementation or the pipeline checks?

Thanks in advance.

vaeng:
The new tests found an "error" in the example solution:
https://github.com/exercism/cpp/blob/acc32555eab077cf786f892b6ab27a22184d81c7/exercises/practice/grade-school/.meta/example.cpp#L13

I would rather fix this in the example solution, so the test can pass, than to exclude the test.

Elahi-cs:
> The new tests found an "error" in the example solution:
> 
> https://github.com/exercism/cpp/blob/acc32555eab077cf786f892b6ab27a22184d81c7/exercises/practice/grade-school/.meta/example.cpp#L13
> 
> I would rather fix this in the example solution, so the test can pass, than to exclude the test.

Thanks for the clarification @vaeng , I will work on fixing the example solution. Follow-up question: Is there a way to run the tests agains the example solution locally? When I run `cmake --build . -- test_grade-school` cmake seems to ignore the rest of the files and just try to compile `grade_school_tests.cpp` by itself. I don't want to be using the CI pipeline as a compiler/test runner. Am I missing something?

vaeng:
When I work locally, I imitate the exercise set-up and write my example code in the cpp and header files. When I am happy with the result I copy everything to the example files and reset the source and headers to their initial state.

I'm sure there is a more elegant solution, but this one is pretty okay.

Elahi-cs:
@vaeng it should be ready know but I'm hesitant about the quality of the solution. Could you give it a glance and let me know if there's something that should be improved? And thanks for everything!

PR #923, opened 2024-11-02
Add /WX (warnings as errors) compile flag when compiling with MSVC.

### LINKED ISSUE #922: Exercises are not compiled with "warnings as errors" with MSVC (Microsoft Visual C++).
When compiling with CLang or GCC warnings are treated as errors. The [relevant part of CMakeLists.txt](https://github.com/exercism/cpp/blob/3d106a5b8768c17c716fd0692013ffe4ae9937de/exercises/practice/hello-world/CMakeLists.txt#L48) is this:

```cmake
if("${CMAKE_CXX_COMPILER_ID}" MATCHES "(GNU|Clang)")
    set_target_properties(${exercise} PROPERTIES
        COMPILE_FLAGS "-Wall -Wextra -Wpedantic -Werror"
    )
endif()
```

For MSVC no such option is set. As a consequence code that contains mistakes that the user should be guarded against (e.g. not returning a value in all code paths) compiles without errors. Also when uploading the code to the website the stricter rules on the server will make code that compiled locally fail there.

Setting warnings as errors for MSVC is not that hard. There is an [MSVC specific part in CMakeLists.txt](https://github.com/exercism/cpp/blob/3d106a5b8768c17c716fd0692013ffe4ae9937de/exercises/practice/hello-world/CMakeLists.txt#L57):

```cmake
# Tell MSVC not to warn us about unchecked iterators in debug builds
if(${MSVC})
    set_target_properties(${exercise} PROPERTIES
        COMPILE_DEFINITIONS_DEBUG _SCL_SECURE_NO_WARNINGS)
endif()
```

This can be changed to:

```cmake
# Tell MSVC not to warn us about unchecked iterators in debug builds
if(${MSVC})
    set_target_properties(${exercise} PROPERTIES
        COMPILE_DEFINITIONS_DEBUG _SCL_SECURE_NO_WARNINGS
        COMPILE_FLAGS "/WX")
endif()
```

Compiling with `/WX` will make MSVC treat warnings as errors.

If this change is deemed useful, I can pick it up.

---

As an aside, the GCC/CLang `COMPILE_FLAGS`: `-Wall -Wextra -Wpedantic` also raise the warning level of the compiler.

This is also possible for MSVC. The default warning level for MSVC is `/W3`, this can be set to `/W4` or `/Wall`. I experimented with those levels and I do NOT propose setting those higher levels. E.g. with `/WX /W4` the following code: 

```cpp
char capital_a = std::toupper('a');
```

Will fail to compile because the compiler warns (as error) about a possible loss of precision when setting `capital_a` (a `char`) to the return value of `toupper()` (an `int`).

heijp06:
This PR fixes #922. When compiling with MSVC (Microsoft Visual C++) warnings are now treated as errors for all exercises.

### DIFF
--- a/exercises/concept/doctor-data/CMakeLists.txt
+++ b/exercises/concept/doctor-data/CMakeLists.txt
@@ -57,7 +57,8 @@ endif()
 # Tell MSVC not to warn us about unchecked iterators in debug builds
 if(${MSVC})
     set_target_properties(${exercise} PROPERTIES
-        COMPILE_DEFINITIONS_DEBUG _SCL_SECURE_NO_WARNINGS)
+        COMPILE_DEFINITIONS_DEBUG _SCL_SECURE_NO_WARNINGS
+        COMPILE_FLAGS "/WX")
 endif()
 
 # Run the tests on every build
--- a/exercises/concept/doctor-data/CMakeLists.txt
+++ b/exercises/concept/doctor-data/CMakeLists.txt
@@ -55,10 +55,12 @@ if(${EXERCISM_RUN_ALL_TESTS})
 endif()
 
 # Tell MSVC not to warn us about unchecked iterators in debug builds
+# Treat warnings as errors
+# Treat type conversion warnings C4244 and C4267 as level 4 warnings, i.e. ignore them in level 3
 if(${MSVC})
     set_target_properties(${exercise} PROPERTIES
         COMPILE_DEFINITIONS_DEBUG _SCL_SECURE_NO_WARNINGS
-        COMPILE_FLAGS "/WX")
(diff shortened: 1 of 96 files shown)


heijp06:
@vaeng, I created this PR for #922.

vaeng:
I assume you replaced the lines for all exercises and did not add anything else? I scrolled through the changes and read your explanation in the issue. Thanks for the write-up that helps a lot to speed things up.

I would love to have a single space for those files and have each exercise pull from that folder, but currently, we have to change it everywhere.

It is a good change, if all tests pass, we should merge this.

vaeng (approved):

PR #925, opened 2024-11-19
fix: improve docs to align with tests

vaeng:
Due to markdown formatting issues, the tests were not correctly alligned in the online-editor.

Also there were some leftover text blocks after removing dynamic memory paragraphs.


vaeng:
@IsaacG I just saw that you are also on the reviewers team and can thus help me out here with a review, so I can fix this exercise.

IsaacG (approved):

PR #932, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

IsaacG:
+cc @vaeng

vaeng:
Wow, what a long winded discussion. I'd rather think about a solution, that accepts the old solution but also sets etc.

IsaacG:
> I'd rather think about a solution, that accepts the old solution but also sets etc.

If possible, using a set or sorting the values would do the trick!

ahans:
@vaeng I can take care of this. Have been too inactive here lately...

ahans:
I opened a PR for this that allows an arbitrary order. But honestly, I think I like the append better. Ending up with a different order would be quite unnatural anyway, so clarifying the instructions is the way to go? I must admit I haven't read the entire thread (yet), and reading the comment here again, I wonder if this is about accepting also a solution that would return a set? We could sure do that. It would complicate the test a bit further, but maybe that's acceptable? @vaeng If you had something else in mind, let's discuss over at #933. But I'd also be fine with merging this and just closing #933.

IsaacG:
> I opened a PR (#933) for this that allows an arbitrary order. But honestly, I think I like the append better.

I think allowing any order makes more sense, to be honest :) While this order seems the most reasonable, I don't think a different order should be wrong. But also, I'm not a track maintainer here.

ahans:
Closing in favor of #933.

PR #943, opened 2025-02-20
triangle: add smaller size equality test

kossnikita:
This tests the case where the sides are sorted. If the two smaller sides are equal, the code below will falsely classify the triangle as scalene rather than isosceles. The current tests do not test this case.

```cpp
flavor kind(float a, float b, float c) {
    std::array<float, 3> sides{a, b, c};
    std::sort(sides.begin(), sides.end()); // Sides sorting
    if (!is_positive(sides)) {
        throw std::domain_error("Sides must be positive");
    }
    if (!triangle_inequality(sides)) {
        throw std::domain_error("The triangle inequality is violated");
    }
    if (sides[0] == sides[2]) {
        return flavor::equilateral;
    } else if (sides[1] == sides[2]) { // This check is incorrect
        return flavor::isosceles;
    } else {
        return flavor::scalene;
    }
}
```

### DIFF
--- a/exercises/practice/triangle/.meta/tests.toml
+++ b/exercises/practice/triangle/.meta/tests.toml
@@ -33,6 +33,9 @@ description = "isosceles triangle -> first two sides are equal"
 [d2080b79-4523-4c3f-9d42-2da6e81ab30f]
 description = "isosceles triangle -> first and last sides are equal"
 
+[eca8ae18-e191-4795-836f-c6457e649d80]
+description = "isosceles triangle -> two smaller sides are equal"
+
 [8d71e185-2bd7-4841-b7e1-71689a5491d8]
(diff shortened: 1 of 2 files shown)


siebenschlaefer:
Thank you for the PR.

The source of Exercism's C++ practice exercises is the language-independent [problem-specifications repository](https://github.com/exercism/problem-specifications). All changes to the problem statement or the tests should be discussed on the [forum](https://forum.exercism.org) first and reach a consensus before creating a PR for the problem-specifications repository. Once a problem has been modified it can be updated here on the C++ track.  
That might seem tedious but it helps keeping the exercises consistent across the various language tracks.

kossnikita:
[Forum topic](https://forum.exercism.org/t/triangle-add-smaller-size-equality-test/15873)

Hope I chose the right category.

PR #945, opened 2025-02-25
docs: add ranged based for loop concept files

vaeng:


concepts/ranged-loops/introduction.md
```diff
@@ -0,0 +1,88 @@
...
+}
+// Output: TrustNo1
+```
+
+The ranged loop syntax eliminates the need for manual indexing, making the code cleaner and less error-prone.
+Other languages reference to a similar access style as `for each` loop.
+
+The example above shows access by value.
+Although the `part` variable is changed inside the loop, the original vector remains unchanged.
+This means each element will be a copy and might produce overhead and may thus impact performance.
```
tasxatzial, line 39: Each element of what? Does it mean each part will be a copy?

concepts/ranged-loops/introduction.md
```diff
@@ -0,0 +1,88 @@
...
+
+```cpp
+std::vector<std::string> file_name {"Jraphics", "Interchange", "Format"};
+for (const std::string part : file_name) {
+    std::cout << part.front();
+}
+// Output: JIF
+```
+
+This prevents accidental modification of `number` but will still make a copy with every side-effect from above.
```
tasxatzial, line 53: This looks like an error. Number refers to what?

concepts/ranged-loops/introduction.md
```diff
@@ -0,0 +1,88 @@
...
+
+std::array<std::string, 3> trilobites {"Johnny", "Joey", "Dee Dee", "CJ"};
+
+for (std::string& member : trilobites) {
+    member += " Ramone";
+}
+// trilobites now contains "Johnny Ramone", "Joey Ramone", "Dee Dee Ramone", "CJ Ramone"};
+```
+
+If it is not necessary to change the container content it is advisable to use a `const` reference, which can be further optimized by the compiler.
```
vendethiel, line 72: I think all non-mutting traversals should use `const&`, so this probably deserves a bit more than a sentence

concepts/ranged-loops/introduction.md
tasxatzial: Aha, I see. Your suggestion conveys that more clearly 👍

vaeng:
> This might deserve a part on using it with `auto` as well, which is very often used and sometimes less error-prone, like if you forget the `const` on the key type when traversing a map and it has to copy the element.

Do you think we should change the current state further?

The auto keyword comes a bit later in the syllabus, together with smart pointers.

vendethiel:
Good point, I'm not sure. Maybe it should be introduced later, when they might actually mess it up :-)

iHiD (approved):

concepts/ranged-loops/introduction.md
```diff
@@ -0,0 +1,94 @@
...
+    // Use temp_element inside the loop
+}
+```
+
+- `element_type`: The type of each element in the container.
+- `temp_element`: A temporary variable that holds the value of each element.
+- `container`: The array or vector to loop over.
+
+## Example Usage
+
```
iHiD, line 21: This took me a minute to work out but I like it! :)

Maybe it would be worth saying...

```suggestion
In this code, we loop over each word, retrieving the first letter each time, to reveal the obscured password.

```

PR #946, opened 2025-02-27
fix: have test name on one line again

vaeng:
I found some.

### DIFF
--- a/exercises/practice/complex-numbers/complex_numbers_test.cpp
+++ b/exercises/practice/complex-numbers/complex_numbers_test.cpp
@@ -122,8 +122,9 @@ TEST_CASE("Arithmetic -> Subtraction -> Subtract purely imaginary numbers",
 }
 
 TEST_CASE(
-    "Arithmetic -> Subtraction -> Subtract numbers with real and imaginary "
-    "part",
+    // clang-format off
+    "Arithmetic -> Subtraction -> Subtract numbers with real and imaginary part",
+    // clang-format on
     "[f876feb1-f9d1-4d34-b067-b599a8746400]") {
     const Complex c1{1.0, 2.0};
     const Complex c2{3.0, 4.0};
@@ -148,8 +149,9 @@ TEST_CASE("Arithmetic -> Multiplication -> Multiply purely imaginary numbers",
 }
 
 TEST_CASE(
-    "Arithmetic -> Multiplication -> Multiply numbers with real and imaginary "
-    "part",
+    // clang-format off
+    "Arithmetic -> Multiplication -> Multiply numbers with real and imaginary part",
+    // clang-format on
     "[4d1d10f0-f8d4-48a0-b1d0-f284ada567e6]") {
     const Complex c1{1.0, 2.0};
     const Complex c2{3.0, 4.0};
@@ -197,8 +199,9 @@ TEST_CASE("Absolute value -> Absolute value of a negative purely real number",
 }
 
 TEST_CASE(
-    "Absolute value -> Absolute value of a purely imaginary number with "
-    "positive imaginary part",
+    // clang-format off
+    "Absolute value -> Absolute value of a purely imaginary number with positive imaginary part",
+    // clang-format on
     "[bbe26568-86c1-4bb4-ba7a-da5697e2b994]") {
     const Complex c{0.0, 5.0};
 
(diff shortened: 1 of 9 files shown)


vaeng:
We have to fix the issue in the test runner test discovery script.
And also we should make sure that our tests in the main repository are run under the same conditions as the docker version over in the test-runner.

vaeng:
The pre-commit checks will not let me merge this :D

ahans:
> The pre-commit checks will not let me merge this :D

Exactly, you need to put `// clang-format off` and `on` around a long name, e.g.:
```
TEST_CASE(
    // clang-format off
    "non-numeric, non-space char in the middle with a sum that's divisible by 10 isn't allowed",
    // clang-format on
    "[8b72ad26-c8be-49a2-b99c-bcc3bf631b33]") {
    REQUIRE(false == luhn::valid("59%59"));
}
```

ahans:
> We have to fix the issue in the test runner test discovery script.

Agreed, it's not exactly robust.

> And also we should make sure that our tests in the main repository are run under the same conditions as the docker version over in the test-runner.

I think they mostly do, don't they? What exactly is different?

I think I know why I didn't catch this. I ran the test runner on exercises in the `build_exercises` directly that CMake creates, so that the example solutions are in place. I must have accidentally used an old version there. :( TBH, this build_exercises copy is something that I don't particularly like. I will try to come up with something better.

ahans:
I can confirm you found them all.

exercises/practice/complex-numbers/complex_numbers_test.cpp
```diff
@@ -266,90 +270,70 @@ TEST_CASE("Complex exponential function -> Exponential of a purely real number",
 }
 
 // Extra Credit
-TEST_CASE(
-    "Complex exponential function -> Exponential of a number with real and "
-    "imaginary part",
+TEST_CASE("Complex exponential function -> Exponential of a number with real and imaginary part",
```
vaeng:
```suggestion
TEST_CASE(
    // clang-format off
    "Complex exponential function -> Exponential of a number with real and imaginary part",
    // clang-format on
```

exercises/practice/complex-numbers/complex_numbers_test.cpp
```diff
@@ -266,90 +270,70 @@ TEST_CASE("Complex exponential function -> Exponential of a purely real number",
...
     "[08eedacc-5a95-44fc-8789-1547b27a8702]") {
     const Complex c{std::log(2.0), M_PI};
 
     require_approx_equal(Complex(-2.0, 0.0), c.exp());
 }
 
-TEST_CASE(
-    "Complex exponential function -> Exponential resulting in a number with "
-    "real and imaginary part",
+TEST_CASE("Complex exponential function -> Exponential resulting in a number with real and imaginary part",
```
vaeng:
```suggestion
TEST_CASE(
    // clang-format off
    "Complex exponential function -> Exponential resulting in a number with real and imaginary part",
     // clang-format on
```

exercises/practice/complex-numbers/complex_numbers_test.cpp
```diff
@@ -266,90 +270,70 @@ TEST_CASE("Complex exponential function -> Exponential of a purely real number",
...
     "[d2de4375-7537-479a-aa0e-d474f4f09859]") {
     const Complex c{std::log(2.0) / 2.0, M_PI / 4.0};
 
     require_approx_equal(Complex(1.0, 1.0), c.exp());
 }
 
-TEST_CASE(
-    "Operations between real numbers and complex numbers -> Add real number to "
-    "complex number",
+TEST_CASE("Operations between real numbers and complex numbers -> Add real number to complex number",
```
vaeng:
```suggestion
TEST_CASE(
    // clang-format off
    "Operations between real numbers and complex numbers -> Add real number to complex number",
    // clang-format on
```

exercises/practice/complex-numbers/complex_numbers_test.cpp
```diff
@@ -266,90 +270,70 @@ TEST_CASE("Complex exponential function -> Exponential of a purely real number",
...
     "[06d793bf-73bd-4b02-b015-3030b2c952ec]") {
     const Complex c{1.0, 2.0};
 
     require_approx_equal(Complex(6.0, 2.0), c + 5.0);
 }
 
-TEST_CASE(
-    "Operations between real numbers and complex numbers -> Add complex number "
-    "to real number",
+TEST_CASE("Operations between real numbers and complex numbers -> Add complex number to real number",
```
vaeng:
```suggestion
TEST_CASE(
    // clang-format off
    "Operations between real numbers and complex numbers -> Add complex number to real number",
    // clang-format on
```

vaeng (commented):
Looks fine now, thanks for the quick help!

PR #949, opened 2025-03-12
Updated Freelancer Rates exercise instructions.

angiokas:
Added information about using the <cmath> library to be able to use floor() and ceil() functions for part 3 and 4 of this exercise.


vaeng:
Hey there,

thanks for the PR. I think you are right, we should improve the exercise. It is already in the hints for the exercise, but maybe we should be more explicit. I am hesitant to put it as directly as you have proposed though.

PR #953, opened 2025-03-24
Update making-the-grade instructions and hints to remove extra step.

Sencudra:
## What is the problem?
While the instructions section contains only 5 steps described in the instructions.md file, 6 steps are visible on the [web](https://exercism.org/tracks/cpp/exercises/making-the-grade/edit).

## Solution
Quick fix is to move 'How to calculate the intervals' to hints.md.

### DIFF
--- a/exercises/concept/making-the-grade/.docs/hints.md
+++ b/exercises/concept/making-the-grade/.docs/hints.md
@@ -24,7 +24,10 @@ Also being familiar with the following can help with completing the tasks:
 
 ## 3. Calculating Letter Grades
 
-- These are _lower thresholds_. The _lower threshold_ for a "D" is a score of **41**, since an "F" is **<= 40**.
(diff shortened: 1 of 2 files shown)


vaeng (approved):
Thanks!

Sencudra:
@superherointj 
1. This PR is about fixing description / UI desyncrhonization. It's not about hints content.
2. Any insights why you actually think there should be no hints? Are there any guidelines for this?

superherointj:
>     1. This PR is about fixing description / UI desyncrhonization. It's not about hints content.

Right. Sorry for the trouble then.

>     2. Any insights why you actually think there should be no hints? Are there any guidelines for this?

I don't think having hints is a problem. Just that the problem statement is unclear without the interval information.
I have already solved the problem, so I no longer care. Only provided feedback since others may find it useful.

PR #957, opened 2025-03-24
Sync exercise tests

IsaacG:


IsaacG:
Will split this into per-exercise PRs.

PR #960, opened 2025-03-24
Sync tests for exercise/practice/pig-latin

IsaacG:


vaeng (approved):
I added a clang format comment, so the name stays in one line. That's important to discovered by the test framework.

PR #969, opened 2025-05-09
Added alphametics exercise to the track

marcelweikum:
Hey,

this is my first PR, please check thoroughly if I missed something or did something wrong!

Thank you very much.

Kind regards,
Marcel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,16 +11,16 @@ on:
 jobs:
   linux-min:
     name: Linux Min Config
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        compiler: [clang++-6.0, g++-10]
+        compiler: [clang++-15, g++-12]
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
     - name: Install Dependencies
       # Boost must be installed only because the CMake version (3.12) can't
       # detect the installed Boost version (1.69)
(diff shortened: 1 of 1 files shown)


exercises/practice/alphametics/.meta/example.h
```diff
@@ -0,0 +1,14 @@
+#ifndef ALPHAMETICS_H
```
ahans, line 1: @vaeng @siebenschlaefer What's our take on include guards? We have 64 exercises that use `#if !defined(...)`, two use `#ifndef ...`, and 21 with `#pragma once`. Shouldn't we unify this? My vote would go to `#pragma once` for being the shortest and least error-prone. The only argument against it is the lack of standardization, but all relevant compilers implement it (and have done so for decades), so that's not a strong argument in my book. Wdyt?

vaeng: I'm all for `once`. I would have to look it up, but I'm pretty sure I favored it in the header exercise as well

marcelweikum: https://en.wikipedia.org/wiki/Pragma_once

> Thus, #pragma once serves the same purpose as [#include guards](https://en.wikipedia.org/wiki/Include_guard), but with several advantages, including less code, avoidance of name clashes, and sometimes improvement in compilation speed.

siebenschlaefer: Not a hard "No" vote but I prefer include guards over `pragma once` because the former ist standadized and the latter involves file system semantics.
See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-guards
But if you folks like `pragma one` more and think all exercises should do the same then go for IT.

PR #974, opened 2025-05-09
Added "book-store" exercise

marcelweikum:
Hey,

I added the book-store exercise with the unified pragma mark.
(Had whole lot of time today)

Kind regards,
Marcel


exercises/practice/book-store/.meta/example.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <map>
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include "book_store.h"
+
+namespace book_store {
+
+static const int PRICE_AFTER_DISCOUNT[6] = {0, 800, 1520, 2160, 2560, 3000};
```
ahans, line 12: This can be `constexpr`.

ahans: Probably would make sense to use `std::array` here. I can imagine all template parameters could be automatically derived, i.e.:
```suggestion
static constexpr std::array price_after_discount = {0, 800, 1520, 2160, 2560, 3000};
```

It looks like we're not quite consistent with uppercase vs lowercase for constants, but I'd be very much in favor of using `ALL_CAPS` only for preprocessor macros. @vaeng Do we have anything official on this?

exercises/practice/book-store/.meta/example.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+static int dfs(const std::array<int, 5>& state,
```
ahans, line 14: Do you want to restrict visibility of this function? Then put it into an anonymous namespace please.

exercises/practice/book-store/book_store.h
```diff
@@ -0,0 +1,10 @@
+#ifndef BOOK_STORE_H
```
ahans, line 1: `#pragma once` please.

exercises/practice/book-store/book_store_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+                {1, 1, 2, 2, 3, 3, 4, 5, 1, 1, 2, 2, 3, 3, 4, 5}) == 10240);
+}
+// clang-format off
+TEST_CASE(
+    "Check_that_groups_of_four_are_created_properly_even_when_there_are_more_groups_of_three_than_groups_of_five",
+    "[5260ddde-2703-4915-b45a-e54dbbac4303]") {
+    REQUIRE(book_store::total({1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2,
+                               2, 3, 3, 3, 3, 3, 3, 4, 4, 5, 5}) == 14560);
+}
+// clang-format on
```
ahans, line 86: You can limit the `// clang-format off/on` to just the test case name:
```suggestion
TEST_CASE(
    // clang-format off
    "Check_that_groups_of_four_are_created_properly_even_when_there_are_more_groups_of_three_than_groups_of_five",
    // clang-format on
    "[5260ddde-2703-4915-b45a-e54dbbac4303]") {
    REQUIRE(book_store::total({1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2,
                               2, 3, 3, 3, 3, 3, 3, 4, 4, 5, 5}) == 14560);
}
```

exercises/practice/book-store/book_store_test.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+TEST_CASE(
+    "One_group_of_one_and_four_is_cheaper_than_one_group_of_two_and_three",
+    "[b0478278-c551-4747-b0fc-7e0be3158b1f]") {
+    REQUIRE(book_store::total({1, 1, 2, 3, 4}) == 3360);
+}
+
+// clang-format off
```
ahans, line 94: Same here.

exercises/practice/book-store/.meta/example.cpp
```diff
@@ -0,0 +1,66 @@
...
+        }
+        int cost = PRICE_AFTER_DISCOUNT[k] + dfs(next, memo);
+        if (cost < best) best = cost;
+    }
+
+    return memo[state] = best;
+}
+
+int total(const std::vector<int>& basket) {
+    std::array<int, 5> counts = {0, 0, 0, 0, 0};
```
ahans, line 54: This would be a bit less redundant:
```suggestion
    std::array<int, 5> counts{};
```

exercises/practice/book-store/.meta/example.cpp
```diff
@@ -0,0 +1,66 @@
...
+    auto it = memo.find(state);
+    if (it != memo.end()) return it->second;
+
+    bool empty = true;
+    for (int c : state) {
+        if (c > 0) {
+            empty = false;
+            break;
+        }
+    }
```
ahans, line 25: You could use `std::all_of` here:
```suggestion
    const bool empty = std::all_of(state.cbegin(), state.cend(), [] (auto state) { return state == 0; });
```

exercises/practice/book-store/.meta/example.cpp
```diff
@@ -0,0 +1,66 @@
...
+    if (empty) return memo[state] = 0;
```
ahans, line 26: I'd rather use two statements here:
```suggestion
    if (empty) {
        memo[state] = 0;
        return 0;
    }
```

exercises/practice/book-store/.meta/example.cpp
```diff
@@ -0,0 +1,66 @@
...
+            if (mask & (1 << i)) {
+                next[idx[i]]--;
+                ++k;
+            }
+        }
+        int cost = PRICE_AFTER_DISCOUNT[k] + dfs(next, memo);
+        if (cost < best) best = cost;
+    }
+
+    return memo[state] = best;
```
ahans, line 50: Same as above.

ahans (changes requested):
Thanks for this implementation! Mostly suggestions for your consideration, except the include guards. 😃

exercises/practice/book-store/.meta/example.cpp
marcelweikum: Both done. Are there any ressources for coding styles for this repo? I don't want to create extra work

exercises/practice/book-store/.meta/example.cpp
marcelweikum: I removed the static for simplicity or do you prefer internal functions and variables?

exercises/practice/book-store/book_store.h
marcelweikum: Sorry.. forgot this one.

exercises/practice/book-store/book_store_test.cpp
marcelweikum: Done.

exercises/practice/book-store/.meta/example.cpp
marcelweikum: True! Done.

exercises/practice/book-store/.meta/example.cpp
marcelweikum: Great idea! Done.

exercises/practice/book-store/.meta/example.cpp
marcelweikum: Done.

exercises/practice/book-store/.meta/example.cpp
marcelweikum: Done.

exercises/practice/book-store/.meta/example.cpp
vaeng: Yes, screaming snake makros. I would adhere to the C++ Core Guidelines. (Which I might not have in some places where it slipped my mind) https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-all-caps

exercises/practice/book-store/book_store_test.cpp
vaeng: I thought about that. The downside is, that the comments would show up in the website's test summary for students and might confuse them? Maybe I'm too careful?

marcelweikum:
For future PRs: what’s the preferred way to give a heads-up before making changes (e.g., adding a new exercise)? As far as I know, the expectations vary between different tracks and maintainers, so I’d like to ask what the best approach would be here. Until today I was opening issues to inform about changes I'm planning to make.

vaeng:
If there is just a few seconds between the PR and the issue I would not bother and go directly for the PR.

If it is something with a lot of work for you, that might be turned down (like updating everything to a new test framework), you should open an issue for discussion first and ping us. For me, I'm fine with GitHub only, but you will get a wider audience on the forums.

I see this mostly to save you from spending time on things that are not necessary.

Also, half done and broken PRs are a pain to guide, moreso if people abandon them and we have more work piecing the pieces together than doing it from scratch. This doesn't seem to be the case with your current PRs though.

If I'm late with a reply, I don't mind another ping. Life can be busy.

marcelweikum:
Thanks for the info! I’ve usually opened an issue and then immediately started working on the exercise, just so someone knows that it’s being worked on.
Most of the time, it doesn’t take me longer than 30 minutes to an hour to finish. That’s why I just posted the next two to three PRs without opening issues first, I hope that’s okay.

My general plan is to go through all current and available exercises from the [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises) repo (excluding the deprecated ones) in alphabetical order and implement them for the C++ track.
It’s a great way for me to practice, I really enjoy it, and it helps expand the track a bit.

Would that approach be okay with you long-term, or is there anything I should keep in mind or reconsider?

ahans:
> My general plan is to go through all current and available exercises from the [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises) repo (excluding the deprecated ones) in alphabetical order and implement them for the C++ track. It’s a great way for me to practice, I really enjoy it, and it helps expand the track a bit.

I don't have that much time reviewing currently, so be prepared to wait a bit. Other than that extending the track sounds great! Thanks for the effort!

> Would that approach be okay with you long-term, or is there anything I should keep in mind or reconsider?

From my perspective sounds great in general. However, could you maybe open a tracking issue first? We had something similar for the #48in24 exercises (#779). First, having an overview of the missing exercises would be nice to have. Second, we could put people's names to "claim" an exercise. I suspect @marcelweikum will claim most, but maybe @vaeng  or myself want to implement one exercise or another as well.

Another topic is around tooling. @vaeng has some exercise generator on a [branch](https://github.com/exercism/cpp/tree/feat--practice-exercise-generator). I'm not sure what's missing there, but finishing that up and getting it on main would be good IMHO. In addition to having something that provides a skeleton, it would be great to have tooling that automatically creates the tests. The Python track has [something](https://github.com/exercism/python/blob/main/bin/generate_tests.py) for this. For each exercise, there's a Jinja template in `.meta`. For the arm64-assembly track, we have [generators](https://github.com/exercism/arm64-assembly/tree/main/generators) written in pure Python. I'm not sure which approach I like better. The template one is maybe a bit cleaner. IMHO we should aim for 100% generated tests, as that would make syncing of exercises much easier and less error-prone. We need a solution for cases where we deviate from the canonical data. For arm64-assembly, for some exercises extra test cases are simply put into the generator code. The approach on the Python track has probably something similar.

@marcelweikum would you be up for working on something like that as well? I think it would make sense to get the tooling in place before adding many more exercises to make sure they match the canonical data and also to avoid extra work: a retrospectively added generator only helps when syncing next time, while having a generator first can also create the initial test cases.

I understand if that's not your cup of tea. In that case I'd try to prioritze this myself and come up with some framework. Wdyt?

marcelweikum:
Hey @ahans,

> First, having an overview of the missing exercises would be nice to have.

That’s a great idea! If no one objects, I’d be happy to go through all exercises in the [problem-specifications repo](https://github.com/exercism/problem-specifications), exclude the deprecated ones, and create an issue that lists the current missing ones. I’d also keep it up to date so it’s easy to see what’s been claimed or completed.

> It would be great to have tooling that automatically creates the tests.

I agree! Long-term, that would be a great step forward and definitely makes sense to reduce manual effort and errors. I’m genuinely interested in working on that. That said, I currently don’t have a solid starting point for how the generators work, apart from using configlet to scaffold exercises, which I believe is a similar idea, just without the test generation from canonical data.

If you or @vaeng could explain briefly what’s missing or where to begin, I’d love to dive into it. I’d need to read up and experiment a bit, but I do have the time and motivation to help with that!

I’m really enjoying contributing to the track and appreciate the opportunity to help shape it.

vaeng:
There is a wiki article about the test and exercise generator in this repo. I'm on mobile, where it is difficult to find.

I like the Python templating, for maintenance, but boy, it's not straightforward to set up. I translated some exercises for the Python track and it took me more time to fiddle with the templating than writing it manually. Each exercise has a bit of nuance in its template so it was error prone for me as a newcomer.

marcelweikum:
> There is a wiki article about the test and exercise generator in this repo. I'm on mobile, where it is difficult to find.

I think I found it: https://github.com/exercism/cpp/wiki/Test-Cases-for-Practice-Exercises

> Each exercise has a bit of nuance in its template so it was error prone for me as a newcomer.

Sounds promising! 😆  But I'll try reading into it and also trying to dive into your [exercise_creation_helper.py](https://github.com/exercism/cpp/blob/8e20ff68698ef1ecf2d6da7ad17f7024434f69a0/bin/exercise_creation_helper.py)

PR #980, opened 2025-05-14
Synced tests in `crypto-square`

marcelweikum:
Hey,

as mentioned in https://github.com/exercism/cpp/issues/775#issuecomment-2875180279

Kind regards,
Marcel


marcelweikum:
I missed one file with #ifndef include guards. https://github.com/exercism/cpp/blob/ec14dfa38859bfb5743e287759ac7e7947c52e05/exercises/practice/affine-cipher/affine_cipher.h#L1-L2
I included it in this PR, hope that's ok.

PR #986, opened 2025-05-18
Remove unnecessary REQUIRE from test

miguelmarcondesf:
Hello, I think we have two uses of `REQUIRE` where it's not necessary due to functions that can not be evaluated by it.


ahans:
The [instructions](https://github.com/exercism/cpp/blob/main/exercises/concept/ellens-alien-game/.docs/instructions.md) clearly state that both methods, `hit()` as well as `teleport`, should return `true`, so it's appropriate to check for that in the tests.

miguelmarcondesf:
> The [instructions](https://github.com/exercism/cpp/blob/main/exercises/concept/ellens-alien-game/.docs/instructions.md) clearly state that both methods, `hit()` as well as `teleport`, should return `true`, so it's appropriate to check for that in the tests.

Thank you @ahans for answering that

PR #993, opened 2025-07-15
ci uses ubuntu-24.04

keiravillekode:


keiravillekode:
I generated a PR like this for each track.
http://forum.exercism.org/t/github-actions-runner-breaking-changes-announced/12612/10

We can close this PR as I understand the cpp track is intentionally retaining support for 22.04

PR #994, opened 2025-07-15
fix: small typo

c-git:
Subject verb agreement

### DIFF
--- a/exercises/concept/last-will/.docs/introduction.md
+++ b/exercises/concept/last-will/.docs/introduction.md
@@ -2,7 +2,7 @@
 
 ## Namespaces
 
-An important method for code organization are namespaces.
+An important method for code organization is namespaces.
 Two functions might have a naming collision, which can be resolved by putting them in different namespaces.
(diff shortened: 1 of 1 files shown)


exercises/concept/last-will/.docs/introduction.md
```diff
@@ -2,7 +2,7 @@
 
 ## Namespaces
 
-An important method for code organization are namespaces.
+An important method for code organization is namespaces.
```
ahans, line 5: While grammatically correct, I think your proposed version sounds a git odd ("is" followed by plural). How about this:
```suggestion
An important method for code organization is the use of namespaces.
```

ahans (commented):
Thanks for the fix and sorry for the long wait! Just a small suggestion ...

c-git:
> Thanks for the fix and sorry for the long wait! Just a small suggestion ...

No problem the wait wasn't that long. That's a great idea. Updated.

ahans (approved):
Thanks!

PR #995, opened 2025-08-03
bugfix: unused function complex_numbers_test.cpp

Lucky-Armin:
in `exercises/practice/complex_numbers/complex_numbers_test.cpp`: when running the first few test cases without the use of the require_approx_equal helper function, a warning is issued which will be treated as an error

### DIFF
--- a/exercises/practice/complex-numbers/complex_numbers_test.cpp
+++ b/exercises/practice/complex-numbers/complex_numbers_test.cpp
@@ -23,13 +23,13 @@ using complex_numbers::Complex;
 
 // Define a margin to use for float comparisons. Catch does not compute a good
 // epsilon for float values near 0.
-static const double eps = 0.005;
+static constexpr double eps = 0.005;
 
 // Custom Catch2 matcher for approximate equality of Complex numbers
 class ComplexApproxMatcher : public Catch::MatcherBase<Complex> {
    public:
-    ComplexApproxMatcher(const Complex& expected, double epsilon = eps)
-        : expected(expected), epsilon(epsilon) {}
+    ComplexApproxMatcher(const Complex& expected, double epsilon)
+        : expected{expected}, epsilon{epsilon} {}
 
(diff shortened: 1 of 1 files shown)


vaeng:
Hm. Is that something new, that is flagged now, but wasn't before? @ahans

vaeng (approved):
Seems reasonable.

Where did the error occur for you? In the test runner?

ahans:
It will probably happen whenever you have all tests commented out that don't use that function. The maybe_unused fixes it, but I think we have better options.

First, having a static function is not recommended in C++ for that usecase. Using an anonymous namespace would be better. Not sure it would not create a warning/error though.

Second, having functions for checks like that I consider a bit of an anti pattern. If there is a test failure, it will be reported for the line of the function and not the line in the test case that we are actually interested in. I would just skip this and use the two calls wherever they are needed. Alternatively, writing a proper Catch2 matcher would be the better option over such a plain function.

blakelewis:
In case you haven't seen the discussion in the [forum](https://forum.exercism.org/t/compile-errors-in-c-complex-number-tests/18742), this code currently fails to compile in the test runner.  The code is written for catch2 v2, while the Dockerfile in the test runner gets the latest (v3) version and there are incompatible changes between v2 and v3.  There seems to be a choice between changing the Dockerfile to get the older version or updating the test code to use v3.

vaeng:
We would "just" update the catch version on this exercise of the client copy with v3. It's a "feature", that the client side is done on a per exercise basis.

It's mostly a cop paste job and the custom comparison has to be redone.. sorry for not seeing this, my headspace currently doesn't allow for exercise tasks to be run 😵

ahans:
@vaeng I have this already fixed locally. Will open a PR in a moment. The only thing that needed fixing was the namespace. "Old-style matchers" are still supported.

What again was the reason we have a different Catch2 version in the runner than we have here next to the exercises and in `test/catch.hpp`?

PR #996, opened 2025-08-11
fix(complex-numbers): use runner's Catch2 version

ahans:
The test runner uses Catch2 v3.4.0, while locally we still had v2.13.6. The recent changes to the complex-numbers test work with v2.13.6, while they don't work with v3.4.0, causing compilation to always fail in the test runner. This commit changes the test to be compatible with v3.4.0 and updates the local Catch2 copy.


vaeng:
Thanks for taking your time fixing this

PR #1013, opened 2026-02-22
Attempt to make flaky test case more reliable

senarclens:
Testing for code not to be vulnerable to race conditions is inherently hard. The prior version of the test unfortunately led to a lot of false negatives (ie the test passed about 1/3 of the time even the tested code wasn't thread safe). With these minimal adjustments the test is much more reliable (although not perfect).


siebenschlaefer:
I don't understand the reason for this PR.  
The test is named "Can handle concurrent transactions" and is meant to catch solutions that do not handle concurrent deposits and withdrawals correctly. Reducing the number of threads makes it less likely that the test catches invalid solutions. Or am I misunderstanding something?

BTW: Thanks for catching the typo in the test name.

senarclens:
The discussion wrt this PR is available on https://forum.exercism.org/t/bank-account-attempt-to-make-flaky-test-case-more-reliable/30262/6 - unfortunately, the discussion is stale since the improvement isn't straightforward. However, I've included date showing how much better the test performs after applying the PR in there.

PR #1017, opened 2026-05-25
Format all `instructions.append.md` to start with an H1 (required but ignored) and H2

IsaacG:
See related discussions [here](https://forum.exercism.org/t/grep-instructions-append-may-need-a-visible-header/53923) as well as https://github.com/exercism/docs/pull/592. Append files should all have an H1. Including an H2 helps break the flow from the non-append file to the append file so it does not read as the same stream of thought.


IsaacG:
+cc @exercism/cpp

vaeng (approved):

PR #1019, opened 2026-06-04
introduction.md: fix wrong term describing assignment operator

amavrin:
"assignment operator assigns a variable with a literal" seems to be incorrect. Better to say "assignment operator assigns a variable with a value"


amavrin:
https://forum.exercism.org/t/freelancer-rates-introduction-md-fix-wrong-term-describing-assignment-operator/56464

> `the assignment operator assigns a value to a variable` is much easier to parse

PR #1020, opened 2026-06-06
Update hello_world.cpp

user22241:


vaeng:
This is intentionally wrong, as the student should correct it. Did you make this PR or some AI agent?

PR #1024, opened 2026-07-02
Update bank_account.h

arajgarhia-gh:
interface


vaeng:
Writing this is part of the exercise, that's why we do not give it to the students.

PR #1025, opened 2026-07-03
Update about.md

abhistark007:
changing binding is not possible but there it showed possible so made changes accordingly


vaeng:
Changing is possible: https://www.programiz.com/online-compiler/2mobAmSs5zltV

PR #1026, opened 2026-07-08
Bump GHA Ubuntu workflow runners to Ubuntu 26.04

IsaacG:


.github/workflows/ci.yml
```diff
@@ -11,10 +11,10 @@ on:
 jobs:
   linux-min:
     name: Linux Min Config
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-26.04
```
ahans: The idea here is to keep this at the latest version we officially support. Typically we increase that once official support ends. So `ubuntu-22.04` here should stay until ~April 2027.

.github/workflows/pre-commit.yml
```diff
@@ -21,6 +21,6 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          python-version: "3.10"
+          python-version: "3.14"
```
ahans: It should actually be save to throw that out and just use the default.

exercises/practice/complex-numbers/test/catch_amalgamated.cpp
```diff
@@ -468,10 +468,10 @@ namespace Catch {
     }
 
 namespace literals {
-    Approx operator "" _a(long double val) {
+    Approx operator ""_a(long double val) {
```
ahans, line 471: I believe we have this exact same file copied to each exercise. I'm not sure if changing only one is a good idea. If this fails on a newer compiler, we should probably upgrade Catch2 altogether (I feel we've discussed this and there was some reason not to upgrade, but I can't quite remember ...).

ahans (changes requested):
Thanks for the push here. If reverting only the `linux-min` change does not work on its own and you don't want to look into Catch2 upgrading etc., I (or some other cpp maintainer) can take care of it.

.github/workflows/ci.yml
IsaacG: Reverted

.github/workflows/pre-commit.yml
IsaacG: Dropped

exercises/practice/complex-numbers/test/catch_amalgamated.cpp
IsaacG: Happy to mass-apply this change across the repo.

IsaacG: https://github.com/exercism/cpp/pull/1027

IsaacG: https://github.com/exercism/cpp-test-runner/pull/101

IsaacG:
@ahans Friendly ping :)

exercises/practice/complex-numbers/test/catch_amalgamated.cpp
ahans: Still puzzled why it shows only in the complex-numbers test ...

ahans (approved):
Sorry for the wait, I was out without a computer for the weekend.

Created #1029 so we don't forget to implement a clean fix for this. But your change is a good stopgap, thanks for that!

exercises/practice/complex-numbers/test/catch_amalgamated.cpp
IsaacG: Hrm. That is odd. After #1027 is applied, it should no longer be there.

ahans: All good, both are merged now and tests still pass. I was just wondering why you needed only the fix for complex-numbers. But TBH I didn't do much digging to find out. It's definitely good to fix it for all exercises, since that also affected students on newer compilers working locally. It already showed in CI for the macos test, where we have `-latest`.

PR #1027, opened 2026-07-09
Remove the space in `"" _a` that makes the newer compiler sad

IsaacG:


ahans (approved):
lgtm, thank you!

Let's look into automating those updates in the context of #1029.

PR #1032, opened 2026-08-06
Add header file with include guards and function declaration

43H1-BOI:
Added include guards and function declaration for convert.


vaeng:
Is that an AI PR? Why would you change the header file pragma? Is there a good reason?

43H1-BOI:
> Is that an AI PR? Why would you change the header file pragma? Is there a good reason?

I do used AI to help me as i wrote the logic by myself but aren't able to figure out the problem  .

